PR #18, opened 2017-04-15
tectonic: Remove dependence on crate md5

### LINKED ISSUE #16: Remove dependency on "md5" crate
We should use rust-crypto for this, since we need that crate for SHA256 sums anyway.

musoke:
Use rust_crypto instead, it is being used elsewhere too.

Resolves #16


musoke:
Let me know what you think.

I didn't see any md5 specific tests, but I think this works.

pkgw:
Thanks, looks great! My only stylistic comment is that the comment in `md5_api.rs` doesn't feel that necessary since, in my opinion, the code there pretty much speaks for itself. But obviously that is not an important issue.

Issue #20, opened 2017-05-04
Use RustCrypto crates instead of rust-crypto


newpavlov:
rust-crypto is no longer [maintained](https://github.com/DaGenix/rust-crypto/pulls). [RustCrypto](https://github.com/RustCrypto) organization aims to succeed it.

Additionally it also provides algorithms divided into separate crates united by the same traits, so you no longer need to pull the whole library just for two hashes.

For your project you can just use `md-5` and `sha2` crates. I can create a draft PR, but I can't check if I made all necessary changes, as unfortunately I get compilation [error](https://gist.github.com/newpavlov/c90e3edfca69fdafda643e86ecfa0a3a) on my system.

pkgw:
Hmmm, is there really no more output from the compilation stage? It looks like the C compiler invocation is failing, but usually that ought to involve some error messages and whatnot. Do you not even have a C compiler installed on the relevant machine?

newpavlov:
I get bunch of warning without clear error. Full output of `RUST_BACKTRACE=1 cargo build` can be seen [here](https://gist.github.com/newpavlov/292203cbb2e54d0eff47d3a922ab2857). Exit status 256 strange too, shouldn't it have a negative status in case of an error?

pkgw:
Ah, [this line](https://gist.github.com/newpavlov/292203cbb2e54d0eff47d3a922ab2857#file-gistfile1-txt-L2053) shows the key error. It is not an error that I've seen before.

What OS / Linux distro / etc are you using to try to build the project? What version?

newpavlov:
Ubuntu 14.04, `libpng` version is `1.2.50-1ubuntu2.14.04.2`.

pkgw:
OK, I can get Tectonic to build on 14.04 as of commit 0dd1873. Does it work for you now?

newpavlov:
Yes, now it compiles without problems! I've created a PR with the discussed changes.

pkgw:
Merged ⇒ issue fixed. Thanks!

PR #22, opened 2017-05-11
Replaced rust-crypto with crates from RustCrypto project

newpavlov:
As was discussed in the #20.


src/digest.rs
```diff
@@ -47,14 +45,13 @@ pub fn hex_to_bytes(text: &str, dest: &mut [u8]) -> Result<()> {
...
 pub const DIGEST_NAME: &'static str = "SHA256SUM";
-pub use crypto::sha2::Sha256 as DigestComputer;
 
 pub fn create() -> DigestComputer {
-    sha2::Sha256::new()
+    Default::default()
 }
 
 
-#[derive(Clone,Copy,Debug,Eq,PartialEq)]
```
pkgw: Is there a reason for removing the Copy derivation here? I can't say that I have a great sense for the Rust customs about this sort of thing.

pkgw: Indeed, I think this removal caused the compile error on the CI.

newpavlov: Hm, I don't remember removing it. Sorry I missed it. Initially I worked on the older version of the code. Maybe it wasn't there? Either way, I'll add it back

pkgw:
Yes, the Copy derive was added only very recently. Thanks so much for the contribution!

Issue #36, opened 2017-05-31
tectonic does't handle filenames with spaces


MrSpock:
When you try to use tectonic with file that contain spaces in filename (MacOS) this is happening:

```
 » tectonic file\ name\ with\ spaces.tex
note: this is a BETA release; report issues at https://github.com/tectonic-typesetting/tectonic/issues
Running TeX ...
note: indexing http://purl.org/net/pkgwpub/tectonic-default
note: resolved to https://dl.bintray.com/pkgw/tectonic/tl2016extras/2016.0r3/tl2016extras.tar
note: downloading SHA256SUM
error: something bad happened inside TeX; its output follows:

===============================================================================
===============================================================================

error: failed to open input file "file"
```

alexander-bauer:
I've been able to reproduce this with a unit test on my fork, and I've been following through the code. I think a potential problem spot is [`tectonic/xetexini.c:2844-2889`](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/xetexini.c#L2844-L2889). I'm aiming to solve this if I can, and will open a pull request when I do.

pkgw:
Yes, I think you're on the right track. The way that the initial filename gets fed into the engine is quite odd — see [this comment](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/xetexini.c#L4167-L4172). Thanks for writing a test case and investigating!

alexander-bauer:
After some more tracking-down, I suspect the actual problem code may be in `scan_file_name` of [`tectonic/xetex0.c:10660-10680`](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/xetex0.c#L10660-L10680). Although this populates `cur_name` somehow, it's unclear to me exactly what `cur_name` is -- it doesn't appear to be a string of any form. Interpreted as a pointer, it leads away from the program address-space, and interpreted as the start of a buffer, it does not resemble a string.

alexander-bauer:
More specifically, I am guessing that [`more_name`](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/xetex0.c#L10425) returns `false` when it first encounters a space. Setting `stop_at_space = false` inside `scan_file_name` gets it to process the whole string (I think), but yields
```
! Emergency stop
*** (job aborted, no legal \end found)
No pages of output.
Transcript written on ?.
```

alexander-bauer:
[...]

PR #44, opened 2017-06-01
Handle filenames with spaces

alexander-bauer:
- This breaks all tests, because expected filenames no longer match the ones that are actually written. @pkgw I would appreciate your input on how to fix this, or what should be done about it.
- There was originally a test added in this PR for the behavior, but it failed even after the change was made (and not for the quoting reason.) I don't know what was wrong with it, but a better test should be written instead.

For my notes while developing this, see my comments in #36.

### DIFF
--- a/tectonic/xetexini.c
+++ b/tectonic/xetexini.c
@@ -2851,6 +2851,7 @@ init_io(string input_file_name)
     int k;
     unsigned char *ptr = (unsigned char *) input_file_name;
     UInt32 rval;
+    bool name_needs_quotes;
 
     stdin_ufile.handle = NULL;
     stdin_ufile.savedChar = -1;
@@ -2862,10 +2863,18 @@ init_io(string input_file_name)
     /* Hacky stuff that sets us up to process the input file, including UTF8
      * interpretation. */
 
+    /* Check if there is a space in the input_file_name. If so, quote it,
+     * because xetex interprets space as the end of the filename. Otherwise, we
+     * leave it unquoted, to maintain backwards compatibility. */
+    name_needs_quotes = (strchr(input_file_name, ' ') != NULL);
+
     buffer[first] = 0;
     k = first;
 
-    buffer[k++] = '"'; /* Quote filename so xetex accepts spaces */
+    /* If the name needs quotes, start them here. */
+    if(name_needs_quotes) {
+        buffer[k++] = '"';
+    }
 
     while ((rval = *(ptr++)) != 0) {
         UInt16 extraBytes = bytesFromUTF8[rval];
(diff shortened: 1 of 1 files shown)


alexander-bauer:
The tests are no longer broken, because filenames are now quoted only if necessary. New test cases will have to be written with this in mind.

pkgw:
Thanks for tackling this problem. I have to say that the current version of the solution feels a bit fragile to me; what happens if the user passes in a filename that itself contains quotation marks? And so on.

Can you point me to the code which ends up reading and parsing the quoted string? I am wondering if we can give the relevant function a special flag that tells it basically "just use this sequence of bytes as the input file name, no questions asked", and then activate that flag for the toplevel input file. I think that would be more robust against the other kinds of funky filenames that might be passed in.

alexander-bauer:
@pkgw The file name scanning happens in `scan_file_name(void)`, in [`tectonic/xetex0.c:10660-10680`](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/xetex0.c#L10660-L10680). This function operates on a bunch of variables that I didn't find the declarations of, so I was hesitant to mess with it.

If we're going to change things at this level, it looks like we should look at `more_name(UTF16_code)` in [`tectonic/xetex0.c:10425-10455`](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/xetex0.c#L10425-L10455). I suspect we could change it to simply terminate if the byte passed in is null. I'd be a little worried about side effects, because I have no idea what else uses that piece of code.

alexander-bauer:
@pkgw In fact, if we modify it to stop on null bytes (I think modification is necessary, but perhaps I'm not reading carefully enough), then we can probably just turn off the flag `stop_at_space`.

alexander-bauer:
I've experimented a little with null-terminating the string and turning off that flag, and I haven't been able to get our code to behave without breaking something else down the line. Maybe it would be acceptable to make the quoting code a bit smarter: any unquoted sequence of spaces is itself quoted, and otherwise the string is left alone.

That is certainly more fragile than I'd like, but there's a lot to untangle in order to make it better, I think.

pkgw:
Thanks for your work on this. I would like to credit you in the release notes for the next release; would  you prefer that I identify you as Alexander Bauer, `@alexander-bauer`, or some other way?

I've been thinking about the intersection of this issue and the other issues relating to input file paths. I think I see a more robust solution, and I hope to implement it soon.

alexander-bauer:
@pkgw Thanks very much for your recognition! I'd love to be credited as Alexander Bauer, and if it's in a context where GitHub user-linking will work, an `@alexander-bauer` link to my profile. Otherwise, just the name would be fantastic!

Issue #4, opened 2017-03-06
Improve installation story: pre-built packages, easy builds


alexreg:
Please add to this list if you are interested in a package for an OS not listed here.

- [x] Mac OS X (Homebrew) [assigned: @alexreg, @iKevinY]

pkgw:
I am not sure how best to do this, but I would love to poll users to get a sense of which kinds of packages would be most useful. For now, any users out there are invited to please add a comment to this bug with the pre-built package that you would most like to see!

elahn:
Windows x64

sebasmagri:
I just did `cargo install tectonic` and it worked. It should be documented.

pkgw:
@sebasmagri Good point! Pull requests to [the website repo](https://github.com/tectonic-typesetting/tectonic-typesetting.github.io) are more than welcome if you've got the inclination 😄. I hope to revamp the install page soon, if I can get a better sense of which install methods would be good for people. I don't want to overwhelm them with too many choices (or overwhelm myself with obligations to make a million different kinds of packages), but the `cargo` option should definitely be there.

alexreg:
> but the cargo option should definitely be there.

Yes, perhaps, although I think it should assume all prerequisites are already installed, and the environment is configured. For anything more complicated, you should be instructed to use your package manager.

zbeekman:
> I just did cargo install tectonic and it worked. It should be documented.

FYI, with most dependencies installed through Homebrew `cargo install tectonic` failed for me due to problems finding icu-uc.pc.... After setting my PKG_CONFIG_PATH so that it was able to find that file, the compilation fails with GCC-6 and with the default apple/clang compiler....

steveklabnik:
You may find https://github.com/japaric/trust useful

g2p:
My preferred installation method would be an apt invocation followed by a cargo invocation (or a tar extraction, if you choose to build binaries). The effort in distributing anaconda packages is laudable, but I prefer working with distribution packages and rust crates. I suggested flatpak before, as that's more standard than anaconda, but since the build isn't esoteric and works with cargo that may not be necessary.

alexreg:
Since the consumers of this package may be both developers and end-users, I think it's important to be able to avoid `cargo install` completely, in honesty.

rekka:
On Ubuntu 17.04, all I needed was

```
sudo apt-get install libfontconfig1-dev libharfbuzz-dev zlib1g-dev
[...]

PR #46, opened 2017-06-01
dist: include PKGBUILD from the AUR

alexander-bauer:
This has been discussed in issue #4. Including the PKGBUILD in this
repository has the benefit of versioning it alongside the code, and
eventually pushing it to the AUR as part of the continuous integration
pipeline. -- Alexander Bauer

### DIFF
--- a/dist/arch/PKGBUILD
+++ b/dist/arch/PKGBUILD
@@ -1,0 +1,23 @@
+# Maintainer: Jan Tojnar <jtojnar@gmail.com>
+pkgname=tectonic
+pkgver=0.1.5
+pkgrel=2
+makedepends=('cargo')
+arch=('i686' 'x86_64')
+pkgdesc='modernized, complete, self-contained TeX/LaTeX engine, powered by XeTeX and TeXLive'
+url='https://tectonic-typesetting.github.io/en-US/'
+license=('MIT')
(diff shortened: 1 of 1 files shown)


pkgw:
Would it complicate any of the workflows to have this stuff maintained in a separate repository, rather than the main tree? I somewhat prefer that option ... but it's definitely a weak preference, so if it would make life easier to have it all in-tree, that's fine.

alexander-bauer:
@pkgw The explicit downside to having it in a separate repository will be trouble triggering rebuilds when certain things in this repository happen. For example, when the version bumps in the code, you'll have to kick off a separate build job for the packages repository. (Depending on what CI system is used, this could be hard or easy.)

From there, the package building systems will probably have to re-clone the code repository. For some systems, this might have to happen anyway; but if they were kept in-tree, we might be able to get away with getting them to use the already-checked-out repository.

(For example, the PKGBUILD refers to a source repository that `makepkg` will automatically clone. If a directory of that name already exists, it will skip that step. This is often accomplished by symlinking back to the repository root.)

Besides those two issues, the biggest argument for doing is this way are the benefits listed on the first comment: easier CI integration and versioning alongside the code.

pkgw:
That sounds pretty reasonable to me.

PR #48, opened 2017-06-01
Support amsrefs

rekka:
This commit changes the condition that decides whether bibtex is run to
finding `\bibdata` in the aux file (was `\citation` and `\bibcite`).
Since bibtex exits with an error if it does not find `\bibdata` anyway,
this avoids running bibtex unecessarily when for instance amsrefs is
used to process the bibliography.

I also replaced `find(...).count() > 0` by `find(...).next().is_some()`: no need to go through the whole file, stop at the first match.

### DIFF
--- a/src/cli_driver.rs
+++ b/src/cli_driver.rs
@@ -548,8 +548,8 @@ impl ProcessingSession {
 
             let use_bibtex = {
                 if let Some(auxdata) = self.io.mem.files.borrow().get(self.aux_path.as_os_str()) {
-                    let cite_aut = AcAutomaton::new(vec!["\\citation", "\\bibcite"]);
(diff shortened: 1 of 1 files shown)


pkgw:
I will confess that I don't fully understand the meaning of the control sequences that you're talking about here, but you sound like you know what you're doing :-) Thanks for the patch!

rekka:
Thanks for merging!

I have to admit that this info is based on some experimentation and on how latex defines the `\bibliography` command in `latex.ltx`:

```latex
\def\bibliography#1{%
  \if@filesw
    \immediate\write\@auxout{\string\bibdata{#1}}%
  \fi
  \@input@{\jobname.bbl}}
```

PR #60, opened 2017-06-03
Rerun build.rs only on changes in the tectonic/ dir

rekka:
This is really only a convenience for developers: there is not need to recompile the C/C++ code on every unrelated change in the repo.


pkgw:
OMG, I didn't know this was a thing!

PR #66, opened 2017-06-04
Fix warnings in tectonic/

ronnychevalier:
This PR starts fixing some warnings. It also adds some additional warning flags.

The only remaining "important" warnings of `-Wall` to fix are those related to `-Wpointer-sign`, otherwise it is only unused parameters/variables. I added a commit to disable `-Wpointer-sign` until these warnings are fixed. It avoids spamming the build output with warnings related to `tectonic/` code.


rekka:
Nice. Looks good to me.

build.rs
```diff
@@ -155,6 +155,20 @@ fn main() {
...
+        .flag("-Wno-unused-variable")
+        .flag("-Wmissing-include-dirs")
+        .flag("-Wredundant-decls")
+        .flag("-Wpointer-arith")
+        .flag("-Wformat=2")
+        .flag("-Wundef")
+        .flag("-Wunreachable-code")
+        .flag("-Wstrict-prototypes")
+        .flag("-Wold-style-definition")
+        .flag("-Wwrite-strings")
```
pkgw: I am concerned that these options could cause problems for platforms where the compiler is not actually `gcc`. (While the Rust crate is named `gcc`, it will use whatever system C compiler it thinks best.) Is there a way to test whether the compiler is `gcc` and only add these flags if we can confirm that it is? If not, I would rather not apply all of these extra flags.

tectonic/XeTeXFontInst.cpp
```diff
@@ -315,6 +315,9 @@ XeTeXFontInst::initialize(const char* pathname, int index, int &status)
     TT_OS2* os2Table;
     FT_Error error;
     hb_face_t *hbFace;
+    size_t sz;
+    ssize_t r;
+    FT_Byte *data;
```
pkgw: Late-time variable declarations are legal in C++, I believe, and they're kind of nice, so I would rather leave them where they were before in this function.

tectonic/core-bridge.c
```diff
@@ -69,7 +71,7 @@ dvipdfmx_simple_main(tt_bridge_api_t *api, char *dviname, char *pdfname)
 {
     extern int dvipdfmx_main(int argc, char *argv[]);
 
-    char *argv[] = { "dvipdfmx", "-o", pdfname, dviname };
+    char *argv[] = { xstrdup("dvipdfmx"), xstrdup("-o"), pdfname, dviname };
```
pkgw: Sorry for being lazy here — why are these `xstrdup`s needed?

tectonic/dpx-fontmap.c
```diff
@@ -1281,7 +1281,7 @@ strip_options (const char *map_name, fontmap_opt *opt)
     return font_name;
 }
 
-#if  DPXTEST
+#ifdef  DPXTEST
```
pkgw: This whole chunk of code can just be deleted because we are not in fact ever going to `#define DPXTEST`.

tectonic/dpx-pdfobj.c
```diff
@@ -1693,8 +1693,9 @@ apply_filter_TIFF2_1_2_4 (unsigned char *raster,
                 }
             }
         }
-        if (outbits > 0)
+        if (outbits > 0) {
             raster[k] = (outbuf << (8 - outbits)); k++;
```
pkgw: Yikes! This changes the semantics of the code — the brace-less `if` only applies to one statement, so the `k++` is executed regardless of the result of the `if` test. It would be good to give this function a careful read and figure out what the correct behavior is.

tectonic/dpx-subfont.c
```diff
@@ -474,7 +474,7 @@ release_sfd_record (void)
 }
 
 
-#if  DPXTEST
+#ifdef  DPXTEST
```
pkgw: Ditto about just deleting the code chunk.

tectonic/dpx-truetype.c
```diff
@@ -949,7 +949,7 @@ pdf_font_load_truetype (pdf_font *font)
         return  -1;
     }
 
-#if  ENABLE_NOEMBED
+#ifdef  ENABLE_NOEMBED
```
pkgw: Hmmm, the same consideration may apply here, but I don't know what `ENABLE_NOEMBED` does, so I'm not sure whether we want to act as if it is defined, or not. Out of scope for this PR, though.

tectonic/xetexini.c
```diff
@@ -3736,7 +3736,7 @@ tt_run_engine(char *dump_name, char *input_file_name)
     /* Make this something invariant so that we can use XDV files to test
      * reproducibility of the engine output. */
 
-    output_comment = "tectonic";
+    output_comment = xstrdup("tectonic");
```
pkgw: Note-to-self that this is another `xstrdup` I don't understand.

pkgw:
By the way, the Travis failure is something unrelated to this PR — I will work on fixing that separately.

pkgw:
If you rebase or merge this PR with the latest `master`, the Travis CI run should be able to succeed.

tectonic/core-bridge.c
rekka: `const char*` was being assigned to `char*`. So there is a potential that the program might try to change the static strings.
Actually, wouldn't just declaring `argv` with `const char *argv[]` solve this since `argv` is not used to mutate the content in the code as far as I can tell? This would need also `dvipdfmx_main` to be declared with `const char* argv`, or would just casting be enough like in `dvipdfmx_main(4, (char**) argv)` (I'm not sure if this is the correct syntax for the cast)?

tectonic/xetexini.c
rekka: This could probably be also solved by declaring `output_comment` as `const char *`.

tectonic/core-bridge.c
pkgw: OK, yes, that was my hunch. It would be much better to const-ify the variables rather than perform extra mallocs, if that's possible — which I think it shoud be in the cases where it came up here.

tectonic/dpx-pdfobj.c
Mrmaxmeier: In this case `k++;` can be safely removed because the variable only ever gets [set](https://github.com/ronnychevalier/tectonic/blob/ababa0acbb2faad2314857ee3b318492c6b883a0/tectonic/dpx-pdfobj.c#L1671) after the increment operation.

Mrmaxmeier: There are some other [questionable lines of code](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/xetex0.c#L23185) though that'll need deeper understanding to be fix with confidence.

rekka: How did that happen? In `xetex.web` the line is `cur_val1:=cur_val1 + cur_val * @"200000;`

Mrmaxmeier: That line from `xetex.web` probably results in [line 23163 in xetex0.c](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/xetex0.c#L23163).

rekka: Fixed in #72

pkgw: Good catch!

tectonic/xetexini.c
ronnychevalier: No, *currently*, it is not possible. In [engine-interface.c](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/engine-interface.c#L38), `output_comment` is allocated. Only here it is a constant. Thus, I use `xstrdup`.

I think it could be fixed but differently, like not using a global variable. However, it requires more significant changes. Another PR improving the code by removing global variables would be more appropriate.

tectonic/dpx-truetype.c
ronnychevalier: Ok I will leave this code then.

tectonic/core-bridge.c
ronnychevalier: It is not possible to use a const. `dvipdfmx_main` calls `do_early_args` and `do_args`, which call `getopt_long`. This function requires a `char * const argv[]`.

The only solution would be to avoid the allocation would be to do an explicit cast. I'm not sure this is the best approach. A better approach would be to fix the code on its own and not have a function that expect `argc` and `argv` (I think such significant change belongs to another PR).

Hence, my solution was to use `xstrdup` which is used elsewhere in such a case. I don't think two additional allocations will have an impact on the performance. It also fixes the issue without using an explicit cast that hide the issue (i.e., any function inside `dvipdfmx_main` could still try to modify the content of `argv[0]` or `argv[1]` resulting in a crash).

tectonic/XeTeXFontInst.cpp
ronnychevalier: Ok. It is just an habit, in my case I prefer to have variables declared at the beginning of the current scope :) I will go back to the previous declaration.

build.rs
ronnychevalier: AFAIK, the rust crate does not allow to test if a flag works or does not [currently provide a way to implement such test](https://github.com/alexcrichton/gcc-rs/issues/48).

Do you prefer to remove the additional flags and wait that the rust crate implements an API to do so? Do you prefer to check first if those flags are supported by other compilers? if so which ones (clang or another one?)?

tectonic/dpx-pdfobj.c
ronnychevalier: Ok, I will add another commit removing `k++;` then. Thanks :)

ronnychevalier:
@pkgw Rebased and applied fixes related to most of the comments. I did not remove the additional flags yet, I am waiting for your comment on that. I also did not change how I fix the assignation of const to a non-const variable. I think it fixes temporarily the issue properly better than casting away the value which just hide the issue, if you think not I can fix the commit to cast the value. A better fix obviously would be to improve the overall quality of the code, but I guess this will be done incrementally :)

pkgw:
@ronnychevalier For the const-to-non-const assignments, the idea we were discussing was to make the destination variables const as well — I think this should be doable without requiring too many changes. If that's not the case, though, we can just take your approach for now and fix up later.

build.rs
pkgw: I think it is better to remove the extra flags for now.

ronnychevalier:
`output_comment` cannot be const, as explained in my previous message, since it is a global variable that is set dynamically in [engine-interface.c](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/engine-interface.c#L38). For `argv` it cannot be const as well since by following the trail of functions that use it, the final `getopt_long` function in glibc does not expect a const. So we would have to cast it from const to non-const.

build.rs
ronnychevalier: Ok. I removed the commit from the branch.

pkgw:
Right, sorry, I wrote my first reply before seeing and reading your replies inline with the code review.

With your explanations, I think this will be good to merge once the extra compiler flags are removed. (For the time being — I would certainly like to enable lots of warnings in general! But I don't want to accidentally break the build on unusual OS setups.)

Issue #54, opened 2017-06-01
Automatically deploy PKGBUILD to AUR


alexander-bauer:
The AUR accepts package updates via Git, so updating an AUR package is as simple as pushing a git repository to their server. This repository must have the PKGBUILD at its root, however. One way I think we can accomplish this is to extract the `dist/arch/` directory as a subtree on-the-fly, and push that to the AUR as one step in a continuous deployment pipeline.  If the subtree is extracted in the same way every time, it will keep the same history, with just the commits relevant to that directory, regardless of when or where it is generated.

Doing this would mean collaborating with @jtojnar, who owns the package on the AUR.

jtojnar:
I think the most reasonable would be creating a new AUR account and adding it as co-maintainer. Then adding the following script into [deploy section](https://docs.travis-ci.com/user/deployment/script/) of `.travis.yml`.

```shell
git clone ssh://aur@aur.archlinux.org/tectonic.git staging/arch
cp dist/arch/PKGBUILD staging/arch
cd staging/arch
mksrcinfo
git add PKGBUILD .SRCINFO
git commit -m "Release $TRAVIS_TAG"
git push origin master
```

For generating [`.SRCINFO`](https://wiki.archlinux.org/index.php/.SRCINFO) we can either use [`makepkg`](https://www.archlinux.org/packages/?name=pacman) or probably easier [`pkgbuild-introspection`](https://www.archlinux.org/packages/?name=pkgbuild-introspection) (only depends on bash).

We would also need to add a private key for deploying. This looks like another area where Travis is not optimal but we could always try to store it in an environment variable and then save it to a proper file in the script.

pkgw:
I will leave it to you two (and other Arch Linux users) to think about how you'd like to do this.

Am I understanding correctly that in this model, every push to `master` would result in an update of the AUR git repo and a new version appearing on the AUR? Is that standard practice? If it is also common to only push updates for new tagged upstream releases, I might prefer that model — life as a maintainer is a bit easier if your users aren't running ever-changing unreleased versions of the package.

jtojnar:
In AUR, there are two kinds of packages: Normal packages that get updated when a new version is tagged, and VCS packages that always fetch the current `HEAD` and need to be updated only when dependencies change. The `tectonic` package is the former one, since it lacks `-git` suffix.

[...]

PR #67, opened 2017-06-04
Deploy releases to AUR automatically

jtojnar:
This uses Travis’s automatic deploy feature. Everytime a new version is tagged, the PKGBUILD file along with updated metadata will be uploaded to AUR.

Decrypting the private key was moved to the `before_deploy` stage, to limit the possibility of the key leaking.

Closes: #54

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,6 @@ matrix:
     - rust: nightly
 
 before_install:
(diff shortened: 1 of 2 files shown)


pkgw:
Great! Looks reasonable to me.

PR #71, opened 2017-06-05
.travis.yml: new encryption stuff cannot run on PRs

pkgw:
For security reasons, random PRs don't get access to our encrypted files and variables, so the `openssl` invocation that decrypts the SSH key has to be gated on whether the Travis run is *not* associated with a PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ before_install:
   - brew update
 
 install:
+  - false
   - brew install graphite2
   - brew install harfbuzz --with-graphite2
   - brew install --force openssl


pkgw:
Never mind, superseded by #67.

jtojnar:
Actually, what prevents an attacker from editing the travisfile to dump the '$encrypted_*' variables?

pkgw:
See [here](https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions) — the variables aren't exported at all unless the PR originates in the project repo, which means that the PR must come from someone affiliated with the project.

PR #72, opened 2017-06-05
Fix a typo introduced in c7b79e6

rekka:


pkgw:
I knew something like that would creep in at some point :-(

Good catch!

Issue #23, opened 2017-05-22
Better Rust style for constructors


pkgw:
In Rust, if you have structs that can be created with a lot of different options, the recommended pattern is to use a series of functions in a factory object, for both forward-compatibility and clarity. See, for instance, [std::fs::OpenOptions](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html).

Several structs in Tectonic should be constructed in this way, but I've done things in an old-fashioned way with single functions that take lots of arguments. Examples include `tectonic::io::Filesystem`, `tectonic::io::LocalCache`, `tectonic::engines::TexEngine`, and so on. These should be refactored to the better, Rust-y style.

pkgw:
Um, let's say that I've dealt with this one.

PR #73, opened 2017-06-05
Add the --synctex command line option

rekka:
As per the discussion in #56, this adds the `--synctex` command line option that activates SyncTeX. The data is saved gzipped.

The 2nd commit implements a builder pattern for `TexEngine`: when adding a new toggle `synctex`, I can as well address #23.

Unresolved:

- disable `--synctex` when `--outfmt` is not `pdf`?


src/engines/tex.rs
```diff
@@ -21,29 +21,39 @@ pub enum TexResult {
...
+    halt_on_error: bool,
+    initex_mode: bool,
+    synctex_enabled: bool,
 }
 
 
 impl TexEngine {
     pub fn new () -> TexEngine {
-        TexEngine {}
+        TexEngine::default()
```
pkgw: I think this will default `halt_on_error` mode to be false, correct? I want it to default to true. Is there a way to override the default for just that one field?

If not, we could always invert its meaning (`continue_on_error` or something).

rekka:
Thanks for the review!

It's easy to implement `Default` manually, and initialize `halt_on_error` to `true`. How about that?

pkgw:
Sure, that'd be great.

rekka:
OK, `halt_on_error` is now `true` by default.

I hit a problem with testing gzipped files---seems that the output is platform dependent. Now the test decompresses the files before comparing them.

It would be nice to run Travis tests on linux as well to catch things like this in the future.

pkgw:
Awesome work! I'll merge as soon as the CI passes.

(Thus far, multi-platform testing hasn't been a problem since I do all of my development on Linux, but now that other folks are contributing it is time to professionalize the infrastructure a bit more ...)

PR #77, opened 2017-06-07
Fix #50: Normalize paths in MemoryIo

### LINKED ISSUE #50: pdfpages cannot include paths with a `.`
I'm using the `pdfpages` package to include other pdf files. It seems like Tectonic can't handle paths with a `.` or `..` in them, while pdflatex can. Quite minimal example:

some.tex:
```tex
\documentclass[]{article}
\begin{document}
test
\end{document}
```

test.tex:
```tex
\documentclass[]{article}
\usepackage{pdfpages}
\begin{document}
\includepdf[pages={-}]{./some.pdf}
\end{document}
```

In this example, the `some.pdf` is in the same directory as the test.tex. Note that using path `some.pdf` instead of `./some.pdf` will work. I get the similar exceptions when:
- I move `some.pdf` in the directory above and use `../some.pdf` to include it.
- I create a `subdir` and use `subdir/../some.pdf`.

Exception:
```
Running TeX ...
error: something bad happened inside TeX; its output follows:

(test.tex
LaTeX2e <2016/03/31>
Babel <3.9r> and hyphenation patterns for 83 language(s) loaded.
(article.cls
Document Class: article 2014/09/29 v1.4h Standard LaTeX document class
(size10.clo)) (pdfpages.sty (ifthen.sty) (calc.sty) (eso-pic.sty (atbegshi.sty
(infwarerr.sty) (ltxcmds.sty) (ifpdf.sty)) (keyval.sty) (xcolor.sty (color.cfg)
(xetex.def))) (graphicx.sty (graphics.sty (trig.sty) (graphics.cfg)))
(ppxetex.def)) (test.aux) (pdflscape.sty (lscape.sty) (ifxetex.sty))
! Unable to load picture or PDF file './some.pdf'.
<to be read again> 
                   }
l.6 \includepdf[pages={-}]{./some.pdf}
                                      
No pages of output.
Transcript written on test.log.

error: halted on potentially-recoverable error as specified
```

rekka:
- Only paths that are valid UTF8 are normalized.
  - `std::path::{Path,PathBuf}` are not used to normalize in a system
    independent way on Unix/Windows.
  - LaTeX identifies OS by testing if `./texsys.aux` exists. This file
    is created as `texsys.aux`. Without normalization, OS is
    misidentified and filename parsing in LaTeX code then produces wrong
    results. (See #50). This fixes the behavior (cache needs to be
    rebuilt).

Note: Normalization currently does not handle path prefix on Windows, but this can be added later (paths on Windows are hard, need to explore what is reasonable).

### DIFF
--- a/src/io/memory.rs
+++ b/src/io/memory.rs
@@ -11,7 +11,7 @@ use std::rc::Rc;
 use errors::Result;
 use status::StatusBackend;
 use super::{InputFeatures, InputHandle, InputOrigin, IoProvider, OpenResult, OutputHandle,
-            try_normalize_path};
+            normalize_tex_path};
 
 
 // MemoryIo is an IoProvider that stores "files" in in-memory buffers.
@@ -118,7 +118,7 @@ impl IoProvider for MemoryIo {
     fn output_open_name(&mut self, name: &OsStr) -> OpenResult<OutputHandle> {
         assert!(name.len() > 0, "name must be non-empty");
 
-        let name = try_normalize_path(name);
+        let name = normalize_tex_path(name);
 
         OpenResult::Ok(OutputHandle::new(&name, MemoryIoItem::new(&self.files, &name, true)))
     }
@@ -134,7 +134,7 @@ impl IoProvider for MemoryIo {
     fn input_open_name(&mut self, name: &OsStr, _status: &mut StatusBackend) -> OpenResult<InputHandle> {
         assert!(name.len() > 0, "name must be non-empty");
 
-        let name = try_normalize_path(name);
+        let name = normalize_tex_path(name);
 
         if self.files.borrow().contains_key(&*name) {
             OpenResult::Ok(InputHandle::new(&name,
--- a/src/io/memory.rs
+++ b/src/io/memory.rs
@@ -10,7 +10,8 @@ use std::rc::Rc;
 
 use errors::Result;
 use status::StatusBackend;
-use super::{InputFeatures, InputHandle, InputOrigin, IoProvider, OpenResult, OutputHandle};
+use super::{InputFeatures, InputHandle, InputOrigin, IoProvider, OpenResult, OutputHandle,
+            try_normalize_path};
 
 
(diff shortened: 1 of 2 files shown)


pkgw:
Brilliant! The code looks great, only a few high-level thoughts:

- I think `normalize_path` should be called `try_normalize_tex_path`. A function named `try_...` should return an Option or Result, I think, so I'd prefer to swap the nomenclature there. And then I think we want to establish a terminology of a "TeX path" being one that we force to obey our simplified semantics: Unix-like syntax (`/` for separators etc), must be Unicode-able, no symlinks allowed such that `..` can be stripped lexically.

- Then I guess `try_normalize_path` should become `normalize_tex_path`, and documentation string should note that the function should operate on `&str` someday, but we need to transition the internals away from `OsStr{,ing}` before that can happen.

- Can you put in some tests for paths containing two consecutive slashes, like `abc//xyz`?

rekka:
Thanks so much for the review and comments. 

Hopefully this commit addresses all of them. 
I used a few of them verbatim in the code.

I was sure that I had `//` tests, nice catch---I sprinkled a few of them around.

pkgw:
Great, I'll merge this momentarily! Three random thoughts:

- I believe that if you write a freestanding line in the commit message that says `Closes #50.`, GitHub will automatically close issue #50 when the commit is merged into master.

- I believe that it's reliable that `/..` = `/`, so the code path that returns  `None` in `try_...` could potentially be removed. But that's not a big deal.

- In fact I'm not a huge fan of letting people write absolute paths at all in their TeX code ... but this seems like an area where we should probably be pragmatic, since I can imagine people that just want to get their damn file to compile and they (feel that they) need to reference some file with an absolute path to get it to work. Something to ponder.

rekka:
Thanks! Now I should be able to use tectonic in my workflow.

- Thanks for the tip.

- You're are right, it seems that `/..` = `/`. Paths are hard...

- Absolute paths should be allowed. Maybe a warning is fine, but that also might quickly get quite annoying.

Issue #32, opened 2017-05-29
Support Windows natively


chris-morgan:
I would *love* to use this for a project that I’ve been stalled on for quite some time, because deciding on a good layout/text rendering stack was too hard and rolling my own was taking too long. If it looks like this can really suit my needs I’ll become a contributor to this project and be forever grateful to you for making it in the first place!

Windows support is an essential for this project of mine.

```
   Compiling mkstemp-rs v0.0.2
error[E0432]: unresolved import `std::os::unix::io::FromRawFd`
  --> C:\Users\me\.cargo\registry\src\github.com-1ecc6299db9ec823\mkstemp-rs-0.0.2\src\lib.rs:17:5
   |
17 | use std::os::unix::io::FromRawFd;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Could not find `unix` in `os`

error[E0425]: cannot find function `mkstemp` in module `libc`
  --> C:\Users\me\.cargo\registry\src\github.com-1ecc6299db9ec823\mkstemp-rs-0.0.2\src\lib.rs:37:33
   |
37 |         let fd = unsafe { libc::mkstemp(ptr) };
   |                                 ^^^^^^^ not found in `libc`

error: no associated item named `from_raw_fd` found for type `std::fs::File` in the current scope
  --> C:\Users\me\.cargo\registry\src\github.com-1ecc6299db9ec823\mkstemp-rs-0.0.2\src\lib.rs:44:29
   |
44 |         let file = unsafe { File::from_raw_fd(fd) };
   |                             ^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile `mkstemp-rs`.
```

I’ll look into this myself in around eight hours’ time if you don’t earlier.

pkgw:
For context, I have not been planning to work on Windows support myself — there’s just too much to do. I would *love* if you and/or others could work on this, though!

I have to say that I expect that supporting Windows will be difficult. A lot of the code *should* be portable, but there are several parts that I *know* rely on Unix-y semantics in various ways, and I'm sure there are many parts that don't. And there are the various dependencies on Unix-y font and image libraries.

I very much want to support your investigations — please get in touch with questions, etc.

MikeChou:
Windows Subsytem for Linux is quite usable now. I suggest anyone who wants to use this on Windows to try that first. If it works it'll save a lot of work adding Windows(win32) support.

tawalaya:
[...]

PR #80, opened 2017-06-08
WIP: Windows (MSVC) support

chris-morgan:
Goal: get Tectonic to compile and run on Windows (#32).

I’m using MSVC; it could well be easier on MinGW but I’m not interested in MinGW. Some of the changes my inadvertently help MinGW.

Current status:

- XeTeX and fontconfig code required (for header files);
- Prerelease version of harfbuzz-sys crate required;
- It’s failing in linking (I haven’t made any attempt to resolve this yet—mostly I think it’ll be a matter of bringing along prebuilt binaries of the various dependencies).


tectonic/msvc-include/unistd.h
```diff
@@ -0,0 +1,2 @@
+// For now, this is a total stub.
+// See https://stackoverflow.com/q/341817 for some things we could want.
```
pkgw: I know essentially nothing about Windows development — is stubbing headers like this a common practice? Feels gross, but you do what you gotta do ...

.gitignore
```diff
@@ -1,2 +1,3 @@
+/Cargo.lock
```
pkgw: So, my thinking to date has been that for most people, the main deliverable is the `tectonic` command line program, not the crate, so that keeping the `Cargo.lock` in version control is the way to go.

I'm not sure the best approach. The reusable crate could just be separated out into a sub-project with its own `Cargo.toml`, of course. But I got the impression that it's not an uncommon pattern for Rust tools to have a single `Cargo.toml` for a featureful crate plus a command-line interface to it ... I should do some digging to figure out the best practices, or ask on [the Rust forums](https://users.rust-lang.org/).

tectonic/msvc-include/unistd.h
chris-morgan: I’d guess it’d be more common to use a HAVE_UNISTD_H flag which is undefined on Windows, but I’m not really sure. I’m not a C/C++ developer and cannot speak for their conventions!

tectonic/xetex-core.h
```diff
@@ -87,6 +87,10 @@ typedef struct XeTeXLayoutEngine_rec* XeTeXLayoutEngine;
 #define FONT_FLAGS_COLORED  0x01
 #define FONT_FLAGS_VERTICAL 0x02
 
+/* Windows? */
+// This one is in xetex-code/source/texk/web2c/xetexdir/xetex.h, but not here? Baffled as to why not.
+#define M_PI 3.14159265358979323846
```
ronnychevalier: It seems that for Windows you need to \#define _USE_MATH_DEFINES` before including `math.h`. So you do not need to redefine it yourself.

See https://msdn.microsoft.com/en-us/library/4hwaceh6.aspx

tectonic/tectonic.h
```diff
@@ -37,7 +37,12 @@
 
 /* basic types */
 
+#if defined(_MSC_VER)
+// On MSVC it’s already defined if rpcndr.h has been loaded (which it has for some of our files), as this:
+typedef unsigned char boolean;
```
ronnychevalier: The `boolean` typedef has been [removed](https://github.com/tectonic-typesetting/tectonic/commit/f0ec7c6f994b02274d9ee1cd5a08d9cb10a3dbb5). So I guess this change does not make sense now.

tectonic/internals.h
```diff
@@ -136,7 +136,7 @@ void set_input_file_encoding(UFILE *f, integer mode, integer encodingData);
 void u_close(UFILE *f);
 int u_open_in(UFILE **f, integer filefmt, const char* fopen_mode, integer mode, integer encodingData);
 int get_uni_c(UFILE* f);
-int input_line(UFILE* f);
+boolean input_line(UFILE* f);
```
ronnychevalier: The `boolean` typedef has been [removed](https://github.com/tectonic-typesetting/tectonic/commit/f0ec7c6f994b02274d9ee1cd5a08d9cb10a3dbb5). You should change the `int` to `bool` instead.

tectonic/io.c
```diff
@@ -220,7 +220,7 @@ apply_normalization(uint32_t* buf, int len, int norm)
 }
 
 
-int
+boolean
```
ronnychevalier: Same here: `int` -> `bool` instead of `boolean`

tectonic/msvc-include/unistd.h
ronnychevalier: Since there is only 3 includes using `unistd.h`, it should be simpler and cleaner to just put an `#ifndef _MSVC` before such include?

I guess having something that provides the `HAVE_UNISTD_H` from gcc-rs would be great in the future.

chris-morgan:
@ronnychevalier I’ve rebased it with the necessary adjustments, so it’s back to failing in linking.

I’ll continue to work on it as I allot time, and prepare a branch for near-future merging.

pkgw:
@chris-morgan Yeah, sorry, there's been a lot of churn aiming to tidy up the C code that is probably giving you a moving target to aim at. If you think you're close to getting things working we can dial that back a bit if that would be helpful.

tectonic/msvc-include/getopt.h
```diff
@@ -0,0 +1,655 @@
+// Taken from https://github.com/skandhurkat/Getopt-for-Visual-Studio/blob/master/getopt.h
+
```
ronnychevalier: We do not use `getopt` or `getopt_long` anymore. Could you instead remove the `#include <getopt.h>` from `internals.h`? The include is useless now.

tectonic/dpx-dpxfile.c
```diff
@@ -19,7 +19,9 @@
 */
 
 #include <time.h>
+#if HAVE_UNISTD_H
```
ronnychevalier: Use `#ifdef` instead. It is not done everywhere but we should avoid adding more.

tectonic/dpx-pdfencrypt.c
```diff
@@ -23,7 +23,9 @@
 #include <string.h>
 #include <time.h>
 
+#if HAVE_UNISTD_H
```
ronnychevalier: Same use `#ifdef` instead of `#if` in case `HAVE_UNISTD_H` is not defined.

tectonic/internals.h
```diff
@@ -31,7 +31,9 @@
 #include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#if HAVE_UNISTD_H
```
ronnychevalier: Same use `#ifdef` instead of `#if`

.gitignore
h4llow3En: As far as I know, the Cargo.lock should be there for executables to ensure the same versions of the dependencies and avoid build failures.

crlf0710:
news?

pkgw:
Superseded by @crlf0710's work in [#210](https://github.com/tectonic-typesetting/tectonic/pull/210). MSVC still not supported (only GNU toolchain) and we're still not passing tests, but that's the first step.

PR #83, opened 2017-06-08
More warning fixes and some cleanups

ronnychevalier:


tectonic/dpx-pdfparse.c
```diff
@@ -387,7 +387,7 @@ parse_pdf_boolean (const char **pp, const char *endptr)
     }
   }
 
-  dpx_warning("Not a boolean object.");
+  dpx_warning("Not a bool object.");
```
pkgw: Tiny nit, but I think this one should be kept as-is.

tectonic/dpx-spc_tpic.c
```diff
@@ -76,7 +76,7 @@ struct spc_tpic_
 
   /* state */
   double     pen_size;
-  int        fill_shape; /* boolean */
+  int        fill_shape; /* bool */
```
pkgw: Here too ... or turn it into an actual `bool` if convenient?

tectonic/dpx-dvipdfmx.c
```diff
@@ -757,10 +713,6 @@ dvipdfmx_main (int argc, char *argv[])
   pdf_load_fontmap_file("kanjix.map", FONTMAP_RMODE_APPEND);
   pdf_load_fontmap_file("ckx.map", FONTMAP_RMODE_APPEND);
 
-  /* End config file fakery. */
-
-  do_args (argc, argv, NULL, 0);
```
pkgw: Can the `do_args` function now be removed? I'm not seeing it show up in the diff.

tectonic/dpx-pdfparse.c
ronnychevalier: Fixed

tectonic/dpx-spc_tpic.c
ronnychevalier: Fixed. I added an additional commit to change two variables (including this one) from int to bool in this file.

tectonic/dpx-dvipdfmx.c
ronnychevalier: Not a the moment unfortunately.`read_config_special` calls `do_args`. This function is used in `dpx-dvi.c`, it seems like it is possible to put additional command arguments in the dvi itself, which is then processed and it calls the `do_args` eventually again.

I am not sure if such feature is really used, so I took the safe choice to keep it. But we could investigate a bit more to know whether or not we could remove all of this.

ronnychevalier:
Fixed. I also added 2 additional fixes.

pkgw:
Looks good!

PR #84, opened 2017-06-08
tests: write executable and regression testing suite

alexander-bauer:
The `tests/executable.rs` module includes tools for running the built
executable on a temporary directory, and for copying resources into the
temporary directory easily.

This will more easily let us reproduce reported issues, and test for
them automatically to prevent regressions.

This includes a regression test for #36

### DIFF
--- a/tests/executable/test space.tex
+++ b/tests/executable/test space.tex
@@ -1,4 +1,3 @@
-\documentclass{article}
-\begin{document}
 This file has a name which contains spaces.
-\end{document}
+
+\bye
--- a/tests/executable/test space.tex
+++ b/tests/executable/test space.tex
@@ -1,0 +1,4 @@
+\documentclass{article}
+\begin{document}
+This file has a name which contains spaces.
+\end{document}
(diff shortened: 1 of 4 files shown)


tests/executable/test space.tex
```diff
@@ -0,0 +1,4 @@
+\documentclass{article}
```
pkgw: i would prefer if this test stuck to plain TeX, rather than LaTeX, so as not to require a full download of all of the LaTeX support files to run it. In principle it could even pass a `--format` argument to `tectonic` so that it could use the plain format file generated in the `tex-outputs` test, which can be generated without touching the network at all. But I am OK with a version that needs to touch the network — which will still need a `--format` flag to change the standalone executable away from its default of using the `latex` format.

alexander-bauer: I must admit, I don't really understand enough about plain TeX to be able to make any intelligent progress on this. It would be nice to use the already-existing format files, but I'm not sure if we can make guarantees about whether they'll be generated already. What do you suggest?

pkgw: The `the_letter_a` test case is just about the simplest TeX document you can write when using the plain format:

```
a\bye
```

You should be able to use that as the contents of this file if you make sure to pass the argument `--format=plain.fmt.gz` to the `tectonic` executable.

This won't reliably reuse the format generated in-tree, but I'm OK with that for now.

alexander-bauer:
@pkgw I think these two changes address the comments you made; let me know if there's anything else to be done.

pkgw:
Great!

By the way, I have just launched a [Tectonic forums](https://tectonic.newton.cx/) site for discussions that don't fit well into the "issues" framework. You should have received an email from `noreply@tectonic.newton.cx` inviting you to create an account; I hope you'll give it a try!

Issue #31, opened 2017-05-29
Figure out how to deal with input documents outside of the current directory


Ohjeah:
Assuming this structure

```
|_tex
   |_ main.tex
   |_ content
      |_ 1.tex
```

and `main.tex` contains `\include{content/1.tex}` you need to call tectonic from within the tex directory.
If you call it like `tectonic tex/main.tex` you get an error:

> ! LaTeX Error: File `misc/header.tex' not found.

Would be nice to have an `--inplace` option which always sets the working directory to the directory of your main.tex, i.e. `tectonic tex/main.tex --inplace` would be valid.

pkgw:
Ah, good catch. As much as possible I would like to have Tectonic just Do The Right Thing™ without needing special options, so I will think about whether there's a way to get it to work without needing an option. At first blush maybe it would be reasonable to *always* act as if the proposed `--inplace` was active?

Ohjeah:
Yes that would be perfect. I cannot think of a case were you'd want it other way around.

pkgw:
CC @skrat.

skrat:
@pkgw did you find the codes that assume CWD = parentDir(root.tex) ?

rekka:
Is anyone working on this?

PR #86, opened 2017-06-10
Reproduce #31

alexander-bauer:
This PR adds a failing test case for issue #31. The test case is prefixed with a reasonably specific `should_panic` decorator, and is noted with a `// FIXME` referring to issue #31. When that issue is fixed, this test case will begin to fail (because it will no longer panic) and the directive can be removed.

I would prefer that this test didn't simply succeed, but gave an "expected failure"-type annotation in the results, like PyTest does. I don't know that this behavior is possible in the current state of Cargo testing.

### DIFF
--- a/tests/executable.rs
+++ b/tests/executable.rs
@@ -95,7 +95,7 @@ fn help_flag() {
 }
 
 #[test]
-#[should_panic(expected = "failed to open input file \"content/1.tex\"")] // FIXME: GitHub #31
+#[ignore] // FIXME: GitHub #31
 fn relative_include() {
     let tempdir = setup_and_copy_files(&["subdirectory/relative_include.tex",
                                          "subdirectory/content/1.tex"]);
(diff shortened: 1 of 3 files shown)


pkgw:
Very nice!

It seems that the Rust recommended practice is to [ignore](https://doc.rust-lang.org/book/second-edition/ch11-02-running-tests.html#ignore-some-tests-unless-specifically-requested) tests that are expected to fail; [see brief discussion here](https://github.com/rust-lang/rust/issues/11505). I guess I feel that ignore is a bit better than should_panic, since the latter should ideally be reserved for cases where we *want* the code to panic, not just that it *does* panic right now.

alexander-bauer:
@pkgw I've just changed the test case to be `ignore`d. I like that the test output notes that it was ignored, but it'd be really nice if we could provide a reason. ("Expected failure; see #31"). It doesn't look like that exists, though.

pkgw:
Yeah ... I tend to think that a comment explaining why the test is ignored, like you've done here, probably pretty much does the job most times. But I do see the appeal of having that meta-data be expressed in a somewhat more structured way.

Anyway, thanks, nice respectable development methodology!

Issue #81, opened 2017-06-08
Something bad happened inside BibTeX: you've exceeded BibTeX's number of strings 4000


pfasante:
For my daily work I need to use a bibtex collection ([github/cryptobib](https://github.com/cryptobib/export)) with several thousand entries. While the following code
```
\documentclass{article}
\begin{document}
\cite{ToSC:KraLeaWie17}
\bibliographystyle{alpha}
\bibliography{abbrev0,crypto}
\end{document}
```
gets build correctly with `latexmk -pdf`, tectonic aborts during the bibtex run with the following error:

```
Running BibTeX ...
error: something bad happened inside BibTeX; its output follows:

===============================================================================
Sorry---you've exceeded BibTeX's number of strings 4000
Aborted at line 2083 of file abbrev0.bib
(That was a fatal error)
===============================================================================

error: unspecified fatal bibtex error
```

pfasante:
sorry, I oversaw #79, which covers the same problem

PR #87, opened 2017-06-10
update: Update MAX_STRINGS value

### LINKED ISSUE #79: Handling large bib files
By default, it seems that tectonic does not handle well large bib files.

```
Running BibTeX ...
error: something bad happened inside BibTeX; its output follows:

Sorry---you've exceeded BibTeX's number of strings 4000
Aborted at line 1707 of file cryptobib/abbrev3.bib
(That was a fatal error)

error: unspecified fatal bibtex error
```

The bib files are [cryptobib files](https://github.com/cryptobib/)

mrshu:
* Set the value of MAX_STRINGS to 15000. It turns out 10000 was not
  enough to cleanly build the example mentioned in #81 using
  (https://github.com/cryptobib/export_crossref/blob/118d812c4e1e8c462d3b1f226c02b9b1c6e4b505/abbrev0.bib).

* Fixes #79

### DIFF
--- a/tectonic/bibtex.c
+++ b/tectonic/bibtex.c
@@ -137,7 +137,7 @@ static jmp_buf error_jmpbuf, recover_jmpbuf;
 #define aux_stack_size 20
 #define MAX_BIB_FILES 20
 #define POOL_SIZE 65000L
-#define MAX_STRINGS 4000
+#define MAX_STRINGS 15000
 #define MAX_CITES 750
 #define WIZ_FN_SPACE 3000
 #define SINGLE_FN_SPACE 50


mrshu:
@pkgw I am not sure if this is the correct way of fixing this, but it seems that the aforementioned file is actually an edge case (with regards to the number of entries). The code in this PR was able to generate a PDF from the example mentioned in #81.

Please feel free to point me to a better way of dealing with this.

Thanks!

pkgw:
@mrshu If you've gone to the work to get the test case to work, I think what you've done must be exactly right! Thanks for the contribution.

I'm trying to get discussions of the more user-facing side of Tectonic going at [its new forum site](https://tectonic.newton.cx/); please sign up if you're interested!

mrshu:
Thanks for the invitation @pkgw, will do!

Issue #51, opened 2017-06-01
Clearer error messages on network failure


mikeger:
# Issue

Cannot use tectonic to process the tex files.

# Technical info

Machine information:
- MacBook Pro Retina 2013.
- MacOS Sierra 10.12.5 (16F73).
- zsh 5.2.
- TexLive installed.

# Steps to reproduce:
1. Install miniconda
2. Do `export PATH="/Users/user/miniconda3/bin:$PATH"`
3. Run "tectonic some-tex-file.tex"

Expected:
Parsed the tex file and produce a pdf document.

Result:
Failure:

```
tectonic file.tex
note: this is a BETA release; report issues at https://github.com/tectonic-typesetting/tectonic/issues
note: indexing http://purl.org/net/pkgwpub/tectonic-default
note: resolved to https://dl.bintray.com/pkgw/tectonic/tl2016extras/2016.0r4/tlextras-2016.0r4.tar
note: downloading SHA256SUM
warning: failure requesting "SHA256SUM" from network
caused by: Connection closed
caused by: Connection closed
warning: failure requesting "SHA256SUM" from network
caused by: Connection closed
caused by: Connection closed
warning: failure requesting "SHA256SUM" from network
caused by: Connection closed
caused by: Connection closed
warning: failure requesting "SHA256SUM" from network
caused by: Connection closed
caused by: Connection closed
error: failed to retrieve "SHA256SUM" from the network
```

pkgw:
This means that your computer had some problem downloading the URL noted before the warnings: https://dl.bintray.com/pkgw/tectonic/tl2016extras/2016.0r4/tlextras-2016.0r4.tar . Which means I think it is very likely that this was some problem with your computer's network at the time, not with Tectonic.

But, if your computer got past the `indexing` step, that means it was able to fetch a companion file. Hmmm.

Does it work if you just try again at at time when your computer has a nice reliable internet connection?

mikeger:
@pkgw OK, now it can download the necessary resources. Thanks for the response!

pkgw:
Well, I'll leave this one open with a different spin: in these cases we should make it clear that the problem is (probably) not Tectonic's fault. I can see how the current message is a bit ambiguous about that.

mikeger:
I shouldn't have closed it, I see the issue again:

```
# tectonic whitepaper.tex
note: this is a BETA release; report issues at https://github.com/tectonic-typesetting/tectonic/issues
note: generating format "latex"
[...]

PR #88, opened 2017-06-10
update: Clearer message on network failure

mrshu:
* Add clearer message which clearly mention that the issues are not
  caused by Tectonic, as per #51.


pkgw:
I like the phrasing. Thanks!

Issue #74, opened 2017-06-06
A more robust manifest format


rekka:
Currently the manifest format for the local cache is stored in a space-separated format without any escaping. This leads to invalid entries being inserted if for instance files with spaces are requested:

```tex
\input{"with space.tex"}\bye
```

produces entries

```
with space.tex 0 -
with space.tex.tex 0 -
```

It is not really a serious issue at the moment, but a future-proofing of the format is desirable.

I believe that one appropriate solution would be to use the [csv crate](https://crates.io/crates/csv), which also allows for [space-separated entries](https://docs.rs/csv/1.0.0-beta.3/csv/struct.ReaderBuilder.html#method.delimiter). So it should be backwards compatible with the current format. It also offers easy parsing using `serde`. Moreover, it will  most likely hit version 1 soon.

I am happy to implement this.

pkgw:
Yes, it would be good to tighten things up here.

I might prefer to stick with the current format but just reverse the ordering: size, SHA256, filename. That way you can just split on the first two spaces and take the rest as a lump.

Although I guess we're still fragile if anyone uses a filename containing a newline ...

rekka:
That is actually a pretty good solution too. Now files with newlines...

alexander-bauer:
I was just browsing through looking at this, and poked around until I found the relevant parts of the code. It looks like [`src/io/local_cache:176`](https://github.com/tectonic-typesetting/tectonic/blob/master/src/io/local_cache.rs#L176) is where it gets written; I'm not sure where it gets read from the manifest.

rekka:
It is a bit above that, [in the `new` method](https://github.com/tectonic-typesetting/tectonic/blob/c85e179e3f22a07d9c01866b3cc807ce05dd3a04/src/io/local_cache.rs#L114-L122).

alexander-bauer:
@rekka I can go ahead and make a PR for this, but I'd also hate to duplicate work, if you'd like to

rekka:
I'm not currently working on this, but I think we should reach a consensus with @pkgw on what the best approach here is.

I am not leaning towards a simple format as suggested by @pkgw with: size SHA256 filename, and completely disallowing files with new lines to enter the manifest. This can be easily parsed by `str::splitn` method.

pkgw:
@rekka Do you perhaps mean "now" and not "not" in your most recent comment?

[...]

PR #89, opened 2017-06-11
A more robust parsing of the manifest

rekka:
This partially addresses #74 without actually changing the manifest format.

Now the lines in the manifest are split from the end (`rsplitn`) into exactly three pieces and only on
spaces (' ').

This allows filenames to have spaces in their names.

A sanity check has been added to not insert files with newline characters in their names. (I guess to prevent the potential to doctor a manifest line by crafting a file name).

Question: Is it worth inserting files with names that are not valid unicode, i.e. [use `to_string_lossy`](https://github.com/rekka/tectonic/blob/82269bbab803450b9d862b0e677f386a7fad2221/src/io/local_cache.rs#L179)? This record is then useless anyway because it will not match the original file.

### DIFF
--- a/src/io/local_cache.rs
+++ b/src/io/local_cache.rs
@@ -176,10 +176,10 @@ impl<B: IoProvider> LocalCache<B> {
         // Lock will be released when file is closed at the end of this function.
         ctry!(man.lock_exclusive(); "failed to lock manifest file \"{}\" for writing", self.manifest_path.display());
 
-        let name_utf8 = name.to_string_lossy();
-
-        if !name_utf8.contains(|c| c == '\n' || c == '\r') {
-            writeln!(man, "{} {} {}", name_utf8, length, digest_text)?;
+        if let Some(name_utf8) = name.to_str() {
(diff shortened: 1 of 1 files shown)


pkgw:
Nice! Yes, if we're trying to be thorough here, I guess we should also not try to record filenames that are invalid UTF-8, although I don't believe that TeX programs are able to construct such filenames.

rekka:
Thanks for having a look. Honestly, I have not been able to figure out how to construct a filename with a newline either to test the check.

pkgw:
I think it might be possible with a `^^J` escape or something, but I am OK with not having a test to cover that particular codepath.

If you can add a change that checks the result of `to_string_lossy` in order to infer whether the path was valid UTF-8, I'd call this good to merge.

rekka:
OK, done. Thanks!

`^^J` indeed works, that's a neat trick. And the file does not get inserted, as expected.

pkgw:
Thanks for the improvement!

PR #91, opened 2017-06-11
Fix #63: do not panic into C

### LINKED ISSUE #63: Do not panic into C/C++ code
It is [undefined behavior to panic from Rust into foreign code](https://doc.rust-lang.org/stable/std/panic/fn.catch_unwind.html) at the moment. Thus the bridge functions in `engines/mod.rs`, etc., should not panic.

At the moment, there are explicit `panic!` invocations for instance [here](https://github.com/tectonic-typesetting/tectonic/blob/328213eff03e652d4c675c2cb9902947bd67aa4f/src/engines/mod.rs#L386) and [here](https://github.com/tectonic-typesetting/tectonic/blob/328213eff03e652d4c675c2cb9902947bd67aa4f/src/engines/mod.rs#L581).

rekka:
Instead of panicking on failure, add a warning and return an error value
(if possible).

Closes #63.

However, these kinds of problems should probably result in a failure since they indicate an internal engine error.

### DIFF
--- a/src/engines/mod.rs
+++ b/src/engines/mod.rs
@@ -602,7 +602,7 @@ fn input_get_size<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, handle
     es.input_get_size(rhandle)
 }
 
-fn input_seek<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, handle: *mut libc::c_void, offset: libc::ssize_t, whence: libc::c_int) -> libc::size_t {
+fn input_seek<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, handle: *mut libc::c_void, offset: libc::ssize_t, whence: libc::c_int, internal_error: *mut libc::c_int) -> libc::size_t {
     let es = unsafe { &mut *es };
     let rhandle = handle as *mut InputHandle;
 
@@ -611,9 +611,11 @@ fn input_seek<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, handle: *m
         libc::SEEK_CUR => SeekFrom::Current(offset as i64),
         libc::SEEK_END => SeekFrom::End(offset as i64),
         _ => {
-                // TODO: Handle the error better. This indicates a bug in the engine.
                 tt_error!(es.status, "serious internal bug: unexpected \"whence\" parameter to fseek() wrapper: {}",
                             whence);
+                unsafe {
+                    *internal_error = 1;
+                }
                 return 0;
         }
     };
--- a/src/engines/mod.rs
+++ b/src/engines/mod.rs
@@ -468,21 +468,21 @@ extern {
 
 // Entry points for the C/C++ API functions.
 
-fn issue_warning<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, text: *const libc::c_char) {
+extern fn issue_warning<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, text: *const libc::c_char) {
     let es = unsafe { &mut *es };
     let rtext = unsafe { CStr::from_ptr(text) };
 
     tt_warning!(es.status, "{}", rtext.to_string_lossy());
 }
 
-fn issue_error<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, text: *const libc::c_char) {
+extern fn issue_error<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, text: *const libc::c_char) {
     let es = unsafe { &mut *es };
     let rtext = unsafe { CStr::from_ptr(text) };
 
     tt_error!(es.status, "{}", rtext.to_string_lossy());
 }
 
-fn get_file_md5<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, path: *const libc::c_char, digest: *mut u8) -> libc::c_int {
+extern fn get_file_md5<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, path: *const libc::c_char, digest: *mut u8) -> libc::c_int {
     let es = unsafe { &mut *es };
     let rpath = osstr_from_cstr(unsafe { CStr::from_ptr(path) });
     let rdest = unsafe { slice::from_raw_parts_mut(digest, 16) };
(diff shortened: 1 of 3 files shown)


src/engines/mod.rs
```diff
@@ -578,7 +580,11 @@ fn input_seek<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, handle: *m
         libc::SEEK_SET => SeekFrom::Start(offset as u64),
         libc::SEEK_CUR => SeekFrom::Current(offset as i64),
         libc::SEEK_END => SeekFrom::End(offset as i64),
-        _ => panic!("Unexpected \"whence\" parameter to fseek() wrapper: {}", whence),
+        _ => {
+                tt_warning!(es.status, "Unexpected \"whence\" parameter to fseek() wrapper: {}",
+                            whence);
+                return 0;
```
ronnychevalier: I think it would make more sense to change `input_seek` to return a signed value. `fseek` returns `-1` and set `errno` to `EINVAL` when there is an invalid `whence` provided. I think it would make sense to return `-EINVAL` (e.g., `ssize_t` is signed). A negative value so that the caller knows there is an error and its value set to an error to know the reason if needed.

rekka: Thanks for the suggestion. Unfortunately, as far as can tell, the semantics of `input_seek` are not clear. It is not really a wrapper around `fseek`, so that message is a bit misleading. Looking at the history, it was used to replace various methods, mostly [`rewind`](http://www.cplusplus.com/reference/cstdio/rewind/), which does not report errors. In fact, the code does not check the return value, except at two places in `dpx-pdfobj.c` like [this inefficient looking search for the beginning of the line](https://github.com/rekka/tectonic/blob/829a9647434579642dbd6c7d3962b8c52f20e2b6/tectonic/dpx-pdfobj.c#L2672-L2678). There it means the current position. I agree that it would be definitely better to abort the engine and report a proper error. I think that it will need a more serious refactoring of the bridge code. This commit at least fixes the undefined behavior.

ronnychevalier: Sure, maybe add a TODO or an issue to keep track of that? We should handle normal errors from seek and also the case of an unknown `whence` parameter.

rekka: Good point. I have also moved the error conversion from `ExecutionState::input_seek` to `input_seek`. I believe it is better to keep the conversion "Rust error" -> "C error" all in one place in the bridge function.

src/engines/mod.rs
```diff
@@ -383,7 +377,9 @@ impl<'a, I: 'a + IoProvider> ExecutionState<'a, I> {
             }
         }
 
-        panic!("unexpected handle {:?}", handle);
+        tt_warning!(self.status, "input close failed: unexpected handle {:?}", handle);
```
pkgw: Can you add statement making it clearer that this is an internal bug? Something like "serious internal bug: unexpected handle in input close".

src/engines/mod.rs
```diff
@@ -578,10 +574,22 @@ fn input_seek<'a, I: 'a + IoProvider>(es: *mut ExecutionState<'a, I>, handle: *m
         libc::SEEK_SET => SeekFrom::Start(offset as u64),
         libc::SEEK_CUR => SeekFrom::Current(offset as i64),
         libc::SEEK_END => SeekFrom::End(offset as i64),
-        _ => panic!("Unexpected \"whence\" parameter to fseek() wrapper: {}", whence),
+        _ => {
+                // TODO: Handle the error better. This indicates a bug in the engine.
+                tt_warning!(es.status, "Unexpected \"whence\" parameter to fseek() wrapper: {}",
```
pkgw: Very minor, but while we're here, I'm trying to be more consistent about these sorts of messages not starting with capital letters.

src/engines/mod.rs
rekka: Do you want me to add "serious internal bug" only to this particular error, or to others as well?

src/engines/mod.rs
rekka: This is also an internal error, maybe also add "serious internal error"?

rekka:
Thanks for the review. Should I change just the `tt_warning`s that this commit touches, or should I do it with the other `tt_warning`s throughout the file?

pkgw:
I'll trust your judgment on both questions. For things like the "whence" problem, I think it is good to somehow try to convey to the user that the program has some crazy problem and they will almost certainly not be able to fix it themselves. In fact, it might be smart to add some infrastructure for "internal bug" type problems that automatically produces a message to this effect, telling the user to visit the issue tracker.

As for `tt_warning` to `tt_error` change, the distinction that I try to maintain is whether the output file is probably going to be fine, or it's probably not going to be fine. For these kinds of problems that represent a failure of internal consistency, I think it is better to assume that the output will probably not be fine ⇒ those sorts of problems should be classified as errors.

rekka:
I see, thanks! I'll try to do something reasonable.

pkgw:
Hi @rekka,

Sorry for being so inactive on this. I'd like to get this PR actually merged! If you have a chance to make the final changes I suggested that would be great. Otherwise do you mind if I extend your PR with a couple of commits of my own that make the changes?

rekka:
Hi @pkgw,

Sorry about not making progress on this. OK, I will try to have a look in the next few days and finish the pull request.

rekka:
Hi, just to give an update, I haven't yet been able to spend much time on this because of a few deadlines that started coming up. If you prefer to get this merged soon, feel free to edit the PR. Sorry about that...

rekka:
Sorry about the delay! I think it's time to fix this possible UB. 

I added some crude handling of the internal bug when calling `input_seek` and `input_close` by aborting the engine using `longjmp`.

I also checked the other conditions that generate `tt_warning` in the bridge API and they all do not seem very serious so a warning is appropriate. They are basically related to IO and not caused by tectonic's bridge API, and the errors are reported by the return value back to xetex. Though the xetex engine is not very consistent at checking the error conditions.

I noticed that all the functions in `engines/mod.rs` that are called from C are missing the `extern` keyword. This seems like a bug since Rust's ABI is not stable and not necessary compatible with C <https://github.com/rust-lang/rfcs/issues/600>. Let me know if you agree and I'll add the keyword to the declarations.

rekka:
I have added the `extern` keywords. The use without `"C"` seems to be more idiomatic <https://github.com/alexcrichton/rust-ffi-examples/blob/a2ddf21e48b8c907d5a049114698283c2cca7847/c-to-rust/src/lib.rs>.

rekka:
#188 is a panic across FFI boundary as well, triggering undefined behavior. Surprisingly, the stack trace works even for the C functions...

We either should be careful not to panic in any Rust functions called by the C code, or wrap everything in [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html).

pkgw:
Sorry for taking so long to merge this!

I think I read somewhere that `longjmp` in C code called by Rust is not guaranteed to work, which will be quite a bummer if it ever starts causing problems ...

PR #92, opened 2017-06-11
More C cleanups

ronnychevalier:
They focus mainly on `dpx-dvipdfmx`.


pkgw:
Can't say that I understand all of the changes here but I trust that more-than-syntax ones make sense. Thanks!

PR #93, opened 2017-06-11
engines/mod: Make `format_to_extension` return Vec

### LINKED ISSUE #78: Support more image filename extensions
Right now I have hardcoded the I/O layer to guess only `.pdf` as an extension for image files, because the code that maps file formats to extensions can only return a single string: [here](https://github.com/tectonic-typesetting/tectonic/blob/master/src/engines/mod.rs#L719).

We also need to guess extensions like `.jpg` that can be used for other kinds of images. This can be done just by changing `format_to_extension` to return a vector of strings, and for its callers to loop and try all suggested extensions if needed.

mrshu:
* Make `format_to_extension` return a Vector of strings.

* Make use of `set_extension` function of PathBuf instead of creating an
  OsString.

* Add multiple vectorized values as results of `format_to_extension`.

* Fixes #78


src/engines/mod.rs
```diff
@@ -132,22 +132,23 @@ impl<'a, I: 'a + IoProvider> ExecutionState<'a, I> {
...
-        } else {
-            self.io.input_open_name(&ext.into_os_string(), self.status)
+        for e in format_to_extension(format) {
+            let mut ext = PathBuf::from(name);
+            if !ext.set_extension(e) {
+                return OpenResult::NotAvailable
+            }
+
+            if let FileFormat::Format = format {
+                if let OpenResult::Ok(r) = self.io.input_open_format(&ext.into_os_string(), self.status) {
```
mrshu: @pkgw There is most probably a better way of doing this than by repeating `OpenResult::Ok(r)` twice (same below). Unfortunately, I was not able to figure one out. Do let me know if you can, I will gladly change this.

Mrmaxmeier: I guess `@`-bindings are the preferred solution in this case:
```rust
if let result @ OpenResult::Ok(_) = compute_result() {
    return result
}
```

src/engines/mod.rs
```diff
@@ -703,29 +704,29 @@ enum FileFormat {
...
-        FileFormat::Vf => ".vf",
+        FileFormat::AFM => vec!["afm"],
+        FileFormat::Bib => vec!["bib"],
+        FileFormat::Bst => vec!["bst"],
+        FileFormat::Cmap => vec!["cmap"], /* XXX: kpathsea doesn't define any suffixes for this */
+        FileFormat::Enc => vec!["enc"],
+        FileFormat::Format => vec!["fmt.gz"],
+        FileFormat::FontMap => vec!["map"],
+        FileFormat::MiscFonts => vec!["miscfonts"], /* XXX: no kpathsea suffixes */
+        FileFormat::Ofm => vec!["ofm"], /* XXX: also .tfm */
```
mrshu: @pkgw I tried adding `tfm` here too, but it turns out that adding it here fails the `test_space` test -- the code actually fails at this point (https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/dpx-tfm.c#L579).

I therefore omitted it here.

src/engines/mod.rs
mrshu: @Mrmaxmeier Oh, may thanks!

src/engines/mod.rs
```diff
@@ -703,29 +704,29 @@ enum FileFormat {
...
+        FileFormat::OpenType => vec!["otf", "OTF"], /* XXX: also OTF */
+        FileFormat::Ovf => vec!["ovf", "vf"], /* XXX: also .vf */
+        FileFormat::Pict => vec!["pdf", "jpg", "eps", "epsi"], /* XXX: also .eps, .epsi, ... */
+        FileFormat::Pk => vec!["pk"],
+        FileFormat::ProgramData => vec!["programdata"], /* XXX no suffixes */
+        FileFormat::Sfd => vec!["sfd"],
+        FileFormat::Tex => vec!["tex", "sty", "cls", "fd", "aux", "bbl", "def", "clo", "ldf"], /* also .{sty,cls,fd,aux,bbl,def,clo,ldf} */
+        FileFormat::TexPsHeader => vec!["pro"],
+        FileFormat::TFM => vec!["tfm"],
+        FileFormat::TrueType => vec!["ttf", "ttc", "TTF", "TTC", "dfont"], /* XXX: also .ttc, .TTF, .TTC, .dfont */
```
Mrmaxmeier: A bunch of the trailing `XXX` comments can now be removed.

mrshu: @Mrmaxmeier roger that, will do. Thanks!

mrshu: @Mrmaxmeier I left at least a few ones that are still relevant.

mrshu:
@Mrmaxmeier @pkgw I am wondering if we can put together a test for this -- especially given the problem with `tfm`, I believe it would make these changes much more reliable and the whole PR safer to merge.

src/engines/mod.rs
pkgw: Ah ... I believe that by fixing this bug you're exposing a lingering problem in the font parsing code. I will look at this a bit more but I think the thing to do would be to file an issue about the problem and `#[ignore]` the failing test until the issue gets fixed.

src/engines/mod.rs
pkgw: Sorry, I'm not familiar with the `@` syntax and I can't see the key part of the diff. @Mrmaxmeier is your example syntax equivalent to this?

```rust
if let OpenResult::Ok(r) = compute_result() {
    return OpenResult::Ok(r)
}
```

Mrmaxmeier: Yup, @pkgw.<sup>*</sup>

###### *Recreating the enum variant inside of the `if let` statement could lead to lost/different type information but in this case its equivalent.

src/engines/mod.rs
mrshu: @pkgw Thanks, I also added a `FIXME` to the appropriate `#[ignore]` line.

tests/executable.rs
```diff
@@ -144,6 +144,7 @@ fn help_flag() {
 }
 
 #[test] // GitHub #31
+#[ignore] // FIXME: GitHub #95
```
mrshu: @pkgw it turns out I had to ignore a couple more tests in order for the tests to pass.

pkgw:
So, sorry — upon further investigation, I think that I was just wrong before. I now think the `Ofm` format should *not* have `tfm` as an allowed extension, which means you should be able to un-ignore the executable tests. Apologies for misleading you there.

rekka:
Commit 7c2e4e6 seems to break `tests/xenia/paper.tex`. Should I look into it or do you already know the cause?

Engine output:

```
(paper.tex
LaTeX2e <2016/03/31>
Babel <3.9r> and hyphenation patterns for 83 language(s) loaded.
(aastex6.cls

LaTeX Warning: You have requested document class `aastex6',
               but the document class provides `aastex'.

Document Class: aastex 2016/02/16 1.0/AAS markup document class
(times.sty) (lineno.sty) (revtex4-1.cls
Document Class: revtex4-1 2010/07/25/20:33:00 4.1r (http://publish.aps.org/revt
ex4/ for documentation)
ltxutil[2010/07/25/20:33:00 4.1r utilities package (portions licensed from W. E
. Baxter web at superscript.com)]
ltxfront[2010/07/25/20:33:00 4.1r frontmatter package (AO,DPC)]
ltxgrid[2010/07/25/20:33:00 4.1r page grid package (portions licensed from W. E
. Baxter web at superscript.com)]

Class revtex4-1 Warning: Failed to recognize \@vspace, \@vspacer, \@no@pgbk, \@
newline, and \\; no patches applied. Please get a more up-to-date class, .

(aps4-1.rtx) (aps10pt4-1.rtx) (textcase.sty) (url.sty) (natbib.sty
Local config file natbib.cfg used
(natbib.cfg

! LaTeX Error: Command \bibstyle@chicago already defined.
               Or name \end... illegal, see p.192 of the manual.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.206 ...tyle@chicago{\bibpunct{(}{)}{;}{a}{,}{,}}
                                                  
No pages of output.
Transcript written on paper.log.
```

rekka:
Investigating this, the problem is that the file `natbib.cfg` does not exist, but by trying `natbib` with various tex extensions, the file `natbib.sty` is opened instead. But this is clearly not what the code expects, since it is opening `natbib.cfg` from inside `natbib.sty`.

Unfortunately this breaks tectonic for me. :)

I believe that we should **not** change the extension of the file, if provided. Actually, how is the `FileFormat` used in the original XeTeX?

pkgw:
Gah, I did not test thoroughly enough. Thanks for #100.

PR #100, opened 2017-06-14
Do not change the extension of files requested by the engine.

rekka:
Fixes a regression caused by 7c2e4e622dde666a038154a9c925829e309a988f, for instance for `tests/xenia/paper.tex`.

Specifically, now tectonic tries to add default file format extensions only if the requested file does **not** have an extension already.

As I mentioned in #93, I believe that tectonic should not change the extension of files requested by a tex code, for instance return `natbib.sty` instead of `natbib.cfg`.

I also cleaned up the code to minimize allocations.


pkgw:
Thanks for fixing this so quickly! I should have tested more carefully.

PR #101, opened 2017-06-14
Remove an unused global NSAutoreleasePool

### LINKED ISSUE #47: Tectonic outputs NSAutoreleasePool log message
I get an `NSAutoreleasePool` log message when running Tectonic 0.1.5 on macOS 10.12.5. Doesn't seem to impede the PDF generation in any way, though.

```
$ tectonic resume.tex
note: this is a BETA release; report issues at https://github.com/tectonic-typesetting/tectonic/issues
Running TeX ...
warning: could not calculate MD5 of file "resume.out": it does not exist
note: warnings were issued by the TeX engine; use --print and/or --keep-logs for details.
Rerunning TeX because "resume.aux" changed ...
2017-05-31 21:27:30.151 tectonic[4170:87768] *** -[NSAutoreleasePool release]: This pool has already been released, do not drain it (double release).
Running xdvipdfmx ...
Writing resume.pdf (25785 bytes)
Skipped writing 2 intermediate files (use --keep-intermediates to keep them)
```

rekka:
The problem was a double free after all.

  - The global `pool` was initialized only once, but freed multiple times (on every
    engine run). Never used...

  - Closes #47.

### DIFF
--- a/tectonic/XeTeXFontMgr_Mac.mm
+++ b/tectonic/XeTeXFontMgr_Mac.mm
@@ -207,6 +207,8 @@ NSAutoreleasePool* pool = NULL;
 void
 XeTeXFontMgr_Mac::initialize()
 {
+    // An NSAutoreleasePool is needed to deallocate "autorelease" objects
+    // created and returned by Cocoa calls.
     pool = [[NSAutoreleasePool alloc] init];
 }
 
@@ -215,6 +217,7 @@ XeTeXFontMgr_Mac::terminate()
 {
     if (pool != NULL) {
         [pool release];
+        pool = NULL;
     }
 }
 


pkgw:
From doing a bit of digging, I think that the NSAutoreleasePool has an effect even if the variable itself is never explicitly referenced. See, e.g., [this chunk of code](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/XeTeXFontMgr_Mac.mm#L91-L102). So I don't think that removing it altogether is the right answer.

Does the warning go away if you just set `pool` back to NULL after freeing it?

rekka:
Oh, yes, you are completely right. Thanks for catching this. Looking at the [NSAutoreleasePool documentation](https://developer.apple.com/documentation/foundation/nsautoreleasepool), the pool is indeed needed to release "autoreleased" objects. Memory management is hard without Rust...

Yes, setting `pool` to `NULL` after a release fixed the warning in my test case (running the engine multiple times). I also added a comment for the future someone who is eager to remove unused globals.

But this got me a little worried, since before the fix, the code was releasing a different pool every time, which shouldn't have been a problem, unless the calls to `initialize` and `terminate` are not matched.

rekka:
Please do not merge yet, we still have a problem since indeed `terminate` is called multiple times, but `initialize` only on the first pass, based on [this code](https://github.com/rekka/tectonic/blob/69dd225779f449a72494d0b28f92f2c7344373f6/tectonic/XeTeXFontMgr.cpp#L66-L90).

rekka:
So looking at the code of `XeTeXFontMgr` a bit, it seems that this class pretty much assumes that it does not have to clean up since it appears to leak allocations. I am not sure why it even bothers to clean up partially with `terminate`.

It turns out that `NSAutoreleasePool` is not such a big deal compared to the [implementation that uses FontConfig](https://github.com/rekka/tectonic/blob/3c3e9b4745d4827e9b02f9d6d5a7585c327acbe8/tectonic/XeTeXFontMgr_FC.cpp#L335-L350). Here `terminate` deallocates a few things, and if they are then needed on the following tex pass, we get a segfault I believe (null pointer).

Unfortunately, according to [the comment](https://github.com/rekka/tectonic/blob/3c3e9b4745d4827e9b02f9d6d5a7585c327acbe8/tectonic/XeTeXFontMgr.cpp#L86-L88)

```
        // we don't actually deallocate the manager, just ask it to clean up
        // any auxiliary data such as the cocoa pool or freetype/fontconfig stuff
        // as we still need to access font names after this is called
```

 we cannot just `delete sFontManager` and set `sFontManager = NULL` in `Terminate` :(.

I might be overreacting now, but I guess to really settle this we need to properly reinitialize `XeTeXFontMgr` before every tex pass, i.e., all globals (statics), and then deallocate everything to not leak stuff.

pkgw:
Ah, yes, looks like this will need more investigation. Perhaps this is an area where updating to the latest version of XeTeX will clarify and/or improve things.

rekka:
I finally had time to have a look again. :) Turns out that the issue seems to have be resolved on the `master` branch. I did a `git bisect` run and it was fixed in 0e78c9564be430ec836c6168548540173d833b6a by properly deallocating `FontManager` after each xetex run. I am closing this pull request.

@pkgw: Thanks for fixing the leaks.

PR #103, opened 2017-06-15
Keep logs on error if --keep-logs provided

rekka:
This implements what is, I believe, the expected behavior of `--keep-logs`: Even if the xetex engine errors out, the logs are saved. This matches the tex standard and has at least two uses:

- Sometimes it is necessary to inspect the logs to understand why an error occurred.

- Editors, for instance `vim` with `vim-latex`, use the logs to locate errors and warnings.

A simple test is also provided.

I basically just extracted the code in `run` method that saves the outputs, etc., into `write_files` and added a parameter `only_logs`.

### DIFF
--- a/src/cli_driver.rs
+++ b/src/cli_driver.rs
@@ -587,11 +587,16 @@ impl ProcessingSession {
 
         // Do the meat of the work.
 
-        match self.pass {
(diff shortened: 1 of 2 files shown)


pkgw:
Looks great!

PR #104, opened 2017-06-16
Add an option to specify the output directory

### LINKED ISSUE #96: Specify output path
Unless tectonic writes the resulting file in the input file directory there should be a possibility to specify an output path (and default could be probably `.`)

h4llow3En:
This closes #96.

### DIFF
--- a/src/cli_driver.rs
+++ b/src/cli_driver.rs
@@ -463,7 +463,17 @@ impl ProcessingSession {
         }
 
         if let Some(dir) = args.value_of("outdir") {
-            output_path = PathBuf::from(dir)
+
+            output_path = {
+                let tmp = PathBuf::from(dir);
+                match tmp.extension() {
+                    Some(_) => {
+                        tt_warning!(status, "file extention in outdir; only the directory will be used");
+                        tmp.parent().unwrap().to_path_buf()
+                        }
+                    None => tmp,
+                }};
+            if !output_path.exists() { return Err(ErrorKind::Msg(format!("output directory \"{}\" does not exist", output_path.to_string_lossy())).into());}
         } else {
             output_path = fs_root.clone().to_path_buf()
         }
@@ -645,14 +655,9 @@ impl ProcessingSession {
                 continue;
             }
 
-            let real_path = match self.output_path.file_name() {
-                Some(_) => self.output_path.clone(),
-                None => {
-                    let mut tmp = self.output_path.clone();
-                    tmp.push(name);
-                    tmp
-                }
-            };
+            let mut real_path = self.output_path.clone();
+            real_path.push(name);
+
 
             status.note_highlighted("Writing ", &real_path.to_string_lossy(), &format!(" ({} bytes)", contents.len()));
 
--- a/src/cli_driver.rs
+++ b/src/cli_driver.rs
@@ -468,12 +468,12 @@ impl ProcessingSession {
                 let tmp = PathBuf::from(dir);
                 match tmp.extension() {
                     Some(_) => {
-                        tt_warning!(status, "file extention in outdir; only the directory will be used");
+                        tt_warning!(status, "file extension in `-o/--outdir`; only the directory will be used");
                         tmp.parent().unwrap().to_path_buf()
                         }
                     None => tmp,
                 }};
-            if !output_path.exists() { return Err(ErrorKind::Msg(format!("output directory \"{}\" does not exist", output_path.to_string_lossy())).into());}
+            if !output_path.exists() { return Err(ErrorKind::Msg(format!("output directory \"{}\" does not exist", output_path.display())).into());}
         } else {
             output_path = fs_root.clone().to_path_buf()
         }
@@ -714,8 +714,7 @@ impl ProcessingSession {
                 continue;
             }
 
-            let mut real_path = self.output_path.clone();
-            real_path.push(name);
+            let real_path = self.output_path.join(name);
 
 
             status.note_highlighted("Writing ", &real_path.to_string_lossy(), &format!(" ({} bytes)", contents.len()));
--- a/src/cli_driver.rs
+++ b/src/cli_driver.rs
@@ -645,8 +645,14 @@ impl ProcessingSession {
                 continue;
             }
 
-            let mut real_path = self.output_path.clone();
-            real_path.push(name);
+            let real_path = match self.output_path.file_name() {
+                Some(_) => self.output_path.clone(),
+                None => {
+                    let mut tmp = self.output_path.clone();
+                    tmp.push(name);
+                    tmp
+                }
+            };
 
             status.note_highlighted("Writing ", &real_path.to_string_lossy(), &format!(" ({} bytes)", contents.len()));
 
--- a/src/cli_driver.rs
+++ b/src/cli_driver.rs
@@ -456,18 +456,9 @@ impl ProcessingSession {
             }
         }
 
-        if let Some(dir) = args.value_of("outdir") {
-            output_path = {
-                let tmp = Path::new(dir);
-                match tmp.extension() {
-                    Some(_) => {
-                        tt_warning!(status, "file extension in `-o/--outdir`; only the directory will be used");
-                        tmp.parent().unwrap()
-                        }
-                    None => &tmp,
-                }
-            };
-            if !output_path.exists() {
+        if let Some(dir) = args.value_of_os("outdir") {
+            output_path = Path::new(dir);
+            if !output_path.is_dir() {
                 return Err(ErrorKind::Msg(format!("output directory \"{}\" does not exist", output_path.display())).into());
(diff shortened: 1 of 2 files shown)


src/cli_driver.rs
```diff
@@ -457,6 +462,22 @@ impl ProcessingSession {
             }
         }
 
+        if let Some(dir) = args.value_of("outdir") {
+
+            output_path = {
+                let tmp = PathBuf::from(dir);
+                match tmp.extension() {
+                    Some(_) => {
+                        tt_warning!(status, "file extention in outdir; only the directory will be used");
```
rekka: extention => extension
Maybe also `outdir` => `` `-o/--outdir` ``, so it's clear that outdir means the command line option?

src/cli_driver.rs
```diff
@@ -363,7 +363,7 @@ struct ProcessingSession {
     /// This is the virtual "CWD" that our filesystem accesses use. It is the
     /// dirname of `primary_input_path`, or an empty path (i.e., corresponding
     /// to the CWD if `primary_input_path` is None.
-    fs_root: PathBuf,
+    #[allow(unused)] fs_root: PathBuf,
```
rekka: Is `#[allow(unused)]` necessary?

src/cli_driver.rs
```diff
@@ -457,6 +462,22 @@ impl ProcessingSession {
...
+                match tmp.extension() {
+                    Some(_) => {
+                        tt_warning!(status, "file extention in outdir; only the directory will be used");
+                        tmp.parent().unwrap().to_path_buf()
+                        }
+                    None => tmp,
+                }};
+            if !output_path.exists() { return Err(ErrorKind::Msg(format!("output directory \"{}\" does not exist", output_path.to_string_lossy())).into());}
+        } else {
+            output_path = fs_root.clone().to_path_buf()
```
rekka: Isn't this a double clone? Just `clone()` seems enough.

src/cli_driver.rs
```diff
@@ -457,6 +462,22 @@ impl ProcessingSession {
...
+            output_path = {
+                let tmp = PathBuf::from(dir);
+                match tmp.extension() {
+                    Some(_) => {
+                        tt_warning!(status, "file extention in outdir; only the directory will be used");
+                        tmp.parent().unwrap().to_path_buf()
+                        }
+                    None => tmp,
+                }};
+            if !output_path.exists() { return Err(ErrorKind::Msg(format!("output directory \"{}\" does not exist", output_path.to_string_lossy())).into());}
```
rekka: `to_string_lossy` => `display`

src/cli_driver.rs
```diff
@@ -692,9 +714,10 @@ impl ProcessingSession {
                 continue;
             }
 
-            let mut real_path = self.fs_root.clone();
+            let mut real_path = self.output_path.clone();
             real_path.push(name);
```
rekka: This can be written as
```
let real_path = self.output_path.join(name);
```

tests/executable.rs
```diff
@@ -175,6 +175,26 @@ fn test_space() {
...
+    check_file(&tempdir, "subdirectory/1.pdf");
+}
+
+#[test]
+#[should_panic]
+fn test_bad_outdir() {
+    let tempdir = setup_and_copy_files(&["subdirectory/content/1.tex"]);
+
+    let output = run_tectonic(tempdir.path(),
+                              &["--format=plain.fmt.gz", "subdirectory/content/1.tex", "--outdir=subdirectory/non_existant"]);
```
rekka: non_existant => non_existent

src/cli_driver.rs
h4llow3En: I put this in because now it is unused. But maybe it will be needed later?
Else there would be a warning while compiling.

src/cli_driver.rs
h4llow3En: Just `clone()` gives a `&Path` which is mismatched with `PathBuf`

src/cli_driver.rs
rekka: Oh, I see, I misunderstood, I thought the `fs_root` below is this one. I think that it is better to remove it. It's easy to add it back if needed.

src/cli_driver.rs
```diff
@@ -457,6 +462,22 @@ impl ProcessingSession {
...
+                match tmp.extension() {
+                    Some(_) => {
+                        tt_warning!(status, "file extension in `-o/--outdir`; only the directory will be used");
+                        tmp.parent().unwrap().to_path_buf()
+                        }
+                    None => tmp,
+                }};
+            if !output_path.exists() { return Err(ErrorKind::Msg(format!("output directory \"{}\" does not exist", output_path.display())).into());}
+        } else {
+            output_path = fs_root.clone().to_path_buf()
```
rekka: I see, `fs_root` is `Path`. In that case, `fs_root.to_path_buf()` (or `fs_root.to_owned()`) should do.

src/cli_driver.rs
h4llow3En: Okay, I will do this

src/cli_driver.rs
```diff
@@ -449,14 +448,28 @@ impl ProcessingSession {
...
+                output_path = par;
                 io_builder.filesystem_root(par);
             } else {
                 return Err(ErrorKind::Msg(format!("can't figure out a parent directory for input path \"{}\"",
                                                   tex_path.to_string_lossy())).into());
             }
         }
 
+        if let Some(dir) = args.value_of("outdir") {
+
```
pkgw: Please remove this blank line.

src/cli_driver.rs
```diff
@@ -449,14 +448,28 @@ impl ProcessingSession {
...
+            output_path = {
+                let tmp = Path::new(dir);
+                match tmp.extension() {
+                    Some(_) => {
+                        tt_warning!(status, "file extension in `-o/--outdir`; only the directory will be used");
+                        tmp.parent().unwrap()
+                        }
+                    None => &tmp,
+                }};
+            if !output_path.exists() { return Err(ErrorKind::Msg(format!("output directory \"{}\" does not exist", output_path.display())).into());}
```
pkgw: Please don't condense the bodies of `if` statements to single lines. That is, please put the `return` statement and closing bracket on their own lines.

src/cli_driver.rs
```diff
@@ -1052,6 +1065,11 @@ fn main() {
              .help("How much chatter to print when running.")
              .possible_values(&["default", "minimal"])
              .default_value("default"))
+        .arg(Arg::with_name("outdir")
+             .long("outdir")
+             .short("o")
+             .value_name("OUTDIR")
+             .help("The output directory. Default: Same like INPUT."))
```
pkgw: Please format the default description as it appears in the other arguments; something like

`The directory in which to place output files. [default: the directory containing INPUT]`

h4llow3En:
As you wished, I cleaned up the Code

src/cli_driver.rs
```diff
@@ -449,14 +448,30 @@ impl ProcessingSession {
...
-                fs_root = par;
+                output_path = par;
                 io_builder.filesystem_root(par);
             } else {
                 return Err(ErrorKind::Msg(format!("can't figure out a parent directory for input path \"{}\"",
                                                   tex_path.to_string_lossy())).into());
             }
         }
 
+        if let Some(dir) = args.value_of("outdir") {
```
pkgw: This should use [value_of_os](https://docs.rs/clap/2.25.0/clap/struct.ArgMatches.html#method.value_of_os) to get an OsString from the argument.

src/cli_driver.rs
```diff
@@ -449,14 +448,30 @@ impl ProcessingSession {
...
+            output_path = {
+                let tmp = Path::new(dir);
+                match tmp.extension() {
```
pkgw: I don't think this code should special-case the circumstance where the argument path has an extension. It's not common but sometimes directory names have extensions like files, and in those cases you'll get surprising behavior because the program will alter the path that you pass it.

src/cli_driver.rs
```diff
@@ -449,14 +448,30 @@ impl ProcessingSession {
...
+                    Some(_) => {
+                        tt_warning!(status, "file extension in `-o/--outdir`; only the directory will be used");
+                        tmp.parent().unwrap()
+                        }
+                    None => &tmp,
+                }
+            };
+            if !output_path.exists() {
```
pkgw: On the other hand, I think here the check should use `is_dir`, not `exists`. I think this is closer to what we want and it will handle the case where the user inadvertently passes a filename instead of a directory name to the `-o` option.

h4llow3En:
I applied you change requests. Could you rerun the one failed test, please?

pkgw:
Transient test failure resolved. Thanks!

PR #98, opened 2017-06-13
Clean up C/C++ compiler warnings on OSX clang

pkgw:

PR #105, opened 2017-06-17
Prefix TeX (Y|YZ|Z|D)_(OK|NONE|SEEN|FIXED) constants

rekka:
- Prefix is `MOV_`.
  - Resolves the collision of `Z_OK` with a constant defined in
    `zlib.h` as mentioned in #98 .


pkgw:
Thanks!

PR #116, opened 2017-07-06
enable gcc-rs parallel feature

ronnychevalier:
Since we have lots of C code (splitted in many files), it makes sense
to build the code in parallel to reduce compilation time.


ronnychevalier:
I decreased by 10 seconds the compilation time in average on my computer when building in parallel.

pkgw:
Neat! Thanks!

Issue #117, opened 2017-07-08
Build against zlib 1.2.5


pkgw:
As discovered when working on [the Tectonic homebrew formula](https://github.com/Homebrew/homebrew-core/pull/14575) and relatedly in #102. It seems that we can't compile against zlib 1.2.5, which is a bit of a drag for the Homebrew packages and perhaps OSX in general.

I don't know anything about zlib's forward-compatibility guarantees but we should find a way to be more flexible about which versions of zlib we can build against.

pkgw:
This may have been fixed by #120.

pkgw:
Let's close this.

Issue #102, opened 2017-06-15
Build failure on OSX: undeclared identifier z_const


durka:
I followed the homebrew/cargo-install instructions. cargo-install stops at:

```
running: "cc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-I" "." "-I" "/var/folders/7l/cvkn0kbj307fyjc6zkfs6g_w0000gn/T/cargo-install.vXpTqybMO0Q8/release/build/tectonic-65e183164a4968bd/out" "-I" "/usr/local/Cellar/icu4c/58.2/include" "-I" "/usr/local/Cellar/harfbuzz/1.4.6_1/include/harfbuzz" "-I" "/usr/local/Cellar/glib/2.52.2/include/glib-2.0" "-I" "/usr/local/Cellar/glib/2.52.2/lib/glib-2.0/include" "-I" "/usr/local/opt/gettext/include" "-I" "/usr/local/Cellar/pcre/8.40/include" "-I" "/usr/local/opt/freetype/include/freetype2" "-I" "/usr/local/Cellar/graphite2/1.3.10/include" "-I" "/usr/local/Cellar/libpng/1.6.29/include/libpng16" "-DHAVE_GETENV=1" "-DHAVE_LIBPNG=1" "-DHAVE_MKSTEMP=1" "-DHAVE_STDINT_H=1" "-DHAVE_SYS_TYPES_H=1" "-DHAVE_SYS_WAIT_H=1" "-DHAVE_TM_GMTOFF=1" "-DHAVE_ZLIB=1" "-DHAVE_ZLIB_COMPRESS2=1" "-DXETEX_MAC=1" "-o" "/var/folders/7l/cvkn0kbj307fyjc6zkfs6g_w0000gn/T/cargo-install.vXpTqybMO0Q8/release/build/tectonic-65e183164a4968bd/out/tectonic/dpx-pdfobj.o" "-c" "tectonic/dpx-pdfobj.c"
cargo:warning=tectonic/dpx-pdfobj.c:387:68: warning: format specifies type 'unsigned short' but the argument has type 'unsigned char' [-Wformat]
cargo:warning=            _tt_abort("object type %hu not allowed in xref table", type);
cargo:warning=                                   ~~~                             ^~~~
cargo:warning=                                   %hhu
cargo:warning=tectonic/dpx-pdfobj.c:495:64: warning: format specifies type 'long' but the argument has type 'int' [-Wformat]
[...]

PR #120, opened 2017-07-10
add flag_if_supported and warning fixes

ronnychevalier:
This PR adds the `flag_if_supported` function allowing us to check whether a flag is supported by the compiler before adding it to the flag list.

It also enables some useful warning flags that did not generate any warning. Then, several commits fix and enable more warning flags.

I left some warning flags as a TODO (since they will generate warnings currently) if someone wants to work on those.


tectonic/dpx-cs_type2.c
```diff
@@ -731,7 +731,7 @@ do_charstring (card8 **dest, card8 *limit,
       do_operator2(dest, limit, data, endptr);
     } else if (b0 < 32 && b0 != 28) { /* 19, 20 need mask */
       do_operator1(dest, limit, data, endptr);
-    } else if ((b0 <= 22 && b0 >= 27) || b0 == 31) { /* reserved */
+    } else if ((b0 >= 22 && b0 <= 27) || b0 == 31) { /* reserved */
```
pkgw: I'm not sure where the "real" (x-)dvipdfm(x) upstream is, but I'd also suggest pointing out this fix to them.

tectonic/dpx-pdfobj.c
```diff
@@ -2028,7 +2028,7 @@ pdf_add_stream_flate (pdf_obj *dst, const void *data, int len)
 
     z.zalloc = Z_NULL; z.zfree = Z_NULL; z.opaque = Z_NULL;
 
-    z.next_in  = (z_const Bytef *) data; z.avail_in  = len;
+    z.next_in  = data; z.avail_in  = len;
```
pkgw: Ooh, I wonder if this will help with #102 and/or #117.

tectonic/xetexini.c
```diff
@@ -5,6 +5,414 @@
 #include <tectonic/core-bridge.h>
 #include <tectonic/dpx-pdfobj.h> /* pdf_files_{init,close} */
 
+memory_word *the_eqtb;
```
pkgw: Can you add a comment here saying that these variables are prototyped in xetexd.h, and an analogous one in xetexd.h saying that the variables are actuallye declared in xetexini.c?

build.rs
```diff
@@ -131,6 +134,58 @@ pub fn emit_stringpool(listfile: &Path, outstem: &Path) -> Result<()> {
     Ok(())
 }
 
+struct CompilerCheck {
```
ronnychevalier: Done. I will file an issue to gcc-rs later on :)

tectonic/xetexini.c
ronnychevalier: Done

tectonic/dpx-pdfobj.c
ronnychevalier: Yes I think it will fix it. I looked at `zconf.h` in zlib 1.2.5 and it is missing `zconst`, so since we do not use it anymore, it should work now.

tectonic/dpx-cs_type2.c
ronnychevalier: Sure, but I think this is not the first time we encounter a real bug in the dpx code caught thanks to a warning or similar (e.g., I remember fixing a use after free). I think I should try to come up with a list and point them to the commits in tectonic to see what they think. I will do that later when I have some time.

ronnychevalier:
@pkgw I filed an [issue](https://github.com/alexcrichton/gcc-rs/issues/173) to `gcc-rs` regarding `flag_if_supported`.

ronnychevalier:
`flag_if_supported` has been [merged](https://github.com/alexcrichton/gcc-rs/pull/215) in gcc-rs, we just have to wait for the next release to use it.

PR #125, opened 2017-08-17
update to the latest version of gcc to use flag_if_supported

ronnychevalier:
It allows us to remove our own flag_if_supported function.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ dependencies = [
  "error-chain 0.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "flate2 0.2.19 (registry+https://github.com/rust-lang/crates.io-index)",
  "fs2 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "gcc 0.3.51 (registry+https://github.com/rust-lang/crates.io-index)",
+ "gcc 0.3.54 (registry+https://github.com/rust-lang/crates.io-index)",
  "hyper 0.10.12 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


pkgw:
OMG, sorry for the delay here. Real life has intervened dramatically here. The CI is failing because of a problem that looks unrelated; someone will need to fix this first. I aim to do so soon but it's been tough to find the time these days ...

ronnychevalier:
On Mon, Aug 28, 2017 at 4:15 PM, Peter Williams <notifications@github.com>
wrote:

> OMG, sorry for the delay here. Real life has intervened dramatically here.
> The CI is failing because of a problem that looks unrelated; someone will
> need to fix this first. I aim to do so soon but it's been tough to find the
> time these days ...
>
Sure no problem :) I have noticed this error previously when trying to
compile using clang, I think it is just a matter of a missing header, but I
did not have time to fix it yet.

ronnychevalier:
After looking into it, it is not a missing header. We have to declare `-std=c++11` in order for `char16_t` and similar to be available in the C++ code for the unicode headers.

When looking at how compilers (MSVC, GCC, clang) support the `-std` flag, the intersection between them is `-std=c++14`. So I think we should simply add this to the C++ flags.

See also https://github.com/alexcrichton/cc-rs/issues/191

ronnychevalier:
I pushed #127 and the build succeeded. As soon as #127 is merged, I can rebase this one.

pkgw:
OK, should be good to rebase on, or merge in, master now.

ronnychevalier:
Done. The build succeeded :)

pkgw:
Yay! Thanks for sticking with this.

PR #127, opened 2017-10-08
build.rs: add -std=c++14 to have char16_t and similar defined

ronnychevalier:
unicode headers used in the C++ code use such types.

See #125


pkgw:
Thank you!!

Issue #122, opened 2017-07-20
tectonic fail out of memory


webbery:
Hi, I compiled a tex file in VirtualBox. Here is error info:
~~~
Running TeX ...
note: warnings were issued by the TeX engine; use --print and/or --keep-logs for details.
Running BibTeX ...
note: warnings were issued by BibTeX; use --print and/or --keep-logs for details.
Rerunning TeX because bibtex was run ...
Running xdvipdfmx ...
error: Out of memory - asked for 10663802 bytes
~~~
 But tex file was build success after I closed atom editor. Does tectonic consume a lot of memory?

pkgw:
Hmm, that is a very strange error. Tectonic uses preallocated memory arrays that are relatively large compared to a minimal TeX engine, but are extremely manageable for modern machines.

Are you able to produce a `.tex` file that consistently triggers this out of memory error, regardless of whether your editor is open or closed? I have trouble seeing how closing the editor would make the problem go away.

Issue #61, opened 2017-06-03
Sync with XeTeX


khaledhosny:
I’m not sure what version (or git revision) of XeTeX was used as the base here, but it seems to miss some of the new fixes in XeTeX (e.g. some serious bugs in OpenType math have been fixed in the last cycle which seems to be missing here).

pkgw:
Thanks for the heads-up. It's been a while since I've synced up with XeTeX sources.

And I also need to document how the syncing happens. It is possible to report exactly which version of XeTeX we track; we should do so.

alexreg:
Hmm, does it not make to use a Git submodule for the XeTeX bits? This might make the job of syncing sources easier.

Mrmaxmeier:
It looks like `441b4104a39b4df9ee64f2176d56ca7a451ef70c Fix non-portable use of MIN (replace with std::min) in XeTeXOTMath.` (2016-10-03)
is part of `tectonic/` while the child commit `02554b6fc8dc9a45d2740526f39761e4ba5d0f9f Migrate XeTeXOTMath to use new HarfBuzz API.` (2016-10-03) is not applied.

pkgw:
Probably the `MIN` change is a case of convergent evolution, but not actually intentionally applying that specific patch.

I have recently fixed up the [staging repo](https://github.com/tectonic-typesetting/tectonic-staging) to use [Norbert Preining's public git mirror](http://git.preining.info/texlive/) of the TeXLive repo, as well as updating the scripts and explaining the process in more detail. So it should now be somewhat more feasible for people who aren't me to track the evolution of the reference codebase.

(@khaledhosny: BTW, I am tracking the XeTeX sources embedded in the TeXLive repo, rather than the ones hosted at SourceForge. As far as I can tell, they are quite similar, so I just went with TeXLive since the rest of the "staging" process requires that checkout. If you think it would be better to track SourceForge, please let me know.)

I have also updated the reference sources in the staging repo. The final step in the process — that I am studiously avoiding — is manually importing the changes (relative to the last time the sources were synced) into the main Tectonic codebase. There's a lot to go through (and of course the amount to deal with only increases with time ...).

Relatedly, in this case I don't think we want to use a submodule, because many aspects of the Tectonic C code are expected to slowly diverge from the XeTeX source as the codebase evolves in its own direction. We want to follow the *semantics* of changes to XeTex as much as possible, but their expression in code is going to look more and more different.

[...]

PR #129, opened 2017-10-13
Miscellaneous C cleanups

Mrmaxmeier:
Hey,

I've got a bunch of commits awkwardly sitting around in [this](https://github.com/Mrmaxmeier/tectonic/commits/misc) branch.
About 1/3 of the commits are generated by coccinelle / other tools.
The commits can roughly be grouped like this:

##### Stringpool
These commits replace most uses of the `S(string_id)` macro with string literals.

Lots of the methods inside of `output.c` print static string using pointers into the stringpool. I've added versions of these functions based on string literals.
427d1e8 contains a coccinelle patch that replaces method invokations of the scheme `func(S(str_id))` with `func_cstr("string")`. I've applied this patch to a bunch of different places (4184d24,  fdd95e5,  eb75750, bb43e8c, ...) 

##### Includes

Perhaps the most opinionated change of the bunch is f2018bd.
I noticed that a lot of tools were thrown off by the include-structure of the project.
f2018bd replaces all includes of project-internal headers with the relative include syntax.
(`#include <absolute/include.h>` vs `#include "relative/include.h"`)

24881c4 applies Include-what-you-use to the code. The tool generates include statements from actual usage in the code. The main change is that some includes are reduced in scope and the code doesn't rely on transitivity of includes.

The include changes are pretty big and have little-to-no actual gain. If you think they complicate things like #61 I'll drop the commits.

##### Rerun corruption fixes
The XeTeX engine expects to be inside of a fresh process on every run.

2f0b933 resets the global state of xdvipdfmx before running. This prevents segfaults that previously happened everytime I tried to generate multiple PDFs from one process.

65bb3f2 frees most of the big statically allocated arrays after running the XeTeX engine. This might help with #122 because every run allocates at least ~145 MB.
This doesn't fix all memory leaks though, valgrind still reports about 4KB leaked per run (tested on a very basic document).

-----------------------------------------

I know that the changes are not at all perfectly structured and probably not easy to review.
If you'd like to pull in any of the commits I can try to purify the changes to multiple branches and open up new pull requests.


pkgw:
Thanks for the continuing work!

I don't have much time right now, but do you mind explaining how your changes to string pool handling work? There are so many places in the code that use integer identifiers for strings rather that `char *` types that I would be surprised that many of them could be change to avoid the `S(xyz)` macros.

But, I will try to find the time to read over your changes where I expect that I will figure out what you've done.

pkgw:
I haven't really taken a look at the changes in detail but I'm sure they're up to your usual standards so I'll just go ahead and merge. Thanks!

PR #134, opened 2017-11-16
Add Linux in Travis CI

### LINKED ISSUE #128: Add Linux in Travis CI
Currently, only OSX is tested with Travis CI. It would be great to also have Linux.

Mrmaxmeier:
It looks like the Linux CI builds are a lot faster (7 minutes with crates cached) so I've moved the builds on Rust nightly and beta to Linux.
In the future it might be interesting to explore different `gcc`/`clang` builds.

This resolves #128.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,36 +3,53 @@
 # Licensed under the MIT License.
 
 language: rust
-sudo: false
 
 cache: cargo
-
-os: osx
-osx_image: xcode8.2
-
-rust:
-  - stable
-  - beta
-  - nightly
+sudo: false
 
 matrix:
+  include:
+    - os: linux
+      dist: trusty
+      rust: stable
+    - os: linux
+      dist: trusty
+      rust: beta
+    - os: linux
+      dist: trusty
+      rust: nightly
+    - os: osx
+      osx_image: xcode8.2
+      rust: stable
   allow_failures:
     - rust: nightly
 
+addons:
+  apt:
+    packages:
+      - libgraphite2-dev
+      - libharfbuzz-dev
+      - libfontconfig1-dev
+      - libicu-dev
+      - zlib1g-dev
+      - openssl
+
 before_install:
-  - brew update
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,36 +3,53 @@
...
-  - brew update
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update                            ; fi
 
 install:
-  - brew install graphite2
-  - brew install harfbuzz --with-graphite2
-  - brew install --force openssl
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install graphite2                 ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install harfbuzz --with-graphite2 ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install --force openssl           ; fi
```
pkgw: I think these could be written, for example:

```
[[ "$TRAVIS_OS_NAME" != osx ]] || brew install graphite2
```

I'm not a big fan of extra intra-line spaces in lines to align syntax constructs (the `; fi` parts) so I prefer the `||` operator when possible. Although, perhaps it's possible to use multi-line YAML literals here to avoid the redundant tests?

build.rs
```diff
@@ -284,7 +285,7 @@ fn main() {
         .file("tectonic/XeTeX_ext.c")
         .file("tectonic/xetexini.c")
         .file("tectonic/XeTeX_pic.c")
-        .define("HAVE_TM_GMTOFF", "1")
+        // .define("HAVE_TM_GMTOFF", "1")
```
pkgw: When I was making the conda-forge packages for the recent release I ran into this issue too. It looks like `tm_gmtoff` is a BSD/GNU extension, so we shouldn't rely on it, so I think it is best just to remove this line altogether.

pkgw:
Great, thanks!

PR #136, opened 2017-11-17
tectonic/bibtex.c: increase size of various bibtex buffers

pkgw:
These now match the current values that are used in TeXLive installations, to the best of my ability. Yes, this patch was motivated because I exceeded the previous values in one of my documents.


pkgw:
The CI failure seems to be something weird not related to the patch, so: merging.

PR #139, opened 2017-12-09
Update pkgbuild

polyzen:


pkgw:
I have no idea how this all works so I'll trust that this PR fixes things! Thanks!

polyzen:
The build and resulting binary have been tested. :)

jtojnar:
The PKGCONFIG is deployed when you tag the release so you need to update the version and hash beforehand.

pkgw:
@jtojnar Thanks, I overlooked the fact that the version and hash are tracked in the PKGCONFIG file. Next time I'll try to do it right.

jtojnar:
@pkgw Actually, it might be kind of impossible since the tarball contains the PKGCONFIG file. We will need to update the travis script to publish a release tarball so we could know the hash.

pkgw:
Can the package depend on the tarball hosted on crates.io rather than the GitHub tagged file? If so, it looks like I can configure Cargo to exclude the `dist/` directory from that tarball, preventing the (sort of) circular dependency.

PR #141, opened 2018-01-14
Target GNU C dialect

Mrmaxmeier:
A bunch of functions like `strcasecmp` are only available in `glibc`.
This is the major source of warnings (about implicit function declarations).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -190,7 +190,7 @@ fn main() {
         "-Wno-unused-parameter",
         "-Wno-implicit-fallthrough",
         "-Wno-sign-compare",
-        "-std=c11"
+        "-std=gnu11"
     ];
 
     for flag in &cflags {


pkgw:
You are awesome! This was bothering me but I didn't quite know how to make the warnings go away.

In the long term would like everything to *not* be GNU specific, but in the long term I would also like to get rid of all of the C code anyway ...

Will merge if CI is happy.

Mrmaxmeier:
Thanks! Afaik most usages of GNU specific stuff are in (mostly unreachable) parts of `xdvipdfmx` that will probably be ripped out sometime.

[OSX builds](https://travis-ci.org/Mrmaxmeier/tectonic/builds/328746336) seem to be taking quite a while today. Not sure what that's about.

pkgw:
I've found that Travis CI's macOS builds can work up huge backlogs sometimes. I think there's just a lot of demand for them, and you need to run them on physical Mac hardware and there are just not enough machines to go around compared to Linux-like boxes.

PR #143, opened 2018-01-15
Test xdvipdfmx outputs

Mrmaxmeier:
Previously the `xenia/paper.tex` test failed on `macOS` so I've disabled it for the platform. It's probably something related to font handling.

BTW: The new submodule [breaks](https://travis-ci.org/Mrmaxmeier/tectonic/jobs/329134671) CI on my fork. Is its URL supposed to be relative?


pkgw:
Re: the submodule, the Git docs make it sound as if using a relative path for the clone URL will cause The Right Thing to happen, but I would not be shocked if it failed in practice. I think CI is also broken due to requirements for newer harfbuzz introduced by the catchup to TeXLive XeTeX, so there are a couple of things to work through ...

pkgw:
I see you have a branch taking a look at upgrading to `xenial`. It looks like even that isn't new enough. According to the header file, we need harfbuzz 1.3.3 or greater ... I'll see if I can figure out a solution.

tectonic/dpx-dvipdfmx.c
```diff
@@ -400,6 +403,12 @@ dvipdfmx_main (
     tt_aux_set_verbose(verbose);
   }
 
+  if (compress) {
+    pdf_set_compression(9);
+  } else {
+    pdf_set_compression(0);
+  }
+
```
pkgw: Maybe just `pdf_set_compression(compress ? 9 : 0)` ? One of the few cases that is concise enough that I don't find the ternary operator to add any confusion.

tectonic/dpx-vf.c
```diff
@@ -75,6 +75,14 @@ static struct vf *vf_fonts = NULL;
 static unsigned int num_vf_fonts = 0;
 static unsigned int max_vf_fonts = 0;
 
+
+void vf_reset_global_state(void)
```
pkgw: Very minor point, I like to a newline between the function name and the return type. That way you can pull up the function definition with `git grep ^vf_reset_global_state`.

tests/xdvipdfmx-outputs.rs
```diff
@@ -0,0 +1,149 @@
+// Copyright 2016-2017 the Tectonic Project
```
pkgw: It's 2018 now! (Sigh.)

tests/xdvipdfmx-outputs.rs
```diff
@@ -0,0 +1,149 @@
...
+        let mut f = File::create(&n).expect(&format!("failed to create {} for test failure diagnosis", n.to_string_lossy()));
+        f.write_all(observed).expect(&format!("failed to write {} for test failure diagnosis", n.to_string_lossy()));
+    }
+
+
+    println!("hex'd observed output:");
+    for &byte in observed {
+        print!("{:X}", byte);
+    }
+    println!("\n");
```
pkgw: Because the contents are saved to disk, this bit seems unnecessary do me. It will dump a whole lot to the screen if we start testing big files.

tests/xdvipdfmx-outputs.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    // Read in the expected "pdf" output ...
+    p.set_extension("pdf");
+    let pdfname = p.file_name().unwrap().to_owned();
+    let expected_pdf = read_file(&p);
+
+    let mut paper_support_files = PathBuf::from(TOP);
+    paper_support_files.push("tests");
+    paper_support_files.push("xenia");
+    let mut fs_paper_support = FilesystemIo::new(&paper_support_files, false, false, HashSet::new());
```
pkgw: This should disappear if we take the overall approach I suggest.

tests/xdvipdfmx-outputs.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    let mut genuine_stdout = GenuineStdoutIo::new();
+
+    // TODO: keep assets in git?
+    // for now, a change in the default bundle will break the tests
+    let config = PersistentConfig::open(false).unwrap();
+    let mut tb = config.default_io_provider(&mut NoopStatusBackend::new()).unwrap();
+
+    // this hits the network for ~10 MB on every execution .__.
+    // let mut tb = ITarBundle::<HttpITarIoFactory>::new("https://dl.bintray.com/pkgw/tectonic/tl2016extras/2016.0r4/tlextras-2016.0r4.tar");
```
pkgw: These ~8 lines too.

tests/xdvipdfmx-outputs.rs
```diff
@@ -0,0 +1,149 @@
...
+#[test]
+fn the_letter_a_pdf() {
+    do_one("the_letter_a")
+}
+
+#[test]
+#[cfg(not(target_os = "macos"))]
+fn paper_pdf() {
+    do_one("paper")
+}
```
pkgw: And this. And the added binary files.

pkgw:
(PS: I am happy/eager to merge this even with the CI failures, seeing as they are almost certainly unrelated.)

Mrmaxmeier:
Even the most basic tests using the `xdvipdfmx` backend require a bunch of (semi heavy) files like `pdftex.map`. These files could live inside of a [submodule](https://github.com/Mrmaxmeier/tectonic_test_bundle). I'd guess that this setup is flexible enough to build a bunch of new testcases without weighting on the main repository.

pkgw:
OK, I spent some time thinking about this work, and refashioned it somewhat into PR #147. That includes your work on PDF reproducibility, collection of common test code into the `util` module, and PDF tests of `the_letter_a` and `md5_of_hello` integrated into the `tex-outputs` test suite. The latter does the full processing in one unit, which I think is good to have. (Although I can also see how it can/will be useful to test XDV→PDF conversion without requiring a source `.tex` file; that would probably be the fastest path to getting high code coverage of `xdvipdfmx`.) I also changed another bit of platform-dependence in the `xdvipdfmx` output generation related to "unique tags" given to the fonts — although I don't actually understand what that code is for.

I did not include the `TestBundle` code because it didn't feel right for the assets to be packaged as in a Tectonic user cache directory. It really feels like the files should be packaged into a single directory under their TeX names, like the `tests/assets/` directory created in #147. Is there a reason that such an approach didn't seem feasible for the XDV→PDF tests?

Anyway, after #147, a lot of this PR has been integrated, and the non-integrated parts are ones that I think we need to ponder some more. So I'll close this PR, and integration of the `xenia` test and/or other, more demanding XDV→PDF conversions can go into new PRs.

PR #146, opened 2018-01-20
Remove strings.txt

Mrmaxmeier:
🚧 WIP: Even though tests pass I'm not fully confident this doens't break things. 🚧 
This replaces the remaining `S(...)` usages.
As a result the `strings.txt` file is no longer needed.


pkgw:
Wow! I didn't think removal of the string pool was going to be possible. I'll have to understand what's going on, and want to deal with your other PR first, but am excited to see if we can make this work!

Mrmaxmeier:
Note: This does not remove or change the string pool. It simply replaces a bunch of its usages with static C strings. The remaining 'static' references to the string pool are now replaced with insertions at run-time.
The core of the engine still uses `str_number` everywhere to reference strings. This concept loosely resembles [string interning](https://en.wikipedia.org/wiki/String_interning).

pkgw:
Ah, I see.

This all looks good to me. I've merged with `master` (some conflicts) and will merge once it's verified that the CI passes (whenever that is — Travis has been super slow lately).

pkgw:
I guess one could worry about the performance hit of inserting all of the primitive names at engine startup, though. Maybe it's best to keep them in the pre-baked string pool?

Mrmaxmeier:
The insert operation currently doesn't care about duplicates in the string pool. It simply pushes new strings to the end of the pool. Performance should be the same loading a pre-baked string.
I guess I'd be interesting to add some benchmarks to the test suite.

pkgw:
Yes, it doesn't make sense to worry too much without benchmarks — the subjective experience is still extremely fast.

I have [a repo](https://github.com/pkgw/rimphony/) where I set up Rust benchmarks — I remember it was a bit of a hassle to figure out how to do it.

Mrmaxmeier:
It looks like your merge pushed the new commits on top of my branch. I'll rebase on master so the commits won't appear twice.

pkgw:
I think the merge is OK, just GitHub's view of it is funky — if rebasing turns out to be a hassle, I don't believe it is necessary.

PR #151, opened 2018-01-23
Let `cargo-kcov` figure it out

Mrmaxmeier:
Locally `cargo-kcov` reports about 50% coverage with executable tests disabled. Let's see if this dramatically boost the coverage rate.


pkgw:
I skipped that one because I got weird failures, but I didn't put any effort into debugging them — maybe they won't actually be hard to solve.

pkgw:
Well, no errors this time ... Will merge once the full Travis suite passes, whenever that is.

pkgw:
Not sure why this didn't work when I tried it, but whatever. Great!

PR #161, opened 2018-02-23
Store format file uncompressed

Mrmaxmeier:
The gzip decoder is really slow in debug mode.
Disabling the compression for format files results in a 2x speedup of `load_fmt_file` in release builds.
The difference in debug builds is more noticeable and makes `cargo test`-ing more pleasant on _delicate_ hardware (180s => 30s).
This also results format blobs of doubled size though. Clever tweaks to the format could probably save a bit more space and still be faster than general-purpose compression.
Especially in debug mode, I think this performance/file-size trade-off is worth it.

I guess this shouldn't break anything as the format-file-format isn't stable at all.
If there's a special purpose to the compression or you think the file size is more important, feel free to close the PR.

**EDIT**: 7ac7766b2375cdd76e56f5b7eb9fdbdee7b99fe4 seems relevant here

### DIFF
--- a/src/cli_driver.rs
+++ b/src/cli_driver.rs
@@ -833,7 +833,7 @@ impl ProcessingSession {
             TexEngine::new()
                     .halt_on_error_mode(true)
                     .initex_mode(true)
-                    .process(&mut stack, &mut self.events, status, "UNUSED.fmt.gz", "texput")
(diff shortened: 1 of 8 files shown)


pkgw:
Yes, that's noticeably faster. And disk space is cheap.

PR #162, opened 2018-02-25
Work on symbolic accesses to structure fields

pkgw:
To understand the `line_break` routine so that I can futz with it, it helps a lot to turn opaque `memory_word` references into named structure field accesss. As far as I can tell, this has to largely be done manually. So, let's just get started.

Coccinelle helps a lot. Almost none of the field accesses are uniquely interpretable in the current code on their own, but contextual hints can be very useful for figuring out what's being done.


pkgw:
Rust stable CI failure due to [this bug](https://github.com/rust-lang/rust/issues/47309), triggered by [this PR](https://github.com/rust-lang/rust/pull/46916). Looks like it can safely be ignored for now.

Mrmaxmeier:
Nice! :+1:
Slowly but steadily approaching the original code quality.

PR #165, opened 2018-02-28
Cleanup handling of unicode filenames

Mrmaxmeier:
This simplifies the `name_of_file` variable by keeping it utf8-encoded instead of converting to utf16.
Previously filenames with low non-ascii characters like `ö` be wrongly processed while characters outside of the [BMP range](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) that require two utf16 code-points (like emojis) would cause segfaults.

With this patch tectonic now happily writes to `hallöchen 🐨 welt 🌍.pdf` instead of crashing.


pkgw:
I'm sorry, I have no idea why I never followed up this PR. I will see if I can rebase it onto the current master.

pkgw:
This is awesome! Thanks for a great contribution.

PR #169, opened 2018-04-09
Treat Beamer's .toc and .snm files as intermediates

### LINKED ISSUE #121: Intermediate `.snm` files left around when processing Beamer files
As reported by teras [on the forum](https://tectonic.newton.cx/t/minor-bug-with-snm-files/49):

> I have a minor issue. I am using beamer and when I use the \label command, there is some garbage left, namely a file with an .snm extension. Since all other intermediate files are gone (hurray!), I believe this file should be gone too.

It should not be hard to devise a simple test case here.

rekka:
Fixes #121


pkgw:
I would like to avoid hard-coding filename extensions if possible. Tectonic's detection of intermediate files tries to identify files that are written by the TeX engine, then later reread by it. Files that are kept around are ones that get written without ever being read. Do you know if that's what happens with these files?

rekka:
I agree that it's a bit fragile. Unfortunately, Beamer always produces `.toc` and `.snm` files even if it does not use them later: <https://github.com/josephwright/beamer/blob/1ba297ab03a3338bb243213044e9a60d9dc4a91b/base/beamerbasemisc.sty#L162-L183>

It only reads the `.toc` file if a table of contents is requested: <https://github.com/josephwright/beamer/blob/1ba297ab03a3338bb243213044e9a60d9dc4a91b/base/beamerbasetoc.sty#L89>

I have no idea what the `.snm` file is for: <https://github.com/josephwright/beamer/blob/1ba297ab03a3338bb243213044e9a60d9dc4a91b/base/beamerbasemodes.sty#L250>
I guess it depends on the theme?

I am not sure how to handle this better. Beamer seems to be quite popular so maybe adding an exception like this is warranted.

pkgw:
I think it's OK to add some special handling for not-that-unusual cases. I don't feel awesome about manually coding special extensions like this, but I think it's a nice part of the UX to not leave these files lying around and the user can always use `--keep-intermediates` if our logic gets things wrong.

One suggested change, then: can you code the list of special extensions as a constant list of strings, named something like `ALWAYS_INTERMEDIATE_EXTENSIONS`, and have the relevant logic check that list in some more structured way?

I forgot to say, thanks for looking into this!

rekka:
How about this?

pkgw:
Great! Thanks!

Issue #172, opened 2018-04-24
crate-ify the re-compilation logic


jneem:
I want to try creating a compilation server using tectonic (i.e. something that will watch some subset of input files and recompile when necessary). I found myself reimplementing a bunch of the logic that already exists in cli_driver.rs, and so I'm wondering if you'd accept a patch to separate the CLI interface from the logic that determines recompilation.
I've posted an extremely hacky version of what I mean here: https://github.com/jneem/tectonic_driver

pkgw:
Most definitely! The design of the Tectonic Rust code is quite unpolished so it might be a bit of a hassle to refactor it all in a way that works nicely, but any work in that direction would be very welcome!

PR #174, opened 2018-04-30
Make CliIoSetup public and rename it.

jneem:
This is a first step towards [issue 172](https://github.com/tectonic-typesetting/tectonic/issues/172): moving CliIoSetup into the main crate (and documenting it, and making it public). If this approach seems ok, I'll also do the same for ProcessingSession.


pkgw:
Apologies for the delay in looking at this.This is beautiful work! Thanks!

PR #181, opened 2018-06-18
Better format caching

pkgw:
This PR straightens out the story about caching format files so that the `--bundle` and `--web-bundle` options actually work again.

I've also snuck in a few unrelated changes: I'm bumping the version to 0.1.9-dev, and renaming `src/cli_driver.rs` to `src/bin/tectonic.rs`, which is what it should have been named for a while now.


pkgw:
CC @jneem, this changes some APIs that may affect your project.

PR #184, opened 2018-06-24
Fix various small issues in driver module.

jneem:
This makes some fields public, so that it's possible to get interesting
information out of IoEvents. Also, fixes the behaviour of
ProcessingSessionBuilder::output_dir to match its documentation (i.e.,
don't panic if the output_dir is not provided).


pkgw:
Looks great, will merge as soon as CI passes. Thanks!

Issue #200, opened 2018-08-08
URL to the default config is broken


lembacon:
URL to the default config `https://purl.org/net/pkgwpub/tectonic-default` hardcoded [here](https://github.com/tectonic-typesetting/tectonic/blob/fcaf88c5e287952c6601b0f2443eba7226ac4077/src/config.rs#L29) is broken. It is redirected to `https://dl.bintray.com/pkgw/tectonic/tl2016extras/2016.0r4/tlextras-2016.0r4.tar` and then leads to a `403 Forbidden` HTTP status code.

pkgw:
I just got an email from Bintray about rate-limiting my account, so I think this is an intentional choice by th e provider. I will investigate to see if the limitations can be lifted! If not we are going to need to switch to a backup. (Fortunately, the PURL.org redirect would make this transparent for existing installations.)

lembacon:
Thanks for the clarification. :-)

pkgw:
Working on it. In the meantime, I've uploaded the current files with another service. The bundle URL https://tectonic.newton.cx/bundles/tlextras-2016.0r4/bundle.tar should hopefully work. This can be activated either with the `-w` option to the command line program (if you're using version 0.1.8 — it was broken before) or by editing the default configuration file [as described here](https://tectonic.newton.cx/t/how-to-use-tectonic-if-you-can-t-access-purl-org/44).

pkgw:
[Bit more info here on the forums](https://tectonic.newton.cx/t/web-service-outage-workaround-available/140).

PR #203, opened 2018-08-10
Add --only-cached/-C command line flag

rekka:
This is a preliminary pull request proposing to add a **new command line option**. I want to ask for feedback.

With this flag enabled, tectonic will only use resource files that are already stored in the local cache and will not try to fetch them from a web bundle.

Tectonic at the moment accesses the bundle file whenever a new file like a figure is added to a TeX document or a new TeX file is being compiled. In a majority of these cases, this is unnecessary because the files will not be found in the bundle anyway. Usually, this is not a big issue, but it can be annoying to deal with when the web bundle is not accessible. I believe there should be a way to ask tectonic not to try to fetch the unknown files.

This has at least 3 use cases:
* workaround for web bundle problems like #200 
* no internet connection: preparing a presentation on a plane 
* speed up tectonic's edit-build-view cycle when adding a lot of new figures etc.

### DIFF
--- a/src/bin/tectonic.rs
+++ b/src/bin/tectonic.rs
@@ -101,13 +101,17 @@ fn inner(args: ArgMatches, config: PersistentConfig, status: &mut TermcolorStatu
         }
     }
 
+    let only_cached = args.is_present("only_cached");
+    if only_cached {
+        tt_note!(status, "using only cached resource files");
+    }
     if let Some(p) = args.value_of("bundle") {
         let zb = ctry!(ZipBundle::<File>::open(Path::new(&p)); "error opening bundle");
         sess_builder.bundle(Box::new(zb));
     } else if let Some(u) = args.value_of("web_bundle") {
-        sess_builder.bundle(Box::new(config.make_cached_url_provider(&u, status)?));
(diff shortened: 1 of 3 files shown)


pkgw:
Thanks for submitting! This month is very busy for me but I will do my best to give a proper review in timely fashion.

rekka:
Thanks for the suggestion. The thing is, the files that this commit touches do not currently have any  coverage, or more precisely, they do not show in the codecov report. Looking at how the command line arguments are tested at the moment, it would be natural to add a new test to `tests/executable.rs`. But those tests do not show up in the coverage report. They are in fact disabled when the code coverage is being run.

I guess one way to get coverage would be to rewrite the "executable" tests to call the `main` method of the binary directly, passing the command line arguments as an argument to `main` and then load them via [`clap::App::get_matches_from`](https://docs.rs/clap/2.32.0/clap/struct.App.html#method.get_matches_from).

This seems a bit too much for this pull request. Do you have a more straightforward solution?

Another question is how to actually test `--only-cached`. The result depends on the content of the cache.

pkgw:
Thanks for looking into it. All those points are things that should be addressed but I agree that yes, it's all a bit much for this one PR. And the code change is quite simple ... the kind of which of course never comes back to bite anybody ;-)

PR #211, opened 2018-08-14
Implement configuration for automatic local bundle loading.

crlf0710:
This commit adds an optional `local` parameter in `config.toml`, when specified as `true`, tectonic will try to load bundle from a zip file path specified in the usual `url` parameter.


pkgw:
Thanks for submitting this as well!

I like the general idea of this PR, but having a boolean `local` flag that causes the `url` field to be treated differently seems like a design that would be surprising and confusing for users. How about an approach where `file://` URLs are loaded as local Zip bundles?

crlf0710:
Ok, so i changed it to try to detect an `file://` prefix, if the prefix exists, it strips it off and try to load the rest as a path to a zip file. On my Windows machine this works fine.

crlf0710:
@pkgw Please review this when you have time. What can i do to improve code coverage in this case?

src/config.rs
```diff
@@ -85,16 +85,33 @@ impl PersistentConfig {
...
     pub fn default_bundle(&self, status: &mut StatusBackend) -> Result<Box<Bundle>> {
         if self.default_bundles.len() != 1 {
             return Err(ErrorKind::Msg("exactly one default_bundle item must be specified (for now)".to_owned()).into());
         }
 
-        Ok(Box::new(self.make_cached_url_provider(&self.default_bundles[0].url, status)?))
+        let url = &self.default_bundles[0].url;
+        const FILE_SCHEME_PREFIX: &'static str = "file://";
+        if url.starts_with(FILE_SCHEME_PREFIX) {
+			return Ok(self.make_local_file_provider(&url[FILE_SCHEME_PREFIX.len()..], status)?);
```
pkgw: I am a little concerned about URL encoding here. A proper URL might contain a sequence like `%20` to encode a space, but this approach won't convert it as appropriate. I think the solution here is to parse the URL for real — Tectonic already depends on the `url` crate, so that should be used. Then you can check if the `scheme` is `file`, and use the handy [to_file_path](https://docs.rs/url/1.7.1/url/struct.Url.html#method.to_file_path) function.

pkgw:
Thanks for addressing this! I had to merge in changes from another PR to master, but will merge this once the CI passes.

pkgw:
OK, as expected, code coverage remains a challenge, but let's get this merged.

PR #237, opened 2018-09-30
Tidying of the C/C++/ObjC code

pkgw:
Virtually no (intended) changes to the code, but here we organize the naming of the files and try to establish a better pattern of the various C/C++ headers. Old headers with names like "internals.h" and "tectonic.h" had a mish-mash of items specific to the XeTeX engine and items applicable to the overall C library. I think we do better now.

This PR renames a bunch of files, so things like coverage stats and any prior PRs to the C code are going to be more or less wiped out. Apologies if there's any work in progress that will be affected by these renames (but I'm not aware of any).


pkgw:
I don't know why codecov thinks the coverage went down nontrivially, but that's got to be spurious. Away we go!

PR #246, opened 2018-10-14
generate bridge header file

Mrmaxmeier:
This generates a [c header file](https://github.com/Mrmaxmeier/tectonic/blob/bindgen/tectonic/core-bindgen.h) using https://github.com/eqrion/cbindgen. cbindgen automagically includes all public C-representable types as well as primitive constants. This should reduce duplicate type declarations and removes some unsafe typecasts.

:construction: WIP :construction:
Unresolved stuff:
- ~This depends on https://github.com/eqrion/cbindgen/pull/216. https://github.com/eqrion/cbindgen/pull/217 would be nice too.~ PRs have been merged upstream.
- The cargo build.rs rebuild logic has problems. All changes to the bridge should trigger rebuilds of the C codebase.
- ~Including rustfmt-related stuff in this PR is awkward.~ The codebase has been rustfmt'd.
- The new warnings (`pdf_copy_clip`) should be dealt with. A treatment like https://github.com/tectonic-typesetting/tectonic/pull/238 might be appropriate.
- `cbindgen` introduces `serde` as a dev dependency. This breaks the `alpine-linux-musl` target.


pkgw:
Hi @Mrmaxmeier — can you explain a bit about the motivation for this PR? Is it basically adding the glue needed so that the Tectonic engine can be invoked from C as a library? (If so, that's great and would be a great feature to have!)

pkgw:
Done in #643!

PR #252, opened 2018-10-31
Add an all-in-one latex_to_pdf() function and sweet testing infrastructure

pkgw:
The front page of our API docs will now look way better! And the "executable" tests now have better behavior, not touching the local user's cache or the network.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -27,6 +27,22 @@ use io::Bundle;
 use status::StatusBackend;
 
 
+/// Awesome hack time!!!
+///
+/// This is part of the "test mode" described in the `test_util` module. When
+/// test mode is activated in this module, the `default_bundle()` and
+/// `format_cache_path()` functions return results pointing to the test asset
+/// tree, rather than whatever the user has actually configured.
+static mut CONFIG_TEST_MODE_ACTIVATED: bool = false;
+
+#[doc(hidden)]
+pub fn activate_config_test_mode(forced: bool) {
+    unsafe {
+        CONFIG_TEST_MODE_ACTIVATED = forced;
+    }
+}
+
+
 const DEFAULT_CONFIG: &'static str = r#"[[default_bundles]]
 url = "https://purl.org/net/pkgwpub/tectonic-default"
 "#;
@@ -89,7 +105,7 @@ impl PersistentConfig {
             only_cached,
             status,
         )?;
-        
+
         Ok(Box::new(bundle) as _)
     }
 
@@ -105,9 +121,14 @@ impl PersistentConfig {
         Ok(Box::new(zip_bundle) as _)
     }
 
+
     pub fn default_bundle(&self, only_cached: bool, status: &mut StatusBackend) -> Result<Box<Bundle>> {
         use std::io;
         use hyper::Url;
+
+        if unsafe { CONFIG_TEST_MODE_ACTIVATED } {
+            return Ok(Box::new(::test_util::TestBundle::default()));
+        }
 
         if self.default_bundles.len() != 1 {
             return Err(ErrorKind::Msg("exactly one default_bundle item must be specified (for now)".to_owned()).into());
@@ -127,8 +148,13 @@ impl PersistentConfig {
     		return Ok(Box::new(bundle) as _);
     }
 
+
     pub fn format_cache_path(&self) -> Result<PathBuf> {
-        Ok(app_dir(AppDataType::UserCache, &::APP_INFO, "formats")?)
+        if unsafe { CONFIG_TEST_MODE_ACTIVATED } {
+            return Ok(::test_util::test_path(&[]));
+        } else {
+            Ok(app_dir(AppDataType::UserCache, &::APP_INFO, "formats")?)
+        }
     }
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -12,9 +12,10 @@
 
 use std::io::{Read, Write};
 use std::io::ErrorKind as IoErrorKind;
+use std::ffi::OsStr;
 use std::fs::File;
 use std::path::PathBuf;
-use std::ffi::OsStr;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use app_dirs::{app_dir, app_root, get_app_root, sanitized, AppDataType};
 use toml;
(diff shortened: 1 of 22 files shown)


pkgw:
CC @Mrmaxmeier this code might be relevant to #246, if I'm understanding it correctly. The new all-in-one function might be something C users would be interested in. Also I'd be interested in any comments you have in general.

pkgw:
Woohoo, it *was* transient!

@Mrmaxmeier @rekka Would you either of you happen to have a chance to look this over? I'm trying to be better about having my changes reviewed by someone else ...

src/driver.rs
```diff
@@ -948,4 +1049,17 @@ impl ProcessingSession {
...
         Ok(0)
     }
+
+
+    /// Consume this session and return the current set of files in memory.
+    ///
+    /// This convenience function tries to help with the annoyances of getting
+    /// access to the in-memory file data after the engine has been run.
+    pub fn into_file_data(self) -> HashMap<OsString, Vec<u8>> {
+        // There must be a better way to do this ... Note that you *cannot*
```
Mrmaxmeier: This avoids the temporary. It can panic though:
```rust
Rc::try_unwrap(self.io.mem)
  .expect("multiple strong refs to MemoryIo files")
  .into_inner()
```

src/config.rs
```diff
@@ -27,6 +27,22 @@ use io::Bundle;
 use status::StatusBackend;
 
 
+/// Awesome hack time!!!
+///
+/// This is part of the "test mode" described in the `test_util` module. When
+/// test mode is activated in this module, the `default_bundle()` and
+/// `format_cache_path()` functions return results pointing to the test asset
+/// tree, rather than whatever the user has actually configured.
+static mut CONFIG_TEST_MODE_ACTIVATED: bool = false;
```
Mrmaxmeier: An `AtomicBool` here would avoid the few bits of unsafe code. Not sure if it's worth it though. `AtomicBool`s are safe to mutate by reference but slightly awkward to use.

rekka: I'd also suggest using `AtomicBool` to avoid any unnecessary `unsafe`. Using it is not that bad. 
Define using `static CONFIG_TEST_MODE_ACTIVATED: AtomicBool = AtomicBool::new(false);`.
Reading is as easy as `CONFIG_TEST_MODE_ACTIVATED.load(Ordering::SeqCst)` and storing `CONFIG_TEST_MODE_ACTIVATED.store(forced, Ordering::SeqCst)`.

pkgw: Yes, I agree that this is better. Thanks for the suggestion!

src/driver.rs
pkgw: I think I like this more ... avoids rebuilding the hashmap. And I am pretty sure that it will never panic unless the caller has made their own copy of `files`, in which case they're getting what they deserve.

src/io/setup.rs
```diff
@@ -89,14 +89,29 @@ impl IoSetup {
     }
 }
 
+
+/// Where does the "primary input" stream come from?
+enum PrimaryInputMode {
```
rekka: This seems to duplicate code from `driver.rs`. Wouldn't `PrimaryInputMode::Undefined` be better expressed as `None` for `Option<PrimaryInputMode>`?

src/lib.rs
```diff
@@ -67,3 +88,77 @@ pub use errors::{Error, ErrorKind, Result};
...
+/// Messages aimed at the user are suppressed, but (in the default
+/// configuration) network I/O may occur to pull down needed resource files.
+/// No outputs are written to disk; all supporting files besides the PDF
+/// document are discarded. The XeTeX engine is run multiple times if needed
+/// to get the output file to converge.
+///
+/// For more sophisticated uses, use the [`driver`] module, which provides a
+/// high-level interface for driving the typesetting engines with much more
+/// control over their behavior.
+pub fn latex_to_pdf<T: AsRef<str>>(latex: T) -> Result<Vec<u8>> {
```
rekka: Is tectonic thread safe now? If I remember correctly, there used to be unsynchronized static globals in the C core, and the access to the bundle is not synchronized. It seems that this function lets the user run multiple instances of the engine in parallel.

src/io/setup.rs
pkgw: Yes, it duplicates a bit. But in both cases, the types are private to their respective modules — they are more or less implementation details of the "builder"-type objects. Because they barely come with any code, I felt it was better to duplicate a bit than to try share the type within the crate. Reasonable enough?

rekka: That makes sense. I did not see the big picture.

src/lib.rs
pkgw: Good point. We are still not thread safe, and probably won't be for a long time: the TeX engine just has a *lot* of shared global state in the C code. It will take a *long* time to put that all into some kind of non-global struct.

At the moment I can think of two solutions:

1. Bite the bullet and add locking to the engines.
2. Skirt the issue and just mention in this function's documentation that it is not thread-safe in the way you describe.

I lean towards option 2. Thoughts?

rekka: I would strongly prefer option 1. Rust selling point is "threads without data races", enforced by the type system. A quick and dirty solution is to have a Mutex that has to be acquired at the beginning of `latex_to_pdf`(similar to the test infrastructure currently), and add a note in the documentation that at most one instance of the engine can be running at a time and so the function might block until the other instances are finished.

rekka: That's definitely a concern. One way to communicate this would be introducing a singleton struct, say `Tectonic`, that would have methods `latex_to_pdf` and `processing_session_builder`. The user would have to first create the struct using `Tectonic::new()`, which would return an error if `Tectonic` were already in use.

pkgw: I think that approach is a little less desirable because one day, hopefully, we will be thread-safe, and then the singleton `Tectonic` will be unnecessary and a bit inconvenient.

I'll investigate adding locking within the crate and see how it works out.

pkgw: Yeah, that was pretty easy to implement. I tend to think this is the right way to go.

pkgw:
OK, this PR now adds an internal mutex in the Tectonic crate as well as the other changes building up to the all-in-one function.

The Circle CI build failed again due to the same error as before, which was transient the first time. I've relaunched it. If this failure happen frequently we'll need to figure out some way to mitigate it.

Any final comments?

pkgw:
OK, will need to investigate the QEMU/big-endian doctests, but let's go ahead and merge this.

PR #256, opened 2018-11-06
Release 0.1.11

pkgw:
From hell's heart I stab at thee, purl.org SSL certificate! :whale:


pkgw:
OK, the CircleCI build had some kind of external spurious failure, and it takes forever to run, and it seems to have internal spurious failures anyway these days. Merging.

PR #259, opened 2018-11-15
Remove serde-derive dependency

malbarbo:
Musl target does not work with proc-macro.
See https://github.com/rust-lang/rust/issues/40174

Removing serde-derive allows building tectonic for musl targets.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -492,22 +492,6 @@ version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
-name = "proc-macro2"
-version = "0.4.20"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "unicode-xid 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
-name = "quote"
-version = "0.6.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "proc-macro2 0.4.20 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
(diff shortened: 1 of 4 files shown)


pkgw:
Thanks for raising this up to a PR!

It would be really cool to be able to build on musl, but I also expect that there will be an increasing number of parts of Tectonic where using serde (or other proc-macros) would be very valuable.

So, how about adding a [Cargo feature](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section) to allow this to be controlled at build time? My vision would be that the feature would *enable* serde-derive and that it would be a "default" feature, so that, well, by default it would be activated. But it could be deactivated for cases like this. [Here is a simple example](https://github.com/BurntSushi/ripgrep/blob/master/grep/src/lib.rs#L19-L20) of how you can conditionally compile code based on Cargo feature activation.

When the feature is deactivated (i.e. the musl case), I see two possible ways to do things. Either emulate the derived Serde impl, as you've done here, or use some simpler replacement code that basically always yields a default PersistentConfig struct. I'm leaning towards thinking that the second approach might be better, because if we start adding more serializable structs to Tectonic, the manual impls are going to start getting really hairy.

If we use a feature gate, we should add some new documentation in the README to explain why the feature exists. I think the best approach would be to have a section labeled "Features" that would explain that this feature basically just exists to enable the creation of static binary. (If we end up using other proc-macro features in the future, I think it would be better for them to all have their own features, rather than one big grab-bag `stuff-that-breaks-static-linking` feature.)

How does that all sound?

malbarbo:
I made a new [PR](https://github.com/tectonic-typesetting/tectonic/pull/261).

pkgw:
Superseded by #261.

PR #261, opened 2018-11-17
Make serde dependency optional

malbarbo:
This introduces the `confile_file` feature that is enabled by default. Disabling the feature allow building for musl targets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,8 +59,8 @@ toml = "^0.4"
 zip = "^0.4"
 
 [features]
-default = ["config_file"]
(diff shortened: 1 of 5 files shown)


malbarbo:
The information about the config file in the README need to be expanded... Can some one else work on this?

pkgw:
@malbarbo This is great! In a minute am going to update this PR with a commit with a few naming tweaks, but this looks pretty much perfect to me. Please let me know if you think that my modifications look OK?

malbarbo:
@pkgw It's look great to me!

PR #270, opened 2018-12-07
Fix changelog grammar: "the code [...] mode [...] reusable" → "the code [...] has been made [...] reusable"

das-g:
I first noticed this on the [releases page](https://github.com/tectonic-typesetting/tectonic/releases). Probably should be fixed there, too, if possible.


pkgw:
Thanks for the improvement! I'll update the release text.

PR #285, opened 2018-12-30
dist/appimage: tidy this up

pkgw:
The main thing is that by using `target/release` as the AppDir, the potentially multiple gigabytes of build artifacts (!) get included in the AppImage, making it far far larger than necessary. By using a dedicated AppDir I get an AppImage that's 22 MiB.

CC @xtaniguchimasaya.


pkgw:
Haven't gotten confirmation from the AppImage people that this is fully working, but I think we're at least close, so I'm going to merge. If any fixup is needed, can do it in another PR.

PR #317, opened 2019-02-27
Some more cleanup

Mrmaxmeier:


tectonic/xetex-xetex0.c
```diff
@@ -678,17 +678,18 @@ print_subsidiary_data(int32_t p, UTF16_code c)
 
 void print_style(int32_t c)
 {
-    switch (c / 2) {
-    case 0:
+    // DISPLAY_STYLE matches style 0 and 1, etc
+    switch (c ^ (c & 1)) {
```
pkgw: Err, this seems semantically different than the previous code, no? `c / 2` is `c >> 1`, while if I'm thinking this through correctly, `c ^ (c & 1)` is equivalent to `c & 0xFFFE`. Looking at The TeXbook, the existing code seems right. Am I missing something here?

Mrmaxmeier: It basically switches over `(c / 2) * 2` because `TEXT_STYLE == 2, SCRIPT_STYLE == 4, ...`.
I didn't realize that case statements support constant integer expressions.  This feels more obvious:
```c
switch (c / 2) {
case DISPLAY_STYLE / 2:
  print_esc_cstr("displaystyle");
  break;
case TEXT_STYLE / 2:
  print_esc_cstr("textstyle");
  break;
...
```

pkgw: Ohhhhh, I get it a bit better — I didn't notice that the named constants had different values than the hardcoded integers in the existing code. Now I understand what was behind your first approach.

I think I might prefer your first approach to having everything `/ 2`, but this is all pretty low-importance — at first I was wondering if the semantics of the code were actually changing, but now I get that they're not.

PR #328, opened 2019-03-14
Suppress unknown warnings on Clang

spl:
For this line:

```c
#pragma GCC diagnostic ignored "-Wstringop-overflow"
```

Clang prints this warning:

```
warning: unknown warning group '-Wstringop-overflow', ignored [-Wunknown-warning-option]
```

This change imports a library with clever CPP that suppresses warnings for GCC and avoids showing warnings for Clang.

I imported the whole library to include the license and helpful documentation.

I renamed the header files to hint that push and pop need to be included in equal (nested) numbers.

### DIFF
--- a/tectonic/dpx-truetype.c
+++ b/tectonic/dpx-truetype.c
@@ -122,11 +122,13 @@ pdf_font_open_truetype (pdf_font *font)
         length = tt_get_ps_fontname(sfont, fontname, 255);
         if (length < 1) {
             length = MIN(strlen(ident), 255);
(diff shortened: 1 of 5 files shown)


pkgw:
I filed #350 which takes an alternate approach here. The limitation with this library is that it doesn't actually suppress the warnings on GCC — which happens to be my everyday compiler. For the time being I think it's simpler just to explicitly gate the `#pragmas` based on the active compiler. What do you think?

spl:
Happy to close this in favor of #350.

Issue #59, opened 2017-06-02
Support HTTP proxies


l2dy:
```
error: cannot write format file tectonic-format-latex.fmt.gz
caused by: this I/O layer cannot save format files
```

###### Environment
Tectonic 0.1.5
macOS 10.12.5
rustc 1.17.0
cargo 0.18.0

pkgw:
Can you paste the exact command line that you're using to run Tectonic?

I believe that you need to add the argument `--outfmt=format` ... if that works, then clearly there is a bug in the user experience here that needs to be addressed.

l2dy:
`tectonic -w 'https://dl.bintray.com/pkgw/tectonic/tl2016extras/2016.0r4/tlextras-2016.0r4.tar' zh.tex` with https://github.com/l2dy/macports-notes/blob/master/zh.tex.

I tried `--outfmt pdf` but that didn't work.

pkgw:
Ohhhh, sorry, I misinterpreted. You have run into a *different* area where the command-line program is delivering a confusing user experience. The `-w` option is currently a bit busted compared to what happens if you run the program without any arguments.

If you just run `tectonic zh.tex` I think it should work. I am having trouble testing since I don't have the `Noto Serif CJK SC` fonts installed on my system.

l2dy:
In China the default URL is blocked. Can I set HTTP proxy in tectonic?

pkgw:
Oh! I was wondering why you bothered to use the `-w` option.

Tectonic uses the Rust crate `hyper` for web accesses. It looks like proxy support could be enabled with some explicit code in Tectonic: [see this thread](https://github.com/hyperium/hyper/issues/531).

I will change this issue to be about adding proper proxy support.

In the meantime, you should be able to get things working by creating a configuration file that specifies the specific URL. On Linux, this file is looked for at  `~/.config/Tectonic/config.toml`. On OSX, I believe it will be `~/Library/Application Support/Tectonic/config.toml`. The contents should then be:

```
[[default_bundles]]
url = "https://dl.bintray.com/pkgw/tectonic/tl2016extras/2016.0r4/tlextras-2016.0r4.tar"
```

In more encouraging news, I downloaded the fonts and I think that the document will compile correctly if you can access the bundle in this way.

l2dy:
Thanks, it worked. Please document the config file somewhere.

webbery:
[...]

PR #330, opened 2019-03-16
Replace hyper with reqwest

spl:
This is an attempt to replace _hyper_ with _reqwest_.

I'm not familiar with either package, and I'm still pretty new to Rust, so I'm posting a draft PR to get feedback before finalizing it. Also, I left in my intermediate commits, but in the end, it would probably be better to squash them before merging.

My goal is to bring in a newer HTTP dependency than hyper 0.10. First, reqwest seems pretty decent for what's being done in Tectonic. Second, it will allow for a simple solution to introducing a proxy (#59).

Some highlights:
* I removed some intermediate variables (e.g. `client`, `req`) where it didn't seem like they added much to the code.
* I added an error case `UnexpectedHttpResponse` for several errors. I'm not sure if I did everything correctly there.
* The range stuff is a bit different now, thanks to the apparently new headers package. Do the tests cover this? How could it be tested?
* The redirect policy is also different, thanks to reqwest's approach. Can you verify that the logic is the same? Perhaps a test would be good for this?
* I chose an arbitrary `MAX_HTTP_REDIRECTS_ALLOWED`. You may want to do something different here.


pkgw:
@spl Thanks for undertaking this! After a quick skim, it looks like you're very much on the right track here. No major comments about what's in the diffs thus far.

As for testing … Tectonic's testing for this kind of thing is not as thorough as I'd like. To test this particular subsystem well, I'm pretty sure you'd need to stand up your own local HTTP server and hardcode various scenarios. That wouldn't be impossible, but not easy either. Might be worth taking a look at other comparable crates and see if they have any clever approaches to testing that we can borrow.

Mrmaxmeier:
Just an observation: This update adds a bunch of new dependencies from the async/tokio ecosystem.
This has a noticeable effect on the size of the `tectonic` binary.

| build | debug | release |
| --------- | ------ | ------ |
| master (hyper 0.10) | 42M | 6.2M |
| reqwest | 69M | 8.6M |
| without http | 29M | 4.9M |

It might be nice to feature-gate the networking code with an on-by-default flag.

pkgw:
@Mrmaxmeier That's a good point to raise. I'm surprised that tokio/async adds that much to the binary size ... isn't that stuff all supposed to compile down nice and efficiently? But `reqwest` certainly pulls in a lot of dependencies. Does anyone know if we could simply upgrade `hyper` and not pull in all of the extra stuff `reqwest` needs?

spl:
Based on a rough diff of [`hyper`](https://crates.io/crates/hyper) and [`reqwest`](https://crates.io/crates/reqwest), I think these are the dependencies that `reqwest` has in addition to `hyper`'s (some are optional for `hyper` but not for `reqwest`):

* base64
* encoding_rs
* hyper
* libflate
* mime
* mime_guess
* serde
* serde_json
* serde_urlencoded
* tokio
* tokio-executor
* tokio-threadpool
* tokio-timer
* url
* uuid
* hyper-old-types optional
* hyper-rustls optional
* hyper-tls optional
* native-tls optional
* rustls optional
* tokio-rustls optional
* trust-dns-resolver optional
* webpki-roots optional

> Does anyone know if we could simply upgrade `hyper` and not pull in all of the extra stuff `reqwest` needs?

I'm sure we could. It probably won't be as convenient, esp. for things like redirect handling and proxy handling.

Are there any other [HTTP client crates](https://crates.io/search?q=http%20client) that should also be considered?

spl:
This [StackOverflow post](https://stackoverflow.com/questions/54221596/a-synchronous-http-client-for-rust) presents a similar concern about binary size and one answer mentions some smaller-sized HTTP clients. I've looked through them, but I'm not seeing one that is going to be better and better maintained than `hyper` or `reqwest`.

spl:
On redirect handling: It seems like there is no plan to implement this in `hyper` (https://github.com/hyperium/hyper/issues/1719), so `tectonic` would have to do it manually. There is a crate called [`follow-redirects`](https://crates.io/crates/follow-redirects) for `hyper`, but it does not expose a way to customize the policy.

On proxy: There is a [`hyper-proxy`](https://crates.io/crates/hyper-proxy) crate that might work.

spl:
I've done a bit more research on `hyper`. As I said, I'm new to this, so these notes are for me as much as they are for anyone else:

`hyper` v0.12 has a `runtime` Cargo feature. It is enabled by default. When enabled, it uses (as of v0.12.25) `futures-cpupool`, `net2`, `tokio`, `tokio-executor`, `tokio-reactor`, `tokio-tcp`, `tokio-threadpool`, and `tokio-timer`, as a “run time.” (I'm not quite sure how “run time” is defined here, but it seems to be pretty central to the operation of `hyper`.)

It is possible to disable the `runtime` feature and avoid depending on all of these. That aspect is not documented (https://github.com/hyperium/hyper/issues/1613). However, there is a [comment](https://www.reddit.com/r/rust/comments/9qma0n/hyper_cpu_usage_problem_alternatives/e8apbkg/) on using a single-threaded runtime with `tokio::runtime::current_thread::Runtime`. And, as I just discovered, it appears that `reqwest` does this. I don't know if disabling the `runtime` feature and using `tokio::runtime::current_thread::Runtime` gains anything in terms of reduced dependencies: I haven't tried to figure out how the `tokio` dependencies work.

spl:
I looked at [`curl`](https://crates.io/crates/curl), a `libcurl` wrapper, as an alternative. Using it directly adds too much work (IMHO), because the crate does not handle HTTP response headers. There is a crate [`chttp`](https://crates.io/crates/chttp), which wraps `curl` and provides a higher-level HTTP client interface at the level of `reqwest`. It supports a proxy but is missing a customizable redirect handler.

I'm leaning toward biting the bullet on the larger binary size and using `reqwest` for the simple redirect and proxy handling. That seems worth more to me than trying to use `hyper` without `tokio` (or trying to use another crate) and writing more code specific to `tectonic`.

As for a Cargo feature to control the HTTP dependencies, I don't know what that would look like. I suppose, if disabled, would `tectonic` be restricted to a file cache?

Thoughts? What should be the next step?

pkgw:
I also lean towards biting the bullet on this one. I'd like to keep the binary small, but in my view it's more important to deliver good functionality, and adding the ability to honor HTTP proxy settings would be valuable. It looks like there's a gap in the ecosystem for a get-me-the-damn-file HTTP client crate that supports things like redirects and proxies (thanks for researching this, @spl) — if one comes along, we can definitely revisit this in the future.

If we added a feature to disable HTTP, people could use the local cache and/or locally stored bundle files, although there's very little in the way of infrastructure to support the latter. (In principle you can build your own following the instructions in the `tectonic-staging` repo, but I'm not aware of anyone having done so.) One thing that I'd like to do is add more CLI tooling support for some of this stuff — I have an idea that maybe the right way to do things here is have something like the `rustc`/`cargo` split, where the core compiler is one executable, and then there's a separate higher-level tool that provides the main user interface to the functionality.

pkgw:
@spl BTW as far as you're concerned, is this still a "draft" PR, or is this ready for testing and review?

spl:
> BTW as far as you're concerned, is this still a "draft" PR, or is this ready for testing and review?

I think the basic change is done, assuming I got my logic right. My current thoughts on what needs to be done:

* [ ] There are one or two remaining comments in the code mentioning “hyper” that should be updated to say something else. Can you take a look at those, @pkgw?
* [ ] There's `src/io/hyper_seekable.rs`, which isn't used. Should that be removed from the repository? I suppose you could put it in a Gist if you want to keep it around.
* [ ] I think the dependencies should be updated to `master` (so, probably, `master` merged in) and tested with this branch.

I can look at it again tomorrow. I checked the “allow edits from maintainers” checkbox if you want to push directly to this branch. (But preferably no force-pushes since it's now shared.)

spl:
There is an (allowed) [error](https://travis-ci.org/tectonic-typesetting/tectonic/jobs/511474187) building with `docker` for the `x86_64-alpine-linux-musl` target:

> error: cannot produce proc-macro for `headers-derive v0.1.0` as the target `x86_64-alpine-linux-musl` does not support these crate types

This seems to be due to the use of `proc-macro2` in `headers-derive`:

```toml
[dependencies]
proc-macro2 = "0.4"
```

Researching the issue, I came upon #259, which is related to the same problem but with `serde-derive`.

I'm not aware of any way of using `headers` without its dependency `headers-derive`. The use of `headers` _is_ rather small. It's of course nice to have the type safety, but `headers` could be dropped in favor of a string header.

I'm not sure, but there might have been some recent work toward a resolution of this problem in https://github.com/rust-lang/rust/pull/58575 (tracking issue: https://github.com/rust-lang/rust/issues/59302).

@malbarbo: I see that you commented there. Could you help us understand this?

sanmai-NL:
@spl: See rust-lang/cargo#5266 for the upstream issue. Setting the environment variable `RUSTFLAGS="-C target-feature=-crt-static"` (spot the minus symbol) works around it.

pkgw:
@sanmai-NL Thanks for the pointer. If we set the `RUSTFLAGS` in the way you suggest, does that mean that the resulting Tectonic binary will not be fully static? I'd like to be able to keep on producing static binaries if possible ... especially since, as I understand it, the limitation is on the compiler `proc-macro` infrastructure, and not actually on the code directly inside Tectonic.

spl:
@sanmai-NL Thanks. I did read that issue but didn't understand it at the time. Now, having read some of [RFC 1721](https://github.com/rust-lang/rfcs/blob/master/text/1721-crt-static.md), I think I get the gist of it.

The important part is this, I believe:

> Specifying C runtime linkage via `-C target-feature=+crt-static` or `-C target-feature=-crt-static`. This extends `-C target-feature` to mean not just "CPU feature" ala LLVM, but "feature of the Rust target". Several existing properties of this flag, the ability to add, with `+`, or remove, with `-`, the feature, as well as the automatic lowering to `cfg` values, are crucial to later aspects of the design. This target feature will be added to targets via a small extension to the compiler's target specification.

To clarify what I understand, the `x86_64-alpine-linux-musl` target declares that everything must be statically linked. But `RUSTFLAGS="-C target-feature=-crt-static"` tells `rustc` to make an exception to that declaration and dynamically link `libc`. So, the conclusion is that everything is statically linked except for `libc`, which is dynamically linked.

spl:
Since I don't think it's easy to resolve the `proc-macro` issue and keep static builds on `musl`, I took the liberty of cleaning up this PR to remove the `headers` dependency that was causing the problem. I also removed `hyper_seekable` and a comment about `hyper` headers. (If you want to keep `hyper_seekable` around, perhaps something like a Gist would be better?)

As far as I'm concerned, this PR is ready to go.

mati865:
> To clarify what I understand, the `x86_64-alpine-linux-musl` target declares that everything must be statically linked. But `RUSTFLAGS="-C target-feature=-crt-static"` tells `rustc` to make an exception to that declaration and dynamically link `libc`. So, the conclusion is that everything is statically linked except for `libc`, which is dynamically linked.

It's not entirely true, see https://github.com/rust-lang/rust/pull/55566.
Once that gets merged `x86_64-alpine-linux-musl` will truly statically link libs, passing `-C target-feature=-crt-static` will revert to current behaviour.

mati865:
> However, the gist of preceding comments seem to imply that the PR'd solution was not preferred over the one in rust-lang/rust#58575, to which I linked above. But I'm sure I'm missing a lot of context here.

Discussions about musl target desing were scattered all over the place so it's hard to get what is going on.

> The summary in the last comment, rust-lang/rust#55566 (comment), if true, sounds attractive.

That's the current standing. It won't fix Procedural Macros though.

pkgw:
@spl I've locally rebased this onto current `master`; mind if I force-push to your branch? I think it's time to go ahead and merge this.

spl:
No, go for it.

pkgw:
Ah, it seems that the static target once again has a proc-macro issue, due to `failure_derive`, which is pulled in by `cookie_store`.

pkgw:
Ohhhhhhh — I figured something out about the MUSL issue. It might have been obvious to other people, but it wasn't to me, so in case it comes in handy:

What I didn't get, until now, is that one of the key facets of the proc-macro problem is the distinction between the cross-compilation use case and the ... non-cross-compilation one. Proc-macro crates have to be dynamically linked into the compiler. When you're cross-compiling, that means that they're built for the *build* platform/toolchain not the *host* platform/toolchain. So if you're cross-compiling for MUSL from a regular Linux host, things work fine. It gets sticky when, as our build system currently tries, both "host" *and* "build" are static/MUSL.

I am pretty sure that if we could set things up to compile the static executable via cross-compilation, it would work. I don't know if we can set things up to be able to do that while also pulling in external static libraries such as Freetype.

pkgw:
12 hours later ... it works! I have a proof-of-concept that creates a static executable *and* includes the procedural macro features. So @spl go ahead and put `headers` feature back if you'd like.

Bad news #1: the executable that I'm creating right now weighs in at 90 MB.

Bad news #2: I couldn't figure out a way to get it working that didn't involve manually cross-compiling all of our system deps in the musl target, so that's fun.

It's not pretty, but this is a problem that a lot of people have been struggling with: see for instance emk/rust-musl-builder#65, rust-lang/rust#36710 . So I'm pretty happy that I figured something out here. Code forthcoming ...

pkgw:
Even better version pushed! We're not manually compiling the deps anymore, and the final executable is down to 60 MB.

pkgw:
The Travis failure is some unrelated, probably transient, issue on macOS, and Circle has a QEMU timeout. The static build on Travis succeeded, proc macros and all ([log](https://travis-ci.org/tectonic-typesetting/tectonic/jobs/547200704)). Going to go ahead and merge!

spl:
I haven't looked at everything you've done, @pkgw, but it certainly sounds good.

As an aside, you might want to update the Travis-CI macOS image to `osx_image: xcode10.2` to get the latest `harfbuzz`.

PR #336, opened 2019-03-18
cargo clippy: fix redundant_field_names

spl:
I started looking at [`clippy`](https://github.com/rust-lang/rust-clippy) output. There are a large number of warnings. Rather than do them all and have an overwhelming PR, I started with `redundant_field_names`.

I think it would be good to eventually add `cargo clippy` to CI, but, first, I'd like to get feedback on this bit of lint fixing.


pkgw:
Thanks for taking a look at this! As you might guess, I have not been using clippy, but it would be good to tidy these things up.

PR #341, opened 2019-03-19
cargo clippy: fix unused_io_amount

spl:
Fixes [`unused_io_amount`](https://rust-lang.github.io/rust-clippy/master/index.html#unused_io_amount)

Note that this change may affect your logic. But, considering you're not checking for the number of bytes written by `io::Write::write`, I'm guessing it's safe.


pkgw:
Yes, this looks like a genuine bugfix.

PR #345, opened 2019-03-19
cargo clippy: fix needless_lifetimes

spl:
Fixes [`needless_lifetimes`](https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes)

### DIFF
--- a/src/io/setup.rs
+++ b/src/io/setup.rs
@@ -32,7 +32,7 @@ pub struct IoSetup {
 }
 
 impl IoSetup {
-    pub fn as_stack<'a>(&'a mut self) -> IoStack<'a> {
+    pub fn as_stack(&mut self) -> IoStack {
         let mut providers: Vec<&mut IoProvider> = Vec::new();
 
         if let Some(ref mut p) = self.genuine_stdout {


pkgw:
Hmm, will this limit our compatibility with older versions of Rust? I have a sense that over time, more lifetime elisions have been added, and maybe this is something that relies on relatively new lifetime elision support.

NB: I haven't tracked the minimum version of Rust on which Tectonic actually compiles. Would be good to know what that is.

pkgw:
@spl PS, thanks for the PRs! I appreciate you breaking them up by topic — makes review a lot easier.

spl:
> Hmm, will this limit our compatibility with older versions of Rust? I have a sense that over time, more lifetime elisions have been added, and maybe this is something that relies on relatively new lifetime elision support.

I'm not sure. The issue that apparently spawned the `needless_lifetimes` lint is https://github.com/rust-lang/rust-clippy/issues/115, and it's from 2015. I searched a bit for lifetime elision in Rust, but I didn't find anything about when or how it might have changed over time.

> NB: I haven't tracked the minimum version of Rust on which Tectonic actually compiles. Would be good to know what that is.

Definitely good to know.

For an application (and not so much for a library), I don't think you need to worry too much about supporting older versions of Rust. This is especially true considering how easy it is to update Rust components and how often the Rust environment is rolling out improvements. With C/C++ on the other hand, you might try supporting a wide range of compilers, but the environment is pretty slow-changing.

To start with, you might consider updating to the 2018 edition, for which you would need Rust 1.31 released in December 2018.

pkgw:
@spl I don't worry a ton about maintaining compat with old Rusts, but it's also true that not all elements of the ecosystem are always running the latest stable. For instance, our bigendian test VM uses a custom Docker image preloaded with the toolchain, and it's a bit tedious to update so that it often lags behind the latest releases.

As for this particular PR, the comments motivating the lint make it sound to me like this particular elision has been OK since 2015, so I'll go ahead and merge.

I'm interested in upgrading Tectonic to the 2018 edition, but as long as my bandwidth for hacking has been low I've been content to leave it as-is. I'm pretty sure that edition-bumping would require a rebuild of the bigendian Docker image, but other than that I'm not aware of any strong reasons not to go ahead and do it whenever someone has the motivation.

PR #349, opened 2019-03-20
cargo clippy: allow vec_box

spl:
Don't warn about [`vec_box`](https://rust-lang.github.io/rust-clippy/master/index.html#vec_box).

### DIFF
--- a/src/engines/mod.rs
+++ b/src/engines/mod.rs
@@ -124,7 +124,9 @@ struct ExecutionState<'a, I: 'a + IoProvider> {
     io: &'a mut I,
     events: &'a mut IoEventBackend,
     status: &'a mut StatusBackend,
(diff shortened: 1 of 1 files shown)


spl:
This is the last of the `clippy` lints.

pkgw:
OK, I believe that the suppression functionality used here was introduced in 1.30: [related doc issue](https://github.com/rust-lang-nursery/reference/issues/526). That version was released on 2018-10-25. This might cause problems for older Rusts, but (as discussed in #345) we don't have a specific policy about what versions we support, so ¯\_(ツ)_/¯.

spl:
Oh, that's pretty new. It would be nice to see the Rust version mentioned
in the clippy documentation for that feature. But I guess the Rusty world
just keeps rolling along...

PR #350, opened 2019-03-20
Suppress unknown warnings on Clang, alternate version

pkgw:
Inspired by #328 by @spl, but takes a different tack. The limitation with the other PR is that the header file didn't actually suppress the warnings on GCC! At the moment, I think it's just simpler to gate the pragmas by compiler explicitly.

This PR also includes a fix against a race condition in the test suite.


spl:
That works. I would only suggest adding to your comment the reason why _not_ Cang: e.g. to avoid Clang warning about unknown warnings.

PR #352, opened 2019-03-22
cargo clippy tests

spl:
Okay, so now I've figured out how to `cargo clippy` the tests (and all other targets), and these are the results. The warning type fixes are broken down into separate commits. All changes are, I think, pretty innocent.

### DIFF
--- a/tests/executable.rs
+++ b/tests/executable.rs
@@ -41,12 +41,11 @@ fn prep_tectonic(cwd: &Path, args: &[&str]) -> Command {
         .join("tectonic")
         .with_extension(env::consts::EXE_EXTENSION);
 
-    match fs::metadata(&tectonic) {
-        Ok(_) => {}
-        Err(_) => panic!(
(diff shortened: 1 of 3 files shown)


spl:
One of the annoying things about `clippy` is that it does not visit files that have already been `check`ed. So you can `touch` files to force it to see them. But I never did that with the `xdv` directory... until now. In the future, keep this command in mind:

```
$ find src tests xdv -type f -exec touch {} +; cargo clippy --all-targets --all-features
```

spl:
There are perhaps a lot of changes here. Each commit is isolated and commutative with the other commits, so you can consider them separately. I'm not sure 12 separate PRs are better than this. If prefer a different form of submission, let me know.

pkgw:
Yeah, 12 separate PRs would probably have been a bit overkill. This was not hard to review as presented. Thanks for all your work on cleaning up the codebase!

PR #351, opened 2019-03-22
Silence error-chain deprecation warnings

spl:
See the comment in the code.

PR #353, opened 2019-03-22
Run clippy on Travis-CI

spl:
This causes Travis-CI to fail if `clippy` finds any warning lints. It will, of course, fail now, since there are warnings to be found. But if #351 and #352 are merged and this branch is updated, it will pass.

### DIFF
--- a/dist/travis.sh
+++ b/dist/travis.sh
@@ -22,6 +22,7 @@ set -e
 
 # Helpful context.
 
+echo "TRAVIS_ALLOW_FAILURE: $TRAVIS_ALLOW_FAILURE"
 echo "TRAVIS_BRANCH: $TRAVIS_BRANCH"
 echo "TRAVIS_BUILD_ID: $TRAVIS_BUILD_ID"
 echo "TRAVIS_COMMIT: $TRAVIS_COMMIT"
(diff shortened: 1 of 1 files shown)


pkgw:
#351 and #352 are now merged, so if you rebase onto master we should be able to see how this goes. (I tried to do it for you but I guess I haven't enabled the feature that lets maintainers edit other people's PRs directly for this repo.)

spl:
Rebased.

spl:
Oh, I just noticed this on Travis-CI:

> component 'clippy' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel 'nightly'

spl:
I guess `clippy` isn't always available for `nightly`. See this [`rustup` components history](https://rust-lang.github.io/rustup-components-history/index.html). I'm not sure what “present” and “missing” mean. Perhaps it's missing if it fails to build? Perhaps `rustup component add clippy` should not be run for the `nightly` CI.

pkgw:
@spl Yes, I believe that for non-core components such as clippy, it's not guaranteed that they'll be available on nightly. Based on the small sample size, it looks like that happens non-negligibly often, but nightly failures are OK anyway so I don't think we need to do anything special to deal with those failures. Stable and beta will always have clippy, I think.

spl:
I think it's better to to avoid spurious failures even for `nightly`. That way, a (non-network-related) failure is a valid failure that may deserve inspection. So, I pushed the `rustup` operations after the `$is_main_build` check to avoid fetching `rustfmt` and `clippy` for other CI runs.

pkgw:
:+1:

PR #354, opened 2019-03-26
Install Homebrew packages in .travis.yml

spl:
Use the Travis-CI [`homebrew` add-on](https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-macos) to `brew install` packages on macOS.

Comments:
* This does not do a `brew update` since that slows down the build quite a bit.
* I don't think the `--force` flag is necessary for `brew install openssl`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,10 @@ addons:
       - libssl-dev
       - openssl
       - zlib1g-dev
+  homebrew:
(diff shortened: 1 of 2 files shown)


pkgw:
Looks like this shaves the macOS build from ~15 minutes to ~10, not bad!

PR #355, opened 2019-03-26
cargo update

spl:
There are some updates that showed up in #330, but they're not relevant to that change. I guess that happened when I changed the dependencies. So I'd suggest merging this first.

```
Updating cfg-if v0.1.6 -> v0.1.7
Updating crc32fast v1.1.2 -> v1.2.0
Updating openssl v0.10.19 -> v0.10.20
Updating openssl-sys v0.9.42 -> v0.9.43
Updating rand_os v0.1.2 -> v0.1.3
Updating syn v0.15.27 -> v0.15.29
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,7 +65,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "autocfg 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "backtrace-sys 0.1.28 (registry+https://github.com/rust-lang/crates.io-index)",
- "cfg-if 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cfg-if 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.50 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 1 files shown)


pkgw:
Thanks! If you haven't seen, we do use [Dependabot](https://dependabot.com/), so in principle we should get automated updates of dependencies with independent PRs to check for CI problems. But it's also true that sometimes Dependabot seems not to bump certain deps even though a manual `cargo update` will bring them up.

spl:
Thanks. I was aware of the use of Dependabot, though I don't know how it's configured. Some of these deps were updated when I was working on #330, which was over a week ago, and Dependabot hasn't PR'd anything about them. I'm not sure how it decides when to PR or when not to.

Seems like there was an `osx` issue on Travis-CI. I just rebased to `master` to see if that will fix it.

spl:
It now passes CI.

@pkgw Just to be clear, would you prefer individual PRs with each dependency changed? I have no problem with that, though it is quite a bit more work.

While investigating Dependabot myself for the first time, I tried the “Bump now” button for my fork of the `tectonic` repository. It took a while, but it did produce PRs of a large number of dependency bumps, including the ones I bumped here.

If you prefer Dependabot to do the work, you might try the “Bump now” button.

pkgw:
@spl Nah, if a human's doing it, separating out each dep is overkill.

PR #360, opened 2019-04-01
Neaten the dist/travis.sh presentation

spl:
* Make the Travis-CI log easier to follow by reducing the noise:
  - put more things behind folds
  - put folds inside if-then blocks to avoid printing `'+ false'`, etc.
  - make fold titles simpler and more consistent
  - avoid printing the values already found in the printed env vars
* Document the `travis_fold` usage

### DIFF
--- a/dist/travis.sh
+++ b/dist/travis.sh
@@ -18,10 +18,50 @@
 #
 # So, it's complicated.
 
+# Exit immediately if a command exits with a non-zero status.
 set -e
+
+echo ""
+
+# We use `travis_fold` commands to hide chunks of the Travis-CI log. Follow the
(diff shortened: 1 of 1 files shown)


pkgw:
I had chosen to put the fold markers outside of the `if` statements because it helped me understand why certain steps were running, or not running, when I straightened out the Travis infrastructure, but at this point I think the script is a lot better-structured so those markers do seem less necessary.

spl:
@pkgw Sure, I understand. You can always add more information to the fold titles if it would help you. I would just recommend making the text more meaningful than `true` or `false` so that one doesn't have to look at the script source to figure out what variables are being evaluated to get those values.

PR #387, opened 2019-06-03
replace usage of deprecated tempdir crate with tempfile.

ratmice:
It's only used in tests, but the tempdir crate has been deprecated here:
https://github.com/rust-lang-deprecated/tempdir

This updates the tests to use tempfile.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -523,18 +523,6 @@ dependencies = [
 
 [[package]]
 name = "rand"
-version = "0.4.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "fuchsia-cprng 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.55 (registry+https://github.com/rust-lang/crates.io-index)",
- "rand_core 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "rdrand 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
-name = "rand"
 version = "0.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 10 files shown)


spl:
I was just looking into making this change in another context (`cc`). I didn't end up doing so because `tempdir` is not broken and `tempfile` brings in a few more dependencies. But for Tectonic, I think it's not a big issue (either option).

ratmice:
This actually ends up removing quite a few dependencies,
tectonic already depends upon tempfile, tempdir not being updated anymore was causing it to bring in an older/duplicate copy of rand 0.4.6.

spl:
@ratmice Sounds good then!

spl:
Hmm, that makes me think that it may actually be a good idea to change `cc` in order to be in sync with the rest of the ecosystem, in which downstream dependencies will probably use `tempfile` over `tempdir`.

pkgw:
Thanks for submitting! This looks like a good change.

CI is currently failing due to an unrelated issue with `rustfmt` finding the `util` module used in our test suite. I feel like that shouldn't be hard to fix, but someone will need to sit down and figure out what's going on. I'd prefer to get the CI fixed before merging just because it's something we should work out anyway.

(Clicking through to the Travis logs also reveals another unrelated issue, that our static builds aren't working right now due to what looks like a libpng versioning issue.)

ratmice:
@pkgw Sorry I hadn't seen this, I upgraded to the latest rustfmt, and then could reproduce it.
I at least found a work around for this, let me know if that is not the right fix, I can back it out.

pkgw:
@ratmice That's a fine solution as far as I'm concerned!

Annoyingly there is another janitorial bug unrelated to this PR that's popping up in the CI ([logs](https://travis-ci.org/tectonic-typesetting/tectonic/jobs/540817083#L498-L504)). Fortunately it should be a super easy fix:

```
error: lint `clippy::cyclomatic_complexity` has been renamed to `clippy::cognitive_complexity`
   --> xdv/src/lib.rs:278:13
    |
278 |     #[allow(clippy::cyclomatic_complexity)]
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use the new name: `clippy::cognitive_complexity`
    |
    = note: `-D renamed-and-removed-lints` implied by `-D warnings`
error: lint `clippy::cyclomatic_complexity` has been renamed to `clippy::cognitive_complexity`
   --> xdv/src/lib.rs:278:13
```

If it's easy for you to make this change too, hopefully that should unstick things. If not, I can do it — it shouldn't be your responsibility to clear out these CI issues.

pkgw:
Great, thank you for going the extra mile!

PR #388, opened 2019-06-09
Cleanup rust warnings and update to 2018 edition

Mrmaxmeier:
`rustc v1.37` introduces a warning about the bare trait syntax being deprecated.
The [new `dyn Trait` syntax](https://doc.rust-lang.org/edition-guide/rust-2018/trait-system/dyn-trait-for-trait-objects.html) was introduced in rust `rustc v1.27`.
This also adds the `edition = "2018"` flag to the `Cargo.toml`.
The main differences between the 2015 and 2018 edition are changes to the [path resolution](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ readme = "README.md"
 keywords = ["tex", "latex", "typesetting", "font"]
 categories = ["command-line-interface", "parser-implementations", "rendering", "science", "text-processing"]
 license = "MIT"
+edition = "2018"
 
 [badges]
 travis-ci = { repository = "tectonic-typesetting/tectonic" }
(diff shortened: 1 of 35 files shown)


pkgw:
Thanks for submitting! I'm at a conference this week so it will be even slower than usual to review, but it's on the list. Feel free to ping me if it takes more than a week or so to start processing this.

spl:
After a skim, the changes look straightforward (if tedious!) to me. @pkgw previously mentioned a concern about upgrading the edition in https://github.com/tectonic-typesetting/tectonic/pull/345#issuecomment-474664436.

pkgw:
It's been long enough since the 2018 edition has come out that I think we're ready to make the leap. Thanks @Mrmaxmeier and @spl for reviewing!

PR #394, opened 2019-06-18
Update osx_image on Travis-CI

spl:


pkgw:
Thanks, and I see that Travis has already passed.

Issue #318, opened 2019-02-27
Error building: hb-graphite2.h: no such file or directory.


alejandrogallo:
I followed the instructions on the website for ubuntu...

I recorded an asciinema session to show.

https://asciinema.org/a/S1OwTeDe20AAXWQFFbA4Lw4DV

pkgw:
What version of Ubuntu are you using? It looks like your OS has version 1.0.1 of `libharfbuzz-dev`, which is quite old.

We actually provide a custom package of harfbuzz for Xenial for use in our CI systems for just this reason: [see this PPA](https://launchpad.net/~k-peter/+archive/ubuntu/tectonic-ci).

alejandrogallo:
Hi!
thank you for the answer, yes I'm running Xenial 16.04.6, I'm going to test the ppa now.

For more information, an extract of `/etc/os-release`:
```
VERSION="16.04.6 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.6 LTS"
```

pkgw:
Thanks! What's strange to me is that the Tectonic `build.rs` should try to require `harfbuzz >= 1.4`, so I'm not sure how your build got to the point where this error was triggered. If I try the procedure that you ran in a fresh Xenial Docker container, I get the intended error:

```
   Compiling serde_derive v1.0.88
   Compiling tectonic v0.1.11
error: failed to run custom build command for `tectonic v0.1.11`
process didn't exit successfully: `/tmp/cargo-install2AQTgw/release/build/tectonic-82edf7d300c7d38a/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Failure { command: "\"pkg-config\" \"--libs\" \"--cflags\" \"fontconfig harfbuzz >= 1.4 harfbuzz-icu icu-uc freetype2 graphite2 libpng zlib\"", output: Output { status: ExitStatus(ExitStatus(256)), stdout: "", stderr: "Requested \'harfbuzz >= 1.4\' but version of harfbuzz is 1.0.1\n" } }', src/libcore/result.rs:1009:5
```

Maybe your environment is a bit unusual where you've got a newer, local Harfbuzz installed such that `pkg-config` finds it, but it doesn't have graphite2 enabled?

PR #395, opened 2019-06-18
terrible hacks for depending upon crateized harfbuzz/graphite2.

ratmice:
This is a pretty terrible hack, mainly meant as an RFC/not intended to be merged in it's current form.

I was working on getting CI up for my project which depends tectonic,
Somwhere in that process I figured it'd be really nice if we could just depend upon a harfbuzz crate,
and let cargo sort out the dependencies for that system library.

Tectonic itself depends upon harfbuzz and graphite2 directly,
It also relies upon the headers hb-icu.h and hb-graphite2.h, so requires a harfbuzz which has been built against ICU and graphite2.

(So as such graphite is both a direct, and transitive dependency through harfbuzz).

Anyhow, most of the work that needs to be done is probably in the harfbuzz crate itself,
but the ICU/graphite2 crates required some changes as well.

But at least here is a hack which works on exactly 1 system configuration i have tried
when there is no system harfbuzz headers installed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -487,7 +487,7 @@ dependencies = [
 [[package]]
 name = "harfbuzz-sys"
 version = "0.3.2"
-source = "git+https://github.com/ratmice/rust-harfbuzz.git#d18b1a61d8d5cf637fc8da9702ad6d232361d343"
+source = "git+https://github.com/ratmice/rust-harfbuzz.git#4a94c641b54f99e4471bef83bacc22140bff042a"
 dependencies = [
  "cmake 0.1.40 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-graphics 0.17.3 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,6 +78,11 @@ dependencies = [
 
 [[package]]
 name = "base64"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "base64"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
@@ -487,8 +492,9 @@ dependencies = [
 [[package]]
 name = "harfbuzz-sys"
 version = "0.3.2"
-source = "git+https://github.com/ratmice/rust-harfbuzz.git#7aeb79351c028f71856d8f2fc894c476a6cd489e"
+source = "git+https://github.com/ratmice/rust-harfbuzz.git#f0cc35b16708a6b5f00ffeda28351a32893f1144"
 dependencies = [
+ "base64 0.10.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "cmake 0.1.40 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-graphics 0.17.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-text 13.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -497,6 +503,7 @@ dependencies = [
  "graphite2-sys 1.3.13 (git+https://github.com/ratmice/graphite2-sys-rust.git)",
  "icu-sys 0.1.0 (git+https://github.com/ratmice/rust-icu.git)",
  "pkg-config 0.3.14 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -1366,6 +1373,7 @@ version = "0.1.12-dev"
 dependencies = [
  "aho-corasick 0.7.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "app_dirs 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "base64 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "cc 1.0.37 (registry+https://github.com/rust-lang/crates.io-index)",
  "clap 2.33.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "error-chain 0.12.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -1381,6 +1389,7 @@ dependencies = [
  "regex 1.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "reqwest 0.9.18 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.92 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
  "sha2 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "tectonic_xdv 0.1.9-dev",
  "tempfile 3.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -1790,6 +1799,7 @@ dependencies = [
 "checksum autocfg 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)" = "0e49efa51329a5fd37e7c79db4621af617cd4e3e5bc224939808d076077077bf"
 "checksum backtrace 0.3.30 (registry+https://github.com/rust-lang/crates.io-index)" = "ada4c783bb7e7443c14e0480f429ae2cc99da95065aeab7ee1b81ada0419404f"
 "checksum backtrace-sys 0.1.28 (registry+https://github.com/rust-lang/crates.io-index)" = "797c830ac25ccc92a7f8a7b9862bde440715531514594a6154e3d4a54dd769b6"
+"checksum base64 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)" = "a51012ca17f843e723dedc71fdd7feac9d8b53be85492aa9232b2da59ce6bb3b"
 "checksum base64 0.10.1 (registry+https://github.com/rust-lang/crates.io-index)" = "0b25d992356d2eb0ed82172f5248873db5560c4721f564b13cb5193bda5e668e"
 "checksum bitflags 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)" = "3d155346769a6855b86399e9bc3814ab343cd3d62c7e985113d46a0ec3c281fd"
 "checksum block-buffer 0.7.3 (registry+https://github.com/rust-lang/crates.io-index)" = "c0940dc441f31689269e10ac70eb1002a3a1d3ad1390e030043662eb7fe4688b"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -487,7 +487,7 @@ dependencies = [
 [[package]]
 name = "harfbuzz-sys"
 version = "0.3.2"
-source = "git+https://github.com/ratmice/rust-harfbuzz.git#4a94c641b54f99e4471bef83bacc22140bff042a"
+source = "git+https://github.com/ratmice/rust-harfbuzz.git#7aeb79351c028f71856d8f2fc894c476a6cd489e"
 dependencies = [
  "cmake 0.1.40 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-graphics 0.17.3 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -487,7 +487,7 @@ dependencies = [
 [[package]]
 name = "harfbuzz-sys"
 version = "0.3.2"
-source = "git+https://github.com/ratmice/rust-harfbuzz.git#a51ce5858d9304c1440181f52847f60f535f550c"
+source = "git+https://github.com/ratmice/rust-harfbuzz.git#d18b1a61d8d5cf637fc8da9702ad6d232361d343"
 dependencies = [
  "cmake 0.1.40 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-graphics 0.17.3 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -184,6 +184,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "cmake"
+version = "0.1.40"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "cc 1.0.37 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "cookie"
 version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


pkgw:
Thanks for submitting! This is certainly an area where I wish we had a bit of a nicer story.

Do the `DEP_*` environment variables get set automatically when `harfbuzz-sys` is included as a dependency, or is the idea that they'd be set externally when building Tectonic?

There are times when I start feeling that maybe we should just bundle the source code for Harfbuzz, Graphite, and maybe some of the other deps, and compile them all ourselves. That does not feel awesome, but to be honest it might make life easier in some ways — those libraries are, fortunately, nice and straightforward to compile.

The other thing is that if we could isolate the cross-coupling between Tectonic's C/C++ code and the headers of Harfbuzz and friends, that would simplify some things. That ties into a somewhat distinct goal: I'd like to start replacing common subsystems in Tectonic's C/C++ code (e.g., image loading) with versions written in Rust, and due to the spaghetti nature of the code, those subsystems would have to be callable from C/C++. So you could imagine a low level foundation written in Rust, that generates a C header file, that the C/C++ code compiles against, that then is compiled into the final Rust crate. And you could further imagine having that low level foundation depend on Harfbuzz and friends, adding yet another layer to the stack. Convoluted, but I have a feeling that might be the only path forward to Rustifying large chunks of the code. (See @Mrmaxmeier's #246 for an example of making Rust callable from C.)

ratmice:
> Thanks for submitting! This is certainly an area where I wish we had a bit of a nicer story.
> 
> Do the `DEP_*` environment variables get set automatically when `harfbuzz-sys` is included as a dependency, or is the idea that they'd be set externally when building Tectonic?

These are set by cargo automatically,
Lines from harfbuzz-sys/build.rs like:
```rust
    println!("cargo:include={}", out_dir.join("include").join("harfbuzz").display());
```
will correlate to DEP_HARFBUZZ_INCLUDE, somehow cargo will pull the harfbuzz out of name of the library being linked, so this sets the environment variable:

DEP_HARFBUZZ_INCLUDE=/path/to/outdir/include/harfbuzz

in my patch I used a few non-standard variables (include_flags, link_flags, libdir_flags),
these are essentially derived from cargo by the line:
println!("cargo:include_flags={}", include_flags);

I used these nonstandard flags, because the standard cargo:include I think wants a single directory
but when things come from pkg-config and the like there may be multiple -I/-L directories or -l libs.

> There are times when I start feeling that maybe we should just bundle the source code for Harfbuzz, Graphite, and maybe some of the other deps, and compile them all ourselves. That does not feel awesome, but to be honest it might make life easier in some ways — those libraries are, fortunately, nice and straightforward to compile.

Indeed i think this would help, I feel fairly soundly defeated at the moment by the system distributed versions of these libraries in the failed attempts to build on the CI where system libraries are being found.

On the windows CI the system harfbuzz seems to link to glib, but I hadn't figured out where to get a -lglib flag from, On linux/pkg-config, the existing pkg-config probe's return *before* println'ing any of those environment variables (and thus they don't get set by cargo). Making that pkg-config branch println stuff/to export DEP_* I encounter some linkage issue I don't yet fully comprehend.

So at least for the moment being it'd be great if I could establish a baseline of just bundling all the sources, then perhaps add system-library stuff back in on top of that.

> The other thing is that if we could isolate the cross-coupling between Tectonic's C/C++ code and the headers of Harfbuzz and friends, that would simplify some things. That ties into a somewhat distinct goal: I'd like to start replacing common subsystems in Tectonic's C/C++ code (e.g., image loading) with versions written in Rust, and due to the spaghetti nature of the code, those subsystems would have to be callable from C/C++. So you could imagine a low level foundation written in Rust, that generates a C header file, that the C/C++ code compiles against, that then is compiled into the final Rust crate. And you could further imagine having that low level foundation depend on Harfbuzz and friends, adding yet another layer to the stack. Convoluted, but I have a feeling that might be the only path forward to Rustifying large chunks of the code. (See @Mrmaxmeier's #246 for an example of making Rust callable from C.)

Indeed I haven't much looked in that regard, A bit afraid that snowball might start rolling downhill :mountain_snow:

spl:
@ratmice I see you've [forked](https://github.com/ratmice/graphite2-sys-rust) my [`graphite2-sys-rust`](https://github.com/spl/graphite2-sys-rust) repository. I've actually been working on that explicitly for supporting Tectonic. I'm also still working on it, so it's not ready to be released. I plan on doing some other `-sys` crates before introducing it to Tectonic.

spl:
> There are times when I start feeling that maybe we should just bundle the source code for Harfbuzz, Graphite, and maybe some of the other deps, and compile them all ourselves. That does not feel awesome, but to be honest it might make life easier in some ways — those libraries are, fortunately, nice and straightforward to compile.

@pkgw The nice thing about putting them in crates (in the way I see it at least) is that the crate allows you to choose whether to use a bundled, statically linked library or a shared, dynamically linked library. Thus, you can configure this transitively from the `tectonic` crate. Plus, all the build settings are localized to the crate itself instead of being lumped in with the rest of  Tectonic.

ratmice:
@spl sorry I should have mentioned that this entailed a fork of your work earlier. I was mostly just curious how close we were from a standalone 'cargo install', and tend to work from the backwards from the goal.  I don't mean to step on any toes, and apologize if that is the case.

spl:
@ratmice No offense taken. I just wanted to note that it's not ready, yet.

I have also been working through the issues with `graphite2` to figure out how to successfully install it on different targets and with different mechanisms (static, shared, `pkg-config`, `vcpkg`, `homebrew`, etc.) Once I'm done there, I plan to apply the same approach to other libraries: `freetype2`, `harfbuzz`, `icu`. Once that is done, I believe it will be ready for Tectonic.

pkgw:
@spl Ah, cool to hear that you've been looking at this problem too! I've always believed that the fundamental blocker on using dep crates like `harfbuzz-sys` has been getting the compiler flags so that Tectonic's C/C++ code can find the necessary headers, but I didn't know abut the `DEP_*` variables. It sounds like those might provide a nice path forward.

An extra Tectonic-specific wrinkle might be that we need a Harfbuzz that's been built with Graphite2 support. I do not know how common it is for packagers to include that feature.

spl:
> I've always believed that the fundamental blocker on using dep crates like `harfbuzz-sys` has been getting the compiler flags so that Tectonic's C/C++ code can find the necessary headers, but I didn't know abut the `DEP_*` variables. It sounds like those might provide a nice path forward.

@pkgw Yes, a `DEP_` env var works for the include file path.

> An extra Tectonic-specific wrinkle might be that we need a Harfbuzz that's been built with Graphite2 support. I do not know how common it is for packagers to include that feature.

FYI, the [`servo/rust-harfbuzz`](https://github.com/servo/rust-harfbuzz) crate includes the following `configure` flags, among others: `--without-icu --without-freetype --without-glib`

After finishing `graphite2`, I will look into `freetype`, `icu`, and `harfbuzz` in a way that allows them to be configured and used by Tectonic (as well as others, of course). There are existing crates for these, but I need to look into them to see if they can be used. I started with `graphite2` since it doesn't have any dependencies.

pkgw:
> FYI, the [`servo/rust-harfbuzz`](https://github.com/servo/rust-harfbuzz) crate includes the following `configure` flags, among others: `--without-icu --without-freetype --without-glib`

Yeah, that kind of thing worries me. In the end, it might make sense for us to bundle our own harfbuzz, but have the rest as external dependencies. Based on my packaging experience, the prospect of going around and trying to ensure that a zillion packaging systems all include the right sub-deps in their harfbuzzes seems very unappealing.

ratmice:
yeah, the https://github.com/ratmice/rust-harfbuzz currently does an DEP_ var hack for configuring --with-icu --with-graphite, I think harfbuzz is somewhat painful in the aspect @pkgw mentioned in that even if we can find it from a system library, we still need to do some configure-esque check that it comes with hb-graphite2.h (there is no harfbuzz-graphite.pc like there is for harfbuzz-icu.

The easiest way I know of to do that is to replace it's makefile.cargo with an actual configure script of some sort and perhaps link a dummy library libtectonic_harfbuzz which basically contains nothing but links to the real system harfbuzz if it's built the way we need it, or builds and links to the the "fallback" harfbuzz if it is unavailable

pkgw:
@ratmice I like that idea a lot!

spl:
> The easiest way I know of to do that is to replace it's makefile.cargo with an actual configure script of some sort and perhaps link a dummy library libtectonic_harfbuzz which basically contains nothing but links to the real system harfbuzz if it's built the way we need it, or builds and links to the the "fallback" harfbuzz if the system harfbuzz is unusable.

@ratmice I'm not sure if I completely understand what you mean here, but my plan to create a `harfbuzz-sys` crate that allows you to configure it in the Rust/Cargo way similar to my `graphite2-sys` crate. That is, there is a general flow of trying `pkg-config`, `vcpkg`, etc. and a fallback of a static build. It will also have the option of depending on `graphite2-sys`. Then, there is the option to choose one of the build approaches if you know which one you want. I believe the way I'm doing it is the best in the long run, even it has taken me some time to make sure it works on different targets and package systems.

ratmice:
@spl I will try and explain the troubles i've had with the pkg-config approach (I don't know anything about vcpkg).  Essentially the granularity of pkg-config is the library version in the .pc file, but tectonic depends upon a specific configuration (with graphite2, which is optional and defaults to configuring without it).

So if the pkg-config/vcpkg picks up an installed version of the library, configured in the wrong way, tectonic is going to fail to build.  What was making me doubt the general flow you describe is essentially even when it's found we need to then check everything and potentially provide the fallback.

I definitely agree with your approach, systems like debian are going to frown upon non-system libraries without a real good excuse, earlier in the thread I was basically lamenting the difficulty of providing all the lookup, only to potentially have to fallback to the static build anyways.  But it really does seem like the right thing to do.

spl:
@ratmice Okay, I'm understanding you better. Thanks for that explanation.

Can we identify which systems have HarfBuzz packages with a configuration that won't work for Tectonic?

ratmice:
@spl I don't know what systems would provide one compiled that way,
but I can definitely come up with something that checks it at build time.  Using AC_CHECK_HEADER, this basically runs the cpp with the `#include` and checks the result, if i recall.

I'm hoping that this kind of approach won't interfere with your vcpkg plans (I imagine the static harfbuzz build machinery already relies upon similar checks).

pkgw:
@spl @ratmice Yeah, I think we'd really need a build-time check to decide if the system Harfbuzz was acceptable or not. I'm confident we could devise something that would work in a `build.rs` script, doing the equivalent of Autoconf's `AC_CHECK_HEADER`, `AC_TRY_LINK`, etc.

spl:
> I can definitely come up with something that checks it at build time. Using AC_CHECK_HEADER, this basically runs the cpp with the `#include` and checks the result, if i recall.

Does that only work if the target is the host machine? It would be preferable to have a mechanism that supported cross-compiling.

crlf0710:
@spl if system harfbuzz was not configured satisfying tectonic usage, `harfbuzz/hb-graphite2.h` and `harfbuzz/hb-icu.h` will not be packaged/installed to INCLUDEDIR.

pkgw:
@spl That's actually not any more difficult, I think — the checks are basically test compilations using exactly the same framework that would be used to compile `tectonic/*.c`. You can't *run* any executables that you produce when you're cross-compiling, but you just need to see if they link successfully.

spl:
But what if the actual target machine has HarfBuzz configured differently? I feel like I'm missing something.

crlf0710:
@spl yes, the system harfbuzz is likely dynamic-linked, and if it is configured and built differently, the program will fail to start due to dynamic linking failure. This happens after the successful build and deploy of the program. The rust `openssl`-binding [library](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/Cargo.toml#L15) suffers from the same program, where they give a separate crate feature called "vendored" to explicitly request the source code of openssl be built locally and statically.

pkgw:
But it's worth pointing out that this can happen with a non-cross build too — the risk is inherent if you're transferring a dynamically linked binary from one machine to another. When you do a cross-compile, your build machine still has to have copies of all the required shared libraries in the target architecture, so as long as those copies are representative of what people on the target platform generally have, there shouldn't be any *additional* wrinkles.

spl:
@crlf0710 Right, that's the static-linking fallback, in which case we don't have the problem.

I guess I'm trying to figure out when will there be a problem with a shared HarfBuzz that is configured without support for ICU, Graphite, etc. I haven't seen any Linux/Homebrew/vcpkg system packages configured that way. So I can only imagine that it happens when HarfBuzz is built and installed manually. I wonder if it's worth the trouble to try to detect that situation.

crlf0710:
For one case, on archlinux, `harfbuzz` and `harfbuzz-icu`(integration) are separate packages.
https://www.archlinux.org/packages/extra/x86_64/harfbuzz/
https://www.archlinux.org/packages/extra/x86_64/harfbuzz-icu/

Maybe someone has installed the former but not the later? I've never experimented with that though.

spl:
@crlf0710 Great! Thanks for the example. Shouldn't `pkg-config` handle that with the requirement for `harfbuzz-icu`?

spl:
That is, either the build will fail if `pkg-config` is required or static linking would be used.

ratmice:
@spl the only instance i know of is issue #318

spl:
@ratmice Thanks, I hadn't seen that before.

I'm thinking it's hard to know what's going on there without more information. And I don't think it's worth trying to support such a strange situation.

Here's a thought. Since this issue would probably only really come up with someone doing a `cargo install tectonic` instead of installing a package through `apt`/`brew`/other, why not make static linking the dependent libraries the default for the crate? That way, the crate always builds regardless of what's on the user's system, and a different configuration can be used for building packages for Debian/Homebrew/other, in which `pkg-config` or something else could be used.

As an aside, the work I plan on doing with the `-sys` crates will support this approach.

crlf0710:
I think statically building and linking everything would be mostly fine. The statically building and deploying of `fontconfig` might be a little tricky, though.

ratmice:
@spl That would seem fine to me :+1:

ratmice:
I wanted to give this one last shot because it seemed like it was fairly close to working on the last go-around

pkgw:
@ratmice Looking at the shell-quoting around where that error happens, I think you need to split the flag on whitespace? Not obvious to me why that would be an issue on the musl build but not the others, though.

ratmice:
@pkgw I believe you are referring to the "-lharfbuzz -lm ...." which are all being sent as a single string,
perhaps splitting that might work.  Honestly though because this is a -o ... -c "tectonic/bibtex.c"
-l/-L options shouldn't even be there until we link libtectonic_c.a/libtectonic_cpp.a.

I will give splitting them a shot though, given the absence of a cc::Build::link_flags or something of the sort.

ratmice:
@pkgw let me know if you want me to try and update the conflicts with master and be more proactive about keeping this up to date.

for what it is worth; I had tried using this branch in a downstream project's CI on azure pipelines,
That wasn't fully successful on every platform (iirc there was an issue on mac os?), but I haven't managed to get back to it yet.

pkgw:
At long last, I think that #724 has fully dealt with this, building on a lot of the groundwork laid here.

Issue #393, opened 2019-06-17
Fails to inlcude pdf file as graphic


timokau:
When trying to include the file `uni-logo.pdf` in the `figures` directory of [this template](https://cs.uni-paderborn.de/fileadmin/informatik/fg/cn/Stuff/thesis.zip) (not sure if I have permission to re-host only the pdf, so I'm linking the zip), tectonic fails:

```
[timo@desk:~/data/documents/uni/ba/paper/thesis]$ tectonic -p minimal.tex 
note: this is a BETA release; ask questions and report bugs at https://tectonic.newton.cx/
Running TeX ...
(minimal.tex
LaTeX2e <2018-04-01> patch level 4
Babel <3.20> and hyphenation patterns for 84 language(s) loaded.
(article.cls
Document Class: article 2014/09/29 v1.4h Standard LaTeX document class
(size10.clo)) (graphics.sty (trig.sty) (graphics.cfg) (xetex.def)) (minimal.aux
)
warning: No trailer.  Are you sure this is a PDF file?
warning: No trailer.  Are you sure this is a PDF file?
warning: buffer:
stream36 /L 113 /Filter /FlateDecode /Length 40 0 R >> eb5b04cb37c65124>]
H�b``g``����������@,
�

 b
  �@��38��E��ľ

              Lo�
                 �( ah$���~��
0�������H3�@�<-

warning: buffer:
stream36 /L 113 /Filter /FlateDecode /Length 40 0 R >> eb5b04cb37c65124>]
Hb``g``@,


b
 @38Eľ

      Lo
        ( ah$~
0H3@<-

warning: Error while parsing PDF file.
warning: Error while parsing PDF file.

! Unable to load picture or PDF file 'figures/uni-logo.pdf'.
<to be read again> 
                   }
l.4 \includegraphics{figures/uni-logo}
                                      
No pages of output.
Transcript written on minimal.log.
error: the TeX engine had an unrecoverable error
caused by: halted on potentially-recoverable error as specified
debugging: backtrace follows:
stack backtrace:
   0:     0x5599787d3c5d - <no info>
   1:     0x5599787d3082 - <no info>
   2:     0x5599787d2818 - <no info>
   3:     0x5599787d2b95 - <no info>
   4:     0x5599785f3f0c - <no info>
   5:     0x5599785dded5 - <no info>
   6:     0x5599785dab1c - <no info>
   7:     0x5599785dc076 - <no info>
[...]

PR #396, opened 2019-06-19
src/io/mod.rs: fix file offsets reported just after an ungetc()

pkgw:
Our seek function was reporting the wrong file offsets when an ungetc() had just been performed. This was exposed in an unusual PDF file as described in issue #393.


pkgw:
I put together a gross PDF file that triggers the bug that doesn't have any copyright concerns as mentioned in #393, so we now have a test case — this should be ready to go.

Issue #397, opened 2019-06-20
Testing Tectonic on the arXiv paper dataset


Mrmaxmeier:
I'm experimenting with a few low-level changes in the engine. Tectonic's test suite is a nice early warning but passing the tests says little about being able to build real LaTeX documents in my experience.
The arXiv archive provides [TeX sources](https://arxiv.org/help/bulk_data_s3) for most of the published papers.
I've hacked up a proof of concept of a [crater](https://github.com/rust-lang-nursery/crater)-like regression testing tool using parts of the arXiv dataset.

Here's a quick rundown:
It'll run the engine on each source file and store logs and output files.
This takes about 30 minutes for 2500 papers.
~~It currently relies on [libfaketime](https://github.com/wolfcw/libfaketime) to get as close to reproducible builds as possible.~~ Tectonic's output is now reproducible when setting `SOURCE_DATE_EPOCH`. This means we're able to detect regressions by checking the output hashes. Visual comparisons of output PDFs might still be interesting though in the future.
There are some issues with deterministic outputs as some samples trigger undefined behavior.
I've borrowed some of crater's CSS and whipped up a [simple front-end](https://tt.ente.ninja) for the test results.

![Screenshot_2019-06-20](https://user-images.githubusercontent.com/3913977/59876926-0be5b080-93a5-11e9-83f8-c23a16aeb603.png)

I'll open separate issues for the false-positives caused by undefined behavior and will keep this one open as a kind-of tracking issue on reproducibility and regression testing.

Ideas / TODOs:
- [ ] valgrind / asan / msan
- [ ] visual diff for PDFs
- [x] remove libfaketime hacks
- [ ] track performance
- [ ] run all samples in-process (might uncover some state corruption)
- [x] automated runs for PRs (https://github.com/Mrmaxmeier/tectonic-on-arXiv/issues/5)

pkgw:
I've wanted to do this for so long, this is amazing!!!

pkgw:
Let me know if you're ever in the Boston area and I'll buy you the beverage(s) of your choice.

XVilka:
Would be interesting to test also the Oxidized fork/branch:
- https://github.com/tectonic-typesetting/tectonic/issues/459
- https://github.com/crlf0710/tectonic/tree/oxidize

Mrmaxmeier:
> Would be interesting to test also the Oxidized fork/branch:
> 
> * https://github.com/crlf0710/tectonic/tree/oxidize

[...]

PR #398, opened 2019-06-20
Fix an off-by-one in usages of the `name_of_file` variable.

Mrmaxmeier:
I missed a bit of pascal-like string handling in my refactoring PR: #165

---------------------------------------------

This is an example use-case for the regression tester (#397).

This comparison reveals the regression:
https://tt.ente.ninja/#/compare/v0.1.9-dirty/v0.1.10
This is the changeset for this PR:
https://tt.ente.ninja/#/compare/v0.1.11/fix-font-name-v0.1.11-276-g62ce204
This confirms that the regression was fixed:
https://tt.ente.ninja/#/compare/v0.1.9-dirty/fix-font-name-v0.1.11-276-g62ce204

_The differences in `CAG_D-17-00218_Edited.pdf` are not related._

### DIFF
--- a/tectonic/bibtex.c
+++ b/tectonic/bibtex.c
@@ -647,7 +647,7 @@ sam_wrong_file_name_print(void)
 {
     ttstub_puts (standard_output, "I couldn't open file name `");
 
-    name_ptr = 1;
+    name_ptr = 0;
     while (name_ptr <= name_length)
         ttstub_output_putc (standard_output, name_of_file[name_ptr++]);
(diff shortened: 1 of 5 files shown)


pkgw:
CI failures were spurious, so this looks good. It would be nice to add a bit more of an explicit test to the regression suite (although I do see that the output in one of the tests is improved), but ... well, I continue to not be as fanatical about building up the test suite and coverage stats than maybe I should be.

Issue #185, opened 2018-06-29
Distribute a static build


malbarbo:
I know what you're thinking: it's easier said than done!

But I did it! I built a static version of tectonic! It toke a lot of time, but the final patches are really small.

I used the following docker alpine container to do the build:

```Dockerfile
FROM alpine:edge

RUN apk update && \
    apk upgrade && \
    apk add \
        g++ \
        git \
        rust \
        cargo \
        fontconfig-dev \
        freetype-static \
        glib-static \
        graphite2-dev \
        graphite2-static \
        harfbuzz-dev \
        harfbuzz-static \
        icu-dev \
        icu-static \
        openssl-dev \
        zlib-dev

ENV TECTONIC_STATIC=1
ENV PKG_CONFIG_ALL_STATIC=1
ENV OPENSSL_STATIC=1
ENV OPENSSL_DIR=/usr
ENV RUSTFLAGS="-C target-feature=+crt-static"

RUN mkdir /out/

CMD git clone https://github.com/malbarbo/tectonic --branch static && \
    cd tectonic && \
    sed -i -e 's/AES_cbc_encrypt/AES_cbc_encrypt2/g' \
              ./tectonic/dpx-dpxcrypt.h \
              ./tectonic/dpx-pdfencrypt.c \
              ./tectonic/dpx-dpxcrypt.c && \
    cargo build --release && \
    cp target/release/tectonic /out/
```

```
docker build . -t tectonic
docker run -v $(pwd):/out tectonic
```

The image install the requirements including [static](https://github.com/alpinelinux/aports/pulls?q=is%3Apr+author%3Amalbarbo) version of the libs.

The env vars indicates to `tectonic`, `pkg-config` and `openssl` packages and `rustc` to do a static build.

The rename of `AES_cbc_encrypt` to `AES_cbc_encrypt2` is to avoid duplicate symbol between tectonic and openssl.

The following [changes](
https://github.com/tectonic-typesetting/tectonic/compare/master...malbarbo:static) to tectonic are needed:

- disable `serde_derive` (it [does not work](https://github.com/rust-lang/cargo/issues/5266) with static musl build). The changes disable loading config, so the final patch needs another approach, maybe implementing `Deserialize`
- add `TECTONIC_STATIC` as an option to `build.rs` to static link `stdc++`
[...]

PR #425, opened 2019-08-07
document using docker for cross compilation

efx:
* ran this on Docker for Mac 2.1.0 to produce a static binary that I
executed successfully on an Amazon Linux Instance
* Partially addresses #185. I think we would want to decide on whether
to automate this and when per https://github.com/tectonic-typesetting/tectonic/blob/master/dist/travis.sh#L272


pkgw:
Thanks! Circle CI failed, but it just does that once in a while, seemingly due to weaknesses in the QEMU approach to the PowerPC build.

PR #427, opened 2019-08-10
create an mdbook for documentation

### LINKED ISSUE #62: First pass at documentation
The Tectonic documentation tool, ironically, has no documentation itself. That needs to be fixed.

One day the docs will be written in TeX, but for the proximate future I think it is more practical to write them in Markdown so that they can appear on the website — with Unicode and decent CSS, the presentation will be sufficiently readable. Translating from MD to TeX will require some effort in the future, but it will just be a bit of busywork, not anything fundamentally challenging.

- [ ] the config file (cc #59)
- [ ] the command line options
- [ ] the idea behind the "bundle" approach
- [ ] what to do if a needed file is missing or too old
- [ ] (lack of) biblatex/biber support
- [ ] location and structure of cache directory (cc #45)
- [ ] "why not just use latemk?" etc
- [ ] emphasize that Tectonic is its own engine, not something that runs other engines
- [ ] how to emulate `latexmk -pvc` (cf #82 and other issues)

efx:
Start to resolve #62 and other documentation related issues.

I found this  [this rust subreddit discussion](https://www.reddit.com/r/rust/comments/7eohmt/whats_the_best_practice_for_documenting_a_rust/) about best practices for documentation. [Doxidize](https://steveklabnik.github.io/doxidize/index.html) seems like it could suit us well but [isn't ready for use](https://github.com/steveklabnik/doxidize/issues/116#issuecomment-421607742).
mdBook may not integrate as nicely with a crate but it seems like the community standard.
I started a preliminary structure but am sure we should establish the audience and then establish a suitable TOC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 /tests/plain.fmt
 /tests/00*00-latex-*.fmt
 /tests/00*00-plain-*.fmt
+
+# ignore mdbook compiled output (HTML)
+docs/book/*
--- a/README.md
+++ b/README.md
@@ -70,3 +70,7 @@ which in turn uses Rust’s `proc_macro` feature. The `proc_macro` functionality
 [is not available on musl targets](https://github.com/rust-lang/rust/issues/40174),
 and so must be turned off if you wish to build a completely static Tectonic
 executable.
+
+## Documentation System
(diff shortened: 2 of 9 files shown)


pkgw:
Thank you for the submission, and sorry for the delay in my getting to this! Let's get things rolling!

efx:
No need to apologize @pkgw! 🚀

PR #428, opened 2019-08-13
Use libz-sys backend for flate2 and disable bzip2 for zip

malbarbo:
There are three zip implementations being used in tectonic:
- from zip crate (libflate implemented in rust)
- from flate2 crate (miniz-sys implemented in c)
- zlib (native dependence in build.rs)

This PR make flate2 use zlib, removing the miniz-sys dependence. We also disable bzip2 feature from zip crate, I don't think it is being used. This reduces 93k from the final binary.

When (or if) https://github.com/mvdnes/zip-rs/pull/107 get merged, we can get rid of libflate dependence.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -127,24 +127,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bzip2"
-version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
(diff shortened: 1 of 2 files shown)


malbarbo:
https://github.com/mvdnes/zip-rs/pull/107 was merged and a new version of zip-rs was released! I updated Cargo.lock so now we have only one zip implementation.

pkgw:
Great, thank you!

PR #429, opened 2019-08-13
Encapsulate app_dirs usage

malbarbo:
This makes it easier to customize the app dirs.


pkgw:
Thanks! (For posterity: CI failure is one of the transient Circle failures, no reason to think it's related to this PR.)

PR #432, opened 2019-08-16
Quickfix for #429 (make sure the cache dirs are created)

rekka:
We need to test `local_cache.rs` and `itarbundle.rs`. 0% coverage currently.


pkgw:
Thanks for this — point taken about the test coverage.

PR #434, opened 2019-08-16
remove outdated documentation comment

efx:
I could not find this example so removed the line.
If the `cli_driver.rs` moved somewhere else I can instead update the text to that location.


pkgw:
Good catch — the comment should now refer to `src/bin/tectonic.rs`.

PR #439, opened 2019-08-20
Make aho-corasick crate optional

malbarbo:


malbarbo:
I'm working on embedding Tectonic and this can save some bytes and compiling time.

Maybe we can remove aho-corasick altogether. What do you think @pkgw?

pkgw:
Thanks @malbarbo! I think I'd like to avoid introducing Cargo features, and I don't have any reason to believe that aho-corasick makes a real difference in the UX, so yeah, let's just remove it altogether. Can you please just include a comment in the code mentioning that aho-corasick used to be used, but that we removed it to reduce the code size?

malbarbo:
Done.

pkgw:
@malbarbo Beautiful, thanks!

pkgw:
@malbarbo Clippy [has a suggestion](https://travis-ci.org/tectonic-typesetting/tectonic/jobs/575422225#L737).

malbarbo:
Ok, let's now.

PR #440, opened 2019-08-20
Disable exceptions and rtti for C++ code

malbarbo:
There is no throw clause in the source code and Converter::Exception is never created, so I think this is dead code.


pkgw:
Thanks, looks good!

Issue #433, opened 2019-08-16
Need tests for the tar bundle and local cache code


rekka:
Currently `local_cache.rs` and `itarbundle.rs` has no test coverage. We should devise tests that check downloading files from http tar bundle and the local caching that would catch things like #432 . Not sure what the best approach would be.

rekka:
I thought a bit about how to approach this. One option is for the test binary to launch a basic static file server on `localhost` serving a tiny tar bundle and have tectonic use it. Looking around for static file servers in rust, there's code like <https://github.com/weihanglo/sfz> (MIT licensed) that we could slim down to the bare minimum. But I do not have experience with testing network code. Is it OK to run such a server on a CI machine or Windows?

pkgw:
@rekka Yes, that should definitely be OK, and I agree that it would be the right way to test this portion of the functionality.

Besides the example you found, I bet that there are a *lot* of sample static-file servers out there; it's a very common demo case. We could also look into whatever [zola](https://www.getzola.org/) uses for its `zola serve` functionality. And then there's the [Big list of http static server one-liners](https://gist.github.com/willurd/5720255) if you open up to considering other languages/frameworks ...

rekka:
@pkgw You are right. I looked into it over the weekend and I could write a fully functional server for tectonic testing purposes using `hyper` that supports RANGE http requests in only about 50 lines of code. We can run it directly in the test binary on a separate thread while executing a tectonic session.

I used `zip2tarindex` from [tectonic-staging](https://github.com/tectonic-typesetting/tectonic-staging/) to create a dummy tar bundle and it seemed to basically work.

I am considering whether it makes sense to use the code `zip2tarindex` directly (as a library) in the testing binary to create a tar bundle in memory on the fly from `tests/assests` directory.

We also need to modify the local cache root to point it to a temporary directory so that we don't pollute the real cache. I have tried to do this using a hacky global variable (env variable) similarly to what tectonic already does for other tests, which again basically works.

I am gonna prepare a pull request with a prototype within a week or so (I hope...).

pkgw:
Cool, I'm glad to hear that implementing a test server in Rust looks to be straightforward!

If we really want to exercise all of the client code, we probably do need to reuse the zip2tarindex code. So we should probably look into turning that into a published crate on crates.io.

[...]

PR #441, opened 2019-08-22
Test local cache and tar bundle

rekka:
This is a working prototype of a test for tar bundle + local cache to address #433.

* Uses a simple hyper server to serve a static "tarindex" created from files in `tests/assets`.

* Implemented test: Run tectonic twice and collect its network requests. Check that the index and the digest file are requested only once.

* Local cache is set to a temporary directory so as not to pollute user's cache. Its location is changed by setting an envvar `TECTONIC_TESTING_CACHE_ROOT`. 

* Tectonic actually does not use the tar format. To keep things simple, we do not need to use `zip2tarindex` but instead just serve the content of the requested file.


rekka:
I think that the tests work. 

There are a lot of small commits so I will squash everything once approved to merge (to keep the history clean).

Using the env. variable to change the cache root is rather hacky at the moment. 

There are now two tests:

- `test_full_session`: runs the full tectonic engine. This is a complete test but rather slow. 
- `test_cached_url_provider`: tests only the bundle given by `config.make_cached url provider`. Fast. I believe that this test is sufficient.

I am considering adding a test of when the bundle digest changes (bundle update). Currently the engine should stop and update the digest in the cache.

rekka:
I think that the tests work. 

There are a lot of small commits so I will squash everything once approved to merge (to keep the history clean).

There are now two tests:

- `test_full_session`: runs the full tectonic engine. This is a complete test but rather slow. 
- `test_cached_url_provider`: tests only the bundle given by `config.make_cached url provider`. Fast. I believe that this test is sufficient.

rekka:
@pkgw This pull request is ready for a review. Please let me know if you have any suggestions. I think it is a good idea to have this merged before #431

PR #420, opened 2019-08-01
fix checkouts of standing data for tests on Windows, enable vcpkg on macOS and Linux

mcgoo:
@crlf0710 added the ability to build using msvc with libraries from vcpkg. 

I'm not sure if it is useful to you, but vcpkg-rs 0.2.7 adds support for Linux and macOS so I enabled it for them also. (This does allow using the exact same versions of libraries across all platforms which may be of interest.) Windows (msvc) works for static and dynamic builds, and Linux and macOS work for static builds. Almost everything builds and passes tests, but on Windows, one of the doctests still fails the first time it is run.

This fixes the issue with the line endings in the test data getting changed to CRLF when it is checked out on Windows, by adding a `.gitattributes` file. This does not update an existing checkout so it is probably necessary to delete the `tests/` directory and check it out again to get the unchanged versions. (This is particularly painful if you clone the repo and `.gitattributes` is not on the master branch yet - git will not change the line endings as you change branches.)

For a Windows static build I used:
```
.\vcpkg install --triplet x64-windows-static fontconfig freetype harfbuzz[icu,graphite2]
set VCPKG_ROOT=<wherever>
set TECTONIC_VCPKG=1
set RUSTFLAGS=-Ctarget-feature=+crt-static
cargo test
```

For a Linux static build I used:
```
./vcpkg install fontconfig freetype harfbuzz\[icu,graphite2\]
set VCPKG_ROOT=<wherever>
set TECTONIC_VCPKG=1
cargo test
```

For a macOS static build I used:
```
./vcpkg install freetype harfbuzz\[icu,graphite2\]
set VCPKG_ROOT=<wherever>
set TECTONIC_VCPKG=1
cargo test
```

For a Windows dynamic build I used:
```
.\vcpkg install --triplet x64-windows fontconfig freetype harfbuzz[icu,graphite2]
set VCPKG_ROOT=<wherever>
set TECTONIC_VCPKG=1
set VCPKGRS_DYNAMIC=1
cargo test
```

PR #451, opened 2019-08-27
document building with vcpkg

efx:
Document how to use vcpkg to build tectonic on macOS.
Adapted the details from https://github.com/tectonic-typesetting/tectonic/pull/420.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -6,3 +6,5 @@
 - [Getting Started](getting-started/index.md)
 - [Tectonic’s Command Line Interface](cli/index.md)
 - [The Tectonic Implementation of the TeX Language](lang-impl/index.md)
+- [Cookbook](cookbook/index.md)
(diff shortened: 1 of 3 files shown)


pkgw:
Thanks for researching linters while you're at it!

FWIW I think I'd ideally like to have this text in a section of the book named Installation or Development (or maybe Contributing?), but as I wrote before, I think it's best to get text in first and rearrange later. I guess my current thinking is that Getting Started should have a Quick Installation section, which should point to an Installation document with less-common cases.

efx:
Certainly!
I think `Installation` or `Development` would be helpful sections. I would want to migrate the other description of how to build a static binary using docker to that section as well.
I also think we could move the fuzzing guide there as well.

efx:
In the spirit of being a Rust project, [`textwrap`](https://lib.rs/crates/textwrap) in a simple CLI would be great.

Issue #452, opened 2019-08-27
Tectonic is slower than xelatex


malbarbo:
This can be observed when generating a pdf for `tests/xenia/paper.tex`.

Using `tectonic` from master compiled with `cargo install --path .` and `xelatex` from Debian 9 in a i3-2330M 2.20GHz machine, I got the following results

```
> time tectonic -C -c minimal -r 1 paper.tex
real    0m1,642s
user    0m1,530s
sys     0m0,112s

> time xelatex -interaction=batchmode paper.tex
real    0m0,842s
user    0m0,796s
sys     0m0,120s
```

Running `perf` show that at least 25% of the time is expended by sha2 crate:

```
# Overhead  Command   Shared Object             Symbol                                                                                                                                                               
# ........  ........  ........................  ...........................................................................................                                                                          
#                                                                                                                                                                                                                    
    23.78%  tectonic  tectonic                  [.] sha2::sha256_utils::compress256                                                                                                                                  
    10.66%  tectonic  tectonic                  [.] get_next                                                                                                                                                         
     6.03%  tectonic  tectonic                  [.] load_fmt_file                                                                                                                                                    
     3.61%  tectonic  tectonic                  [.] <std::io::buffered::BufReader<R> as std::io::Read>::read                                                                                                         
     3.55%  tectonic  tectonic                  [.] <tectonic::io::InputHandle as std::io::Read>::read                                                                                                               
     3.29%  tectonic  tectonic                  [.] sha2::sha256::Engine256::input                                                                                                                                   
     2.51%  tectonic  tectonic                  [.] macro_call                                                                                                                                                       
     2.17%  tectonic  tectonic                  [.] length                                                                                                                                                           
     1.76%  tectonic  [kernel.kallsyms]         [k] clear_page_rep                                                                                                                                                   
     1.71%  tectonic  tectonic                  [.] tectonic::io::InputHandle::getc                                                                                                                                  
     1.57%  tectonic  tectonic                  [.] flush_list                                                                                                                                                       
     1.50%  tectonic  tectonic                  [.] get_uni_c                                                                                                                                                        
     1.39%  tectonic  tectonic                  [.] tectonic::engines::input_getc                                                                                                                                    
     1.16%  tectonic  tectonic                  [.] id_lookup                                                                                                                                                        
     1.16%  tectonic  tectonic                  [.] ht_lookup_table
```

[...]

PR #453, opened 2019-08-28
Do not compute the sha256 digest for read-only files

rekka:
As observed in #452, a big chuck of runtime is spend computing sha256 digests. However, these are for the most part never used.

This pull request adds `InputHandle::read_only: bool` field and `fn
InputHandle::new_read_only` that sets this field to `true`. It indicates
that it is not possible to write into such file, for example, the format
file, the tar bundle or zip bundle, and so `InputHandle` does not need
to compute sha256 digest.

Saves 15--25% of tectonics runtime for a typical document.

On my machine, `tests/xenia/paper.tex` the run time is 1s vs 1.3s, and on a 30 pages tex file 1.9s vs. 2.2s.

### DIFF
--- a/src/io/cached_itarbundle.rs
+++ b/src/io/cached_itarbundle.rs
@@ -680,7 +680,7 @@ impl IoProvider for CachedITarBundle {
             Err(e) => return OpenResult::Err(e.into()),
         };
 
-        OpenResult::Ok(InputHandle::new(
+        OpenResult::Ok(InputHandle::new_read_only(
             name,
             BufReader::new(f),
             InputOrigin::Other,
--- a/src/io/format_cache.rs
+++ b/src/io/format_cache.rs
@@ -84,7 +84,7 @@ impl IoProvider for FormatCache {
             OpenResult::Err(e) => return OpenResult::Err(e),
         };
 
-        OpenResult::Ok(InputHandle::new(
+        OpenResult::Ok(InputHandle::new_read_only(
             name,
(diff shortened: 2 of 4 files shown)


rekka:
Thanks for having a look. 

As I understand the current logic, returning `None` for digest from `InputHandle` shouldn't be a problem. The logic that determines reruns allows this on this line: https://github.com/tectonic-typesetting/tectonic/blob/6ea22a0ed47967d63fc692ce6917114727ef9f7f/src/driver.rs#L597

Then when the engine is rerun, the previously written file will be read from the memory io layer, digest will be computed, and all should be good.

Also, `None` could be returned by `InputHandle` even before this commit when the engine has a weird access pattern, in which case the digest cannot be computed. 

In any case, it would be nice to have a test for this, which is probably easier once #441 is merged.

pkgw:
Oh, yes, that is all correct. I'm happy to merge unless you were hoping to rebase and/or add a test case?

rekka:
@pkgw Yeah, let me see if I can put together a test just to be sure that what I am saying above is correct.

rekka:
@pkgw Rebased. I tried to put together test but realized that it is not really worth it since we (I) are quite confident that the logic is correct... I think it's safe to merge.

pkgw:
Thanks! This and the bundle-index caching should really help the UX a lot ... I'd like to give these a bit of time to shake down, but we should definitely cut a new release soon.

PR #458, opened 2019-09-05
Warnings during the last pass only

rekka:
Currently tectonic prints "warnings were issued ..." warning only the first time the tex engine pass has warnings. However, this is not very useful since we expect warnings in the early passes due to missing references etc. What one cares about is whether the last pass has any warnings.

This commit updates the logic so that only the warnings from the very last pass are reported.

As a minor change, I introduced an enum `RerunReason` to avoid stringly typing it.

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -522,6 +522,12 @@ impl ProcessingSessionBuilder {
     }
 }
 
+#[derive(Debug, Clone)]
+enum RerunReason {
+    Bibtex,
+    FileChange(String),
+}
+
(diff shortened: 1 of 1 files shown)


pkgw:
Thanks, this seems like a nice improvement to me!

Have you looked into adding some tests to get the code coverage up a tiny bit? I can't bring myself to take a really hard line about this stuff, but it would be good to keep that number inching upwards if we can ...

rekka:
I will try to look into it. It's a good idea to have tests for the user messages as well.

Issue #405, opened 2019-07-05
driver (library) chatter referring to the tectonic binary command line options


ratmice:
I had been using the latex_to_pdf functionality.  On first run of the program when it downloads styles and sets up the tectonic cache it takes quite a long time, with no feedback because it's using the chatter level status::NoopStatusBackend::new, and has print_stdout to false.

As such I had started moving towards setting up the engine manually,
And used the chatter level of Normal, The output of such refers to arguments my program doesn't currently accept.  I.e. src/driver.rs refers to arguments accepted by
src/bin/tectonic.rs

It isn't hard to add these to my program, and I shall, but it is worrying if e.g. bin/tectonic adds some new option, driver starts referring to that, and in an update I don't notice so my program starts recommending options it may not recognize.

* [--keep-intermediates](https://github.com/tectonic-typesetting/tectonic/blob/017c5785e4b758125aee4995b40fd48f6f78233c/src/driver.rs#L713-L717)
* [--print/--keep-logs](https://github.com/tectonic-typesetting/tectonic/blob/017c5785e4b758125aee4995b40fd48f6f78233c/src/driver.rs#L990-L993)
  [ditto](https://github.com/tectonic-typesetting/tectonic/blob/017c5785e4b758125aee4995b40fd48f6f78233c/src/driver.rs#L1060)
  [ditto](https://github.com/tectonic-typesetting/tectonic/blob/017c5785e4b758125aee4995b40fd48f6f78233c/src/driver.rs#L1071)
  [finally](https://github.com/tectonic-typesetting/tectonic/blob/017c5785e4b758125aee4995b40fd48f6f78233c/src/driver.rs#L1102-L1109)

I'm not exactly sure I have any good ideas for a good solution, three came to mind
1. Having a specific chatter level for the tectonic binary.
2. A fancy enum callback/formatting scheme.
3. A mechanism my program can use to direct tectonic options directly to the driver.

1 I'm not sure is plausible, as we see some of the chatter are warnings rather than just notes.
2 Seems overkill, especially when we see various similar messages worded slightly differently based upon context, It looks like the most painful option.
3 seems like it could work and be future compatible and least hassle.

I am reminded of the MLWorks compiler driver which worked like 3.
The compiler driver would essentially ignore options hat weren't wrapped in a begin/end marker such as:
`foo --begin-tectonic --print --keep-intermediates --end-tectonic`
[...]

PR #465, opened 2019-09-12
use structopt for CLI argument parsing

efx:
Per conversations [here](https://github.com/tectonic-typesetting/tectonic/issues/405#issuecomment-509207477) and in other issues.

I manually reviewed each of the flags but acceptance testing would be helpful.
- I decided to use the `FromStr` trait for parsing certain switches that end up as an enum. You can use enums [directly](https://github.com/TeXitoi/structopt/blob/master/examples/enum_in_args.rs) in structopt, but you need to wrap them in a clap macro. It changed the existing design too much and I hit a wall with some compiler errors so I took this approach.
- structopt seems to lowercase the `name` argument configuration. E.g. ` <input>    The file to process, or "-" to process the standard input stream"`

### DIFF
--- a/src/bin/tectonic.rs
+++ b/src/bin/tectonic.rs
@@ -2,12 +2,14 @@
 // Copyright 2016-2018 the Tectonic Project
 // Licensed under the MIT License.
 
-use clap::crate_version;
 use tectonic;
 
+use structopt::StructOpt;
+
+use clap::crate_version;
 use clap::{App, Arg, ArgMatches};
 use std::fs::File;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process;
 
 use tectonic::config::PersistentConfig;
@@ -19,6 +21,26 @@ use tectonic::status::{ChatterLevel, StatusBackend};
 
 use tectonic::{ctry, errmsg, tt_error, tt_error_styled, tt_note};
 
+#[derive(Debug, StructOpt)]
+#[structopt(name = "Tectonic", about = "Process a (La)TeX document")]
+struct Opt {
+    /// The name of the "format" file used to initialize the TeX engine
+    #[structopt(long, short, name = "PATH", default_value = "latex")]
+    format: String,
+    /// Use this Zip-format bundle file to find resource files instead of the default
+    #[structopt(
+        takes_value(true),
+        parse(from_os_str),
+        long,
+        short,
+        name = "zip_file_path"
+    )]
+    bundle: PathBuf,
+    /// Use this URL find resource files instead of the default
+    #[structopt(takes_value(true), long, short, name = "url")]
+    web_bundle: String,
+    // TODO add URL validation
+}
 fn inner(
     args: ArgMatches,
     config: PersistentConfig,
@@ -154,6 +176,8 @@ fn inner(
 }
 
 fn main() {
+    let opt = Opt::from_args();
+    println!("{:?}", opt);
     let matches = App::new("Tectonic")
         .version(crate_version!())
         .about("Process a (La)TeX document")
--- a/src/bin/tectonic.rs
+++ b/src/bin/tectonic.rs
@@ -6,7 +6,6 @@ use tectonic;
 
 use structopt::StructOpt;
 
-use clap::ArgMatches;
 use std::fs::File;
 use std::path::{Path, PathBuf};
 use std::process;
@@ -80,20 +79,20 @@ struct Opt {
     outdir: Option<PathBuf>,
 }
 fn inner(
-    args: ArgMatches,
+    args: Opt,
     config: PersistentConfig,
     status: &mut TermcolorStatusBackend,
 ) -> Result<()> {
     let mut sess_builder = ProcessingSessionBuilder::default();
-    let format_path = args.value_of("format").unwrap();
+    let format_path = args.format;
     sess_builder
-        .format_name(format_path)
-        .keep_logs(args.is_present("keep_logs"))
-        .keep_intermediates(args.is_present("keep_intermediates"))
+        .format_name(&format_path)
+        .keep_logs(args.keep_logs)
+        .keep_intermediates(args.keep_intermediates)
         .format_cache_path(config.format_cache_path()?)
-        .synctex(args.is_present("synctex"));
+        .synctex(args.synctex);
 
-    let output_format = match args.value_of("outfmt").unwrap() {
+    let output_format = match &args.outfmt {
         "aux" => OutputFormat::Aux,
         "html" => OutputFormat::Html,
         "xdv" => OutputFormat::Xdv,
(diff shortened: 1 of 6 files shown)


pkgw:
Thanks very much for undertaking this! I am not sure why CodeCov thinks that 0% of the diff is being hit ... Anyway, I aspire to look at this soon, but (as per the usual refrain) my bandwidth is very limited these days. Other folks are more than welcome to review, and feel free to ping me if I'm slow to process this PR!

src/bin/tectonic.rs
```diff
@@ -177,95 +206,16 @@ fn inner(
...
-             .help("How much chatter to print when running")
-             .possible_values(&["default", "minimal"])
-             .default_value("default"))
-        .arg(Arg::with_name("outdir")
-             .long("outdir")
-             .short("o")
-             .value_name("OUTDIR")
-             .help("The directory in which to place output files [default: the directory containing INPUT]"))
-        .arg(Arg::with_name("INPUT")
-             .help("The file to process, or \"-\" to process the standard input stream")
```
ratmice: It wasn't immediately obvious to me where the help strings had gone,
but now I see that they go into the doc comment.

efx: Agreed; this is one of those implicit things about `structopt`.

ratmice:
I'm still fairly new to rust, so not really a reviewer quality developer, but overall beyond a couple of interim things in the series, which were fixed up by the end it looked good to me.

pkgw:
@ratmice Thank you for reviewing! In some ways, being new to Rust is an asset when it comes to this thing ...

efx:
@pkgw no rush at all, thank for you the update. I was confused by the coverage statement too. I can try adding a test sometime.
@ratmice indeed, thank you for reviewing!
Beginners have fresh eyes. I'm in the same camp.

src/bin/tectonic.rs
```diff
@@ -19,49 +20,91 @@ use tectonic::status::{ChatterLevel, StatusBackend};
...
-    status: &mut TermcolorStatusBackend,
-) -> Result<()> {
+#[derive(Debug, StructOpt)]
+#[structopt(name = "Tectonic", about = "Process a (La)TeX document")]
+struct Opt {
+    /// The file to process, or "-" to process the standard input stream"
+    #[structopt(name = "INPUT")]
+    input: String,
+    /// The name of the "format" file used to initialize the TeX engine
+    #[structopt(long, short, name = "PATH", default_value = "latex")]
```
pkgw: If structopt is going to lowercase these, I think we should just go along with it — it is hard for me to see how changing this would cause problems for anyone.

src/bin/tectonic.rs
```diff
@@ -19,49 +20,91 @@ use tectonic::status::{ChatterLevel, StatusBackend};
...
 use tectonic::{ctry, errmsg, tt_error, tt_error_styled, tt_note};
 
-fn inner(
-    args: ArgMatches,
-    config: PersistentConfig,
-    status: &mut TermcolorStatusBackend,
-) -> Result<()> {
+#[derive(Debug, StructOpt)]
+#[structopt(name = "Tectonic", about = "Process a (La)TeX document")]
+struct Opt {
```
pkgw: It looks like this is the structure named recommended by structopt, but I'm not a big fan. How about `CliOptions`?

src/bin/tectonic.rs
```diff
@@ -19,49 +20,91 @@ use tectonic::status::{ChatterLevel, StatusBackend};
...
+    #[structopt(long, name = "hide_path")]
+    hide: Option<Vec<PathBuf>>,
+    /// Print the engine's chatter during processing
+    #[structopt(long = "print", short)]
+    print_stdout: bool,
+    /// The directory in which to place output files [default: the directory containing INPUT]
+    #[structopt(name = "OUTDIR", short, long, parse(from_os_str))]
+    outdir: Option<PathBuf>,
+}
+fn inner(args: Opt, config: PersistentConfig, status: &mut TermcolorStatusBackend) -> Result<()> {
```
pkgw: Random fun idea, for certain values of "fun": can we implement `TryFrom<CliOptions>` for `ProcessingSession`?

src/driver.rs
```diff
@@ -210,6 +212,23 @@ pub enum OutputFormat {
...
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> StdResult<Self, Self::Err> {
+        let actual = match s {
+            "aux" => OutputFormat::Aux,
+            "html" => OutputFormat::Html,
+            "xdv" => OutputFormat::Xdv,
+            "pdf" => OutputFormat::Pdf,
+            "fmt" => OutputFormat::Format,
+            _ => unreachable!(),
```
pkgw: I would have expected these converters to return an `Err` with an unexpected string, rather than panic. Is there a specific reason for the use of `unreachable!`? If so, explanatory comments would be helpful.

src/bin/tectonic.rs
efx: Agreed! I had started doing this and will push up those changes.

src/bin/tectonic.rs
efx: Yes, quite the better name!

src/driver.rs
efx: > Is there a specific reason for the use of unreachable!? If so, explanatory comments would be helpful.

I wondered about this as well. I actually tried to maintain the existing [structure](https://github.com/tectonic-typesetting/tectonic/blob/master/src/bin/tectonic.rs#L42) but given the semantics has changed I should update this error to be more meaningful.

pkgw: Yeah ... in that targeted usage, we know that clap/structopt will prevent the code from ever being sent an unacceptable value. For a broader `FromString` implementation, though, we can't make that assumption.

src/bin/tectonic.rs
efx: @pkgw I have a vague idea of how to do this so perhaps I could try in another PR, or if you get to it I can review. I am stuck on how to share `CliOptions` from `bin/tectonic.rs` to the `driver.rs` file. ([This](https://github.com/efx/tectonic/compare/topic-structopt...efx:topic-structopt-tryfrom?expand=1) is as far as I got)

pkgw: Definitely no worries ­I don't even know if my idea even works, and it wouldn't actually provide any additional functionality. (I *think* you can impl the trait inside `bin/tectonic.rs` since the type parameter would be a type local to the binary crate, but I didn't check whether that's possible.) So, yeah, this should be punted.

src/status/mod.rs
```diff
@@ -18,6 +20,20 @@ pub enum ChatterLevel {
...
 }
 
+impl FromStr for ChatterLevel {
+    type Err = &'static str;
+
+    fn from_str(a_str: &str) -> StdResult<Self, Self::Err> {
+        let actual = match a_str {
+            "default" => ChatterLevel::Normal,
+            "minimal" => ChatterLevel::Minimal,
+            _ => unreachable!(),
```
pkgw: This `FromStr` impl still needs changing away from `unreachable!`.

efx: Ah, thanks for catching that.

PR #473, opened 2019-09-29
omit-frame-pointer and profile feature

ratmice:
This PR is waiting on a release of cc-rs, containing the cc::Build::force_frame_pointer func,
and will at least need updating of Cargo.toml at that time.

Until then:
This implements a profile feature which emits debug info, and enables/disables frame-pointers dependent upon whether the target architecture is known to be capable of profiling with frame-pointers omitted.

cc'ing @rekka for comments since he has a profile branch.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -202,6 +202,22 @@ fn main() {
         ccfg.flag_if_supported(flag);
     }
 
+    const PROFILE_BUILD_PLATFORM_REQUIRES_FRAME_POINTER: bool = cfg!(not(any(
+        // Whitelist of platforms which do not require frame pointers.
+        all(target_os = "linux", target_arch = "x86_64"),
+        // Add more platforms here.
+    )));
+    const PROFILE_BUILD_ENABLED: bool = cfg!(feature = "profile");
+
+    fn profile_config(cfg: &mut cc::Build) {
+        if PROFILE_BUILD_ENABLED {
+            cfg.debug(true)
+                .force_frame_pointer(PROFILE_BUILD_PLATFORM_REQUIRES_FRAME_POINTER);
+        }
+    }
+
+    profile_config(&mut ccfg);
+
     ccfg.file("tectonic/bibtex.c")
         .file("tectonic/core-bridge.c")
         .file("tectonic/core-memory.c")
@@ -332,6 +348,8 @@ fn main() {
     for flag in &cppflags {
         cppcfg.flag_if_supported(flag);
     }
+
+    profile_config(&mut cppcfg);
 
     cppcfg
         .cpp(true)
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -202,6 +202,20 @@ fn main() {
         ccfg.flag_if_supported(flag);
     }
 
+    fn profile_config(cfg: &mut cc::Build) {
+        if cfg!(all(
+            feature = "profile",
+            target_os = "linux",
+            target_arch = "x86_64"
+        )) {
```
pkgw: Here I might suggest using the `cfg!` stanza to declare a `const bool`, which allows a bit more clarity about what the test is getting at:

```rust
const PLATFORM_CAN_PROFILE_WITHOUT_FRAME_POINTER: bool = cfg!(...);

if PLATFORM_CAN_PROFILE_WITHOUT_FRAME_POINTER { ... }
```

Or, you know, just a code comment to that effect.

ratmice: Ok, I added an any() around that all and a comment too, so it's obvious how to add more platforms.

build.rs
```diff
@@ -202,6 +202,24 @@ fn main() {
...
+        ), // Add more platforms here.
+    ));
+
+    fn profile_config(cfg: &mut cc::Build) {
+        if PLATFORM_CAN_PROFILE_WITHOUT_FRAME_POINTER {
+            cfg.debug(true).force_frame_pointer(false);
+        } else if cfg!(feature = "profile") {
+            cfg.debug(true).force_frame_pointer(true);
+        }
+    }
```
ratmice: Sorry for the churn, it occurred to me it might be cleaner
to flip the logic around like so:

```rust
    const PROFILE_BUILD_PLATFORM_REQUIRES_FRAME_POINTER: bool = cfg!(not(any(
        // Whitelist of platforms which do not require frame pointers.
        all(target_os = "linux", target_arch = "x86_64"),
        // Add more platforms here.
    )));
    const PROFILE_BUILD_ENABLED: bool = cfg!(feature = "profile");

    fn profile_config(cfg: &mut cc::Build) {
        if PROFILE_BUILD_ENABLED {
            cfg.debug(true)
                .force_frame_pointer(PROFILE_BUILD_PLATFORM_REQUIRES_FRAME_POINTER);
        }
    }

```

pkgw: No worries, I like it!

pkgw:
@ratmice FYI, I'd rather wait until a new release of `cc` comes out before merging this — sound OK?

build.rs
ratmice: Hmm, I can't decide whether that means to push that change or not :confused:

pkgw: Sorry for being unclear — I like the proposed change.

ratmice:
closing in lieu of #511

Issue #476, opened 2019-10-01
build.rs and cross-compilation


ratmice:
While (trying to) test #473 I ran across the issue that when cross compiling,
build.rs `#[cfg(target_os = "...", target_arch = "...")]` is compiling with `target=host`, 
because build.rs runs on the host, these should probably be migrated to using.

`CARGO_CFG_TARGET_OS`, and `CARGO_CFG_TARGET_ARCH`,

##### pkg_config:
I'm not sure what if anything, `pkg-config` requires from build.rs to work properly.
I didn't see anything that would allow us to send it the target so I imagine it punts to the user.

pkgw:
I am a bit surprised that the `cfg` bits are misleading in this way, but that indeed seems to be how things work. The [target_build_utils](https://docs.rs/target_build_utils/0.3.1/target_build_utils/) crate might be helpful?

For pkg-config, I think the necessary setup has to happen outside of build.rs. For the MUSL cross-compile, and in other cases, there's a `PKG_CONFIG_ALLOW_CROSS` environment variable that we have to set .. some grepping for that might yield helpful examples from our CI setup.

ratmice:
I don't think the `target_build_utils` crate will be helpful, it is deprecated and recommends using the 
it says:

> DEPRECATED: Use environment variables set by cargo from 1.14.0 onwards.
> They look like this: CARGO_CFG_TARGET_OS, CARGO_CFG_TARGET_ENV…

According to the Commits dropdown in #449 we have dependencies that require v.1.34.0, so
we can just use the CARGO_CFG_TARGET_* i think.

I did some looking into how to write proc_macros that emulate `cfg!(...)` It seems like that may be possible under v.1.30+, I haven't managed to think of any other API which is going to be nearly as nice, anyhow i will play around with various options and scour crates.io to see if we missed anything more recent.

Thanks for the tip to the CI setup, i'll have a look.

pkgw:
Ah, good reading comprehension :+1:

Given that proc-macros make life hard for static builds, I would like to avoid using them for the `build.rs` script. I agree that it will probably be much less nice, but on the other hand, `build.rs` is just a bit of infrastructure, not the main API.

ratmice:
Thanks, i'll have a look with that in mind.

pkgw:
Fixed in #477!

PR #477, opened 2019-10-05
move target directed cfg! target checks to using a cfg_support module.

ratmice:
Here is a first stab at #476

Test coverage in this library isn't complete, It's a little terrible.
It could probably be it's own crate instead of added as a subdir module.  
but i wanted to post for feedback before commiting to any API..

There is still some `#[cfg(...)]` stuff left in build.rs, but those are generally dealing with running host tools.  This doesn't currently implement cargo-feature checking, which seems to require it's own special mechanism.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -351,7 +351,7 @@ fn main() {
 
     // Platform-specific adjustments:
 
-    let is_mac_os = build_cfg!(target_os = "macos");
+    let is_mac_os = target_cfg!(target_os = "macos");
 
     if is_mac_os {
         ccfg.define("XETEX_MAC", Some("1"));
@@ -373,7 +373,7 @@ fn main() {
         cppcfg.file("tectonic/xetex-XeTeXFontMgr_FC.cpp");
     }
 
-    let is_big_endian = build_cfg!(target_endian = "big");
+    let is_big_endian = target_cfg!(target_endian = "big");
     if is_big_endian {
         ccfg.define("WORDS_BIGENDIAN", "1");
         cppcfg.define("WORDS_BIGENDIAN", "1");
--- a/build.rs
+++ b/build.rs
@@ -8,8 +8,8 @@
 ///
 /// TODO: this surely needs to become much smarter and more flexible.
 use cc;
-use cfg_support::{CfgBuilder, CfgTarget};
 use pkg_config;
+use tectonic_cfg_support::*;
 use vcpkg;
 
 use std::env;
@@ -351,8 +351,7 @@ fn main() {
 
     // Platform-specific adjustments:
 
-    let target_info = CfgTarget::new();
-    let is_mac_os = target_info.all(&[CfgBuilder::new().target_os("macos").build()]);
+    let is_mac_os = build_cfg!(target_os = "macos");
 
     if is_mac_os {
         ccfg.define("XETEX_MAC", Some("1"));
@@ -374,7 +373,7 @@ fn main() {
         cppcfg.file("tectonic/xetex-XeTeXFontMgr_FC.cpp");
     }
 
-    let is_big_endian = target_info.all(&[CfgBuilder::new().target_endian("big").build()]);
+    let is_big_endian = build_cfg!(target_endian = "big");
     if is_big_endian {
         ccfg.define("WORDS_BIGENDIAN", "1");
         cppcfg.define("WORDS_BIGENDIAN", "1");
@@ -398,11 +397,6 @@ fn main() {
 
     // Tell cargo to rerun build.rs only if files in the tectonic/ directory have changed.
     for file in PathBuf::from("tectonic").read_dir().unwrap() {
-        let file = file.unwrap();
-        println!("cargo:rerun-if-changed={}", file.path().display());
-    }
-    // ditto for cfg_support
-    for file in PathBuf::from("cfg_support").read_dir().unwrap() {
         let file = file.unwrap();
         println!("cargo:rerun-if-changed={}", file.path().display());
     }
--- a/build.rs
+++ b/build.rs
@@ -8,6 +8,7 @@
 ///
 /// TODO: this surely needs to become much smarter and more flexible.
 use cc;
+use cfg_support::{CfgBuilder, CfgTarget};
 use pkg_config;
 use vcpkg;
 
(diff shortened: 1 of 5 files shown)


ratmice:
I need to look through previous iterations of the big endian CI target, and see if this:
`warning: tectonic/dpx-dpxcrypt.c:322:1: warning: '_gcry_bswap64' defined but not used [-Wunused-function]` Is in fact a new warning caused by this patch.

pkgw:
@ratmice That's not new; see e.g. [the logs](https://circleci.com/gh/tectonic-typesetting/tectonic/755?) for the most recent master branch commit for which the CircleCI finished.

cfg_support/src/lib.rs
```diff
@@ -0,0 +1,335 @@
...
+    Uninitialized,
+    Cond(Cfg<'a>),
+}
+
+fn check_same(target: String, other: Option<&'_ str>) -> bool {
+    Some(target) == other.map(|x: &str| x.to_uppercase())
+}
+fn check_conflict(target: String, other: Option<&'_ str>) -> bool {
+    None == other || check_same(target, other)
+}
```
pkgw: I think there ought to be a way to implement this function such that `target` is also an `&str`, not a `String`, saving some allocations.

The function name `check_conflict` sounds to me like it should return `true` if there *is* a conflict. I would suggest renaming to `check_no_conflict`.

cfg_support/src/lib.rs
```diff
@@ -0,0 +1,335 @@
...
+        check_conflict((&*arg.target_endian).into(), self.endian)
+    }
+
+    pub fn pointer_width_conflict(&'a self, arg: &CfgTarget) -> bool {
+        check_conflict((&*arg.target_pointer_width).into(), self.pointer_width)
+    }
+
+    pub fn vendor_conflict(&'a self, arg: &CfgTarget) -> bool {
+        check_conflict((&*arg.target_vendor).into(), self.vendor)
+    }
```
pkgw: I wouldn't expect these functions to be used individually, and it would save a lot of lines of code to just include their checks inside the one `check()` function rather than declaring all of these.

cfg_support/src/lib.rs
```diff
@@ -0,0 +1,335 @@
...
+        assert_eq!(args.is_empty(), false);
+        args.iter().all(|other| other.check(self))
+    }
+}
+
+impl Default for CfgTarget {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
pkgw: When this construct comes up, it means that the `new()` function should just be removed and turned into the implementation of the `default()` function.

cfg_support/src/lib.rs
```diff
@@ -0,0 +1,335 @@
...
+            CfgBuilder::Cond(inner) => *inner,
+        }
+    }
+}
+
+impl<'a> Default for CfgBuilder<'a> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
pkgw: Ditto as above.

build.rs
```diff
@@ -350,7 +351,10 @@ fn main() {
 
     // Platform-specific adjustments:
 
-    if cfg!(target_os = "macos") {
+    let target_info = CfgTarget::new();
+    let is_mac_os = target_info.all(&[CfgBuilder::new().target_os("macos").build()]);
```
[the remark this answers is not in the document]
ratmice: The macro I came up with isn't as flexible as the cfg! macro, but provides some cleanup,
I'll need something more powerful for the profile branch's cfg checks, but I'm not skilled enough in macro-fu,  The macro i came up with basically turns the any/all checks inside out, as it works on arrays of booleans.

I left the any()/all() functions there for the more advanced use case the profile branch is going to need however.

build.rs
```diff
@@ -396,4 +401,9 @@ fn main() {
         let file = file.unwrap();
         println!("cargo:rerun-if-changed={}", file.path().display());
     }
+    // ditto for cfg_support
+    for file in PathBuf::from("cfg_support").read_dir().unwrap() {
+        let file = file.unwrap();
+        println!("cargo:rerun-if-changed={}", file.path().display());
+    }
```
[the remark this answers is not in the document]
ratmice: Apparently not! good catch!

cfg_support/src/lib.rs
ratmice: Fixed the 2nd case, for the first case, I couldn't figure out a way, the issue is with
[std::str::to_uppercase](https://doc.rust-lang.org/std/primitive.str.html#method.to_uppercase)
We could avoid the copy by using: [std::str::make_ascii_uppercase](https://doc.rust-lang.org/std/primitive.str.html#method.make_ascii_uppercase)

If it's absolutely necessary I'll do so in a second pass, but I wanted to get the rest of your comments out of the way first.

cfg_support/src/lib.rs
ratmice: The only issue I think is that default() when gobbling up the environment variables, is non-deterministic, and there is some discussion on whether that is an appropriate default.

Anyhow, I did as you mentioned deterministic or not it's the only way to construct one!

pkgw:
FYI, I merged in `master` and de-conflicted. I will try to take another look at this and see if we can get it merged — I might make a few touchups if anything leaps out at me.

ratmice:
Thanks for that, by all means if you can figure any ways that might be nicer,
I should probably go through #473 and try to do that on top of this before merge?
I think it'll be a bit too complex of a use-case for this macro unfortunately.

pkgw:
Ummmmm so I got kind of obsessed with the challenge of replicating `cfg!` syntax and spent a while figuring out how to allow it. I have a solution that I'm happy with, but it kiiiiinda rewrites the entire module. Do you mind having a look and seeing if what I've come up with seems reasonable to you? I tried to document how the macro works but it's certainly pretty complex.

ratmice:
I haven't gotten through the macro yet, but the usage of lazy_static! to deal with lifetimes coming out of getenv really cleans things up, I hadn't thought of that.

I haven't understood the macro yet, but so far the only complaint I have is you don't give yourself any credit in the authors field of Cargo.toml.

pkgw:
Fair point. I also just added a couple of links that helped me a lot as I worked on this:

- https://users.rust-lang.org/t/top-down-macro-parsing-or-higher-order-macros/8879
- https://danielkeep.github.io/tlborm/book/pat-incremental-tt-munchers.html

PR #487, opened 2019-10-20
PlainStatusBackend

Mrmaxmeier:
This switches to plain output if stdout is not a tty.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ tectonic_xdv = { path = "xdv", version = "0.1.9-dev" }
 termcolor = "^1.0"
 toml = "^0.5"
 zip = { version = "^0.5", default-features = false, features = ["deflate"] }
+atty = "0.2.13"
 
 [features]
 default = ["serialization"]
--- a/src/bin/tectonic.rs
+++ b/src/bin/tectonic.rs
@@ -4,6 +4,7 @@
 
 use tectonic;
 
+use atty;
 use structopt::StructOpt;
 
 use std::fs::File;
@@ -15,10 +16,11 @@ use tectonic::config::PersistentConfig;
 use tectonic::driver::{OutputFormat, PassSetting, ProcessingSessionBuilder};
 use tectonic::errors::{ErrorKind, Result};
 use tectonic::io::zipbundle::ZipBundle;
+use tectonic::status::plain::PlainStatusBackend;
 use tectonic::status::termcolor::TermcolorStatusBackend;
 use tectonic::status::{ChatterLevel, StatusBackend};
 
-use tectonic::{ctry, errmsg, tt_error, tt_error_styled, tt_note};
+use tectonic::{ctry, errmsg, tt_error, tt_note};
 
 #[derive(Debug, StructOpt)]
 #[structopt(name = "Tectonic", about = "Process a (La)TeX document")]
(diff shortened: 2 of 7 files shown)


src/bin/tectonic.rs
```diff
@@ -189,10 +190,11 @@ fn inner(
                     "something bad happened inside {}; its output follows:\n",
                     engine
                 );
-                tt_error_styled!(status, "===============================================================================");
+
+                tt_error!(status, "===============================================================================");
```
pkgw: Am I right that this changes the output in these situations to add an `error: ` prefix to the equals-sign lines?

src/bin/tectonic.rs
```diff
@@ -237,8 +239,19 @@ fn main() {
...
 
-    let mut status =
-        TermcolorStatusBackend::new(ChatterLevel::from_str(&args.chatter_level).unwrap());
+    let chatter_level = ChatterLevel::from_str(&args.chatter_level).unwrap();
+    let use_cli_color = match &*args.cli_color {
+        "always" => true,
+        "auto" => atty::is(atty::Stream::Stdout),
+        "never" => false,
+        _ => unreachable!(),
+    };
```
pkgw: For what it's worth, I have a memory that the Rust CLI WG had some small recommended crate whose job was to help you decide whether to use color or not ... it also checked an environment variable of some kind, I think. But now I'm not able to dig it up, so whatever.

src/status/mod.rs
```diff
@@ -77,6 +78,8 @@ pub trait StatusBackend {
             None,
         )
     }
+
+    fn dump_to_stderr(&mut self, output: &[u8]);
```
pkgw: If this is going to go inside the base trait, I think it should have a more generic name ... `dump_error_logs` or something? Also, perhaps the bit in the CLI that delimits the dump with equals signs could go inside this function? That feels like it might work nicely.

src/status/plain.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl PlainStatusBackend {
+    pub fn new(chatter: ChatterLevel) -> Self {
+        PlainStatusBackend { chatter }
+    }
+}
+
+impl StatusBackend for PlainStatusBackend {
+    fn report(&mut self, kind: MessageKind, args: Arguments, err: Option<&Error>) {
+        let prefix = match kind {
```
pkgw: I think this function needs to respect the `chatter` setting?

Cargo.toml
```diff
@@ -57,6 +57,7 @@ tectonic_xdv = { path = "xdv", version = "0.1.9-dev" }
 termcolor = "^1.0"
 toml = "^0.5"
 zip = { version = "^0.5", default-features = false, features = ["deflate"] }
+atty = "0.2.13"
```
pkgw: I try to keep these lists alphabetized, so this new entry should go near the top. Also, can we just specify the version as "0.2"?

pkgw:
Oh, also: if you find the chance to stub some documentation in the book and/or add a log entry to CHANGELOG.md, that would be great!

ralismark:
It looks like this PR has been inactive for quite a while. If there's anything that still needs to be added I can do it and open a new PR.

pkgw:
@ralismark I had a few stylistic comments but mostly we just sort of lost momentum here. If you want to rebase and circle back, that would be most welcome! I have aspirations to go back through the old PR's and see if I can get them across the finish line, but I could really use help because I am *super* low on free cycles these days.

PR #491, opened 2019-10-29
Make toml crate optional

malbarbo:
The toml crate is only needed when serialization feature is enabled, so it can be made optional.

i had to create dump errors to put in place of the toml errors in the error-chain setup.

This also tests building with --no-default-features.


pkgw:
Thanks! Apologies for taking a while to get to this.

PR #492, opened 2019-10-29
Add DirBundle

malbarbo:
DirBundle allow reading bundle files from a directory. This is much faster than reading the files from a ZipBundle. For someone working offline this is good.


pkgw:
Thanks for the contribution! I feel like there are some things to think about regarding the more general issue of if/how to support directories with files, but this is a pretty targeted change to the CLI client so I think it should be OK to merge as-is.

PR #493, opened 2019-10-31
Zero memory upon allocation in bibtex.

jneem:
This fixes an issue I'm having with memory corruption when calling bibtex multiple times. I have some sample code demonstrating the problem in https://github.com/jneem/tectonic-bibtex-test. Before this patch, running that code reliably produces valgrind warnings (and occasionally crashes). After this patch, it doesn't.


pkgw:
Thanks very much!

Issue #373, opened 2019-05-05
What are the available command line arguments?


PHPirates:
I couldn't find anything about it, and #62 suggests that it's not documented yet, just as the main website page.

The output of a compilation suggests that `--keep-intermediates` is one. 
I am especially interested in support for dvi output, synctex (#56, that's not a command line option?), and `-shell-escape` (but I see #38 is still open) but a more complete list would be kind of nice. In fact, any documentation would be kind of nice ;) 

I am willing to contribute a bit to the documentation, but I have no clue at all where to look in the source code.

pkgw:
Yeah ... it's frankly embarrassing for me :-( I appreciate your persistence in working with the project despite its shortcomings!

Contributions are, of course, more than welcome (in the not passive-aggressive sense!). The arguments to the command-line client are defined [here](https://github.com/tectonic-typesetting/tectonic/blob/master/src/bin/tectonic.rs#L158).

Still to be determined: a framework for writing and publishing non-API docs. I get dissatisfied by GitHub Pages in a lot of ways, but a basic setup would be much better than the "nothing" that we currently have to offer.

PHPirates:
No worries, I think Tectonic is useful as it is, and shortcomings are there to be fixed :) 

Thanks a lot for that link, apparently Tectonic has more features than I thought!

Yeah I found that once some template is in place for the documentation, it becomes much easier to work on it. It doesn't have to be perfect from the start. The only other thing I can suggest is the GitHub Wiki, because you can use Markdown on there. I don't know if it's a good idea to make it publicly editable, no experience with that. I recently started a wiki documentation using Asciidoc, and so far I found it works fine for me, I'm editing it locally.

bjornbm:
Regarding the command line arguments specifically I find that the `--help` documentation is quite good:

```
$ tectonic --help
Tectonic 0.1.11
Process a (La)TeX document

USAGE:
    tectonic [FLAGS] [OPTIONS] <INPUT>

FLAGS:
    -h, --help                  Prints help information
    -k, --keep-intermediates    Keep the intermediate files generated during processing
        --keep-logs             Keep the log files generated during processing
    -C, --only-cached           Use only resource files cached locally
    -p, --print                 Print the engine's chatter during processing
        --synctex               Generate SyncTeX data
    -V, --version               Prints version information

OPTIONS:
    -b, --bundle <PATH>            Use this Zip-format bundle file to find resource files instead of the default
[...]

PR #505, opened 2019-11-28
Start documenting cli options

PHPirates:
As mentioned #373, I started with just copying the output from `tectonic --help` into markdown. If you agree, I will document some more tips and other things that are now sort of hidden in old issues etc.

I don't understand where the option `--hide` would be used for, can you give an example?

### DIFF
--- a/docs/src/cli/index.md
+++ b/docs/src/cli/index.md
@@ -1,4 +1,4 @@
-# Tectonic’s Command Line Interface
+# Tectonic's Command Line Interface
 
 Tectonic has many command line options.
 If you have Tectonic installed, you can view them with `tectonic --help`.
--- a/docs/src/cli/index.md
+++ b/docs/src/cli/index.md
@@ -1,3 +1,37 @@
 # Tectonic’s Command Line Interface
 
-So much to add here! [Go do it!](github.com/tectonic-typesetting/tectonic/pull/new).
+Tectonic has many command line options.
+If you have Tectonic installed, you can view them with `tectonic --help`.
+
+The following are the available flags.
+| Short | Full                      | Explanation                                                                                    |
+|:------|:--------------------------|:-----------------------------------------------------------------------------------------------|
+| `-h`  | `--help`                  | Prints help information                                                                        |
+| `-k`  | `--keep-intermediates`    | Keep the intermediate files generated during processing                                        |
+|       | `--keep-logs`             | Keep the log files gener ated during processing                                                |
+| `-C`  | `--only-cached`           | Use only resource files cached locally                                                         |
+| `-p`  | `--print`                 | Print the engine's chatter during processing                                                   |
+|       | `--synctex`               | Generate SyncTeX data                                                                          |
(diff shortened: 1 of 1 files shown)


pkgw:
Thanks!

The `--hide` option is needed to get repeatable multi-stage document builds with Tectonic using a dependency system like `make` or [ninja](https://ninja-build.org/). For example, I have a workflow that uses a tool that processing the `.aux` file during a build to produce an updated `.bib` file. In pseudo-Makefile code, the workflow looks like this:

```
paper.aux: paper.tex
    tectonic --pass=tex --outfmt=aux paper.tex

paper.bib: paper.aux
    ./generate-bibtex.py <paper.aux >paper.bib

paper.pdf: paper.tex paper.bib paper.aux
    tectonic --pass=bibtex_first paper.tex
```

The problem with the rules as written above is that after the first build, the file `paper.aux` exists on disk, and gets used by the first build command. This essentially introduces a circular dependency, which you can work around in `make` but really makes things a lot more complicated and unstable. By adding a `--hide=paper.aux` option to the first `tectonic` command, the circular dependency is prevented.

Hopefully that helps. Would you want to update this PR with any further changes, or is it good to merge now?

PHPirates:
Wait, you use Tectonic that way (with the python script to update the bib file) or is that how Tectonic implements things? The workflow isn't making any sense to me, why and with what would you update a bib file from the aux file?

Did you use things like Ninja yourself? Someone once tried to get me to use Bazel, in order to have reproducible builds, but I never managed to get it working. I'm not familiar with Ninja, but if it works well with Tectonic it may be useful to note somewhere (if it isn't already).

Thanks for the explanation, I think it makes sense to document it once I understand the use case of the workflow you mentioned, but for now you can merge this PR and I'll create a new one when I have additions.

PHPirates:
I don't know why the circleci build is failing, is it getting stuck on compiling arrayvec and then just timing out on the no_output_limit you set?

pkgw:
@PHPirates The CircleCI failure is a known intermittent issue associated with the fact that that build happens in a QEMU environment. It can safely be ignored.

pkgw:
As for your other question, the example Makefile is a workflow that I use. I have [a tool](https://github.com/pkgw/bibtools) that maintains my bibliography in a real database, rather than a giant BibTeX file, and it has a mode where it can emit a `.bib` file for only the entries used in the `.aux` file, with some options for how they are rendered because BibTeX style files aren't always completely compatible with each other.

I use Ninja for all of my builds; it's really nice! I posted a forum thread documenting my workflow a while ago.

PHPirates:
Thanks, I found the [blog post](https://tectonic.newton.cx/t/tectonic-ninja-reproducible-augmented-document-builds/20), and I see you automated a solution for one of the problems that I talked about with someone only last week: when you maintain a central bib repository it can get so large that it's unmaintainable (he used a makefile and a huge central .bib file).

It sounds promising, I will try it out and copy it to the mdbook also when it works.

PR #511, opened 2019-12-03
 omit-frame-pointer and profile feature (take 2)

ratmice:
Take 2 on Pr #473, built on top of #477 so that would need to go in first.

It only required a few minor changes, (bump to released cc-rs),
using let instead of const, and of a closure for profile_config instead of a local fn.

[comparison against PR 477](https://github.com/ratmice/tectonic/compare/cfg_support...ratmice:pr_473-take2?expand=1)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ name = "tectonic"
 crate-type = ["rlib"]
 
 [build-dependencies]
-cc = "^1.0"
+cc = "^1.0.46"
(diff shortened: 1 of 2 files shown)


ratmice:
Odd... it looks like my change to build.rs wasn't recompiling build.rs so everything appeared to work locally until I did a clean rebuild :(

pkgw:
Thanks, looks good! I just made asmall tweak to take advantage of the capabilities of the fancy target_cfg macro and tried to clarify the associated comment.

PR #525, opened 2019-12-17
Add features to allow choosing between reqwest or curl for downloading

malbarbo:
Tectonic used to use hyper, but in https://github.com/tectonic-typesetting/tectonic/pull/330 it was replaced by reqwest. This PR allow using curl as an alternative to reqwest.

One reason to use curl instead of reqwest is that the number of dependencies is smaller (117 vs 239), reducing the compilation time (9min vs 15min of cpu itme) and size of the binary (3,7MB vs 7,2MB).

Another reason is that curl have builtin support for proxy (although I haven't test it) which would solves https://github.com/tectonic-typesetting/tectonic/issues/59.

The curl crate will build libcurl if one is not installed in the build system and static link it,  so this does not increase the native dependencies.

This also adds a mock download backend allowing to build tectonic without download support, which can be interest to assert that tectonic cannot make network connections.


src/io/download_curl.rs
```diff
@@ -0,0 +1,96 @@
...
+pub fn get_range_inclusive(
+    client: &mut Client,
+    url: &str,
+    start: u64,
+    end: u64,
+) -> Result<Response> {
+    client.get(url, Some((start, end)))
+}
+
+pub fn resolve_url(url: &str, _status: &mut dyn StatusBackend) -> Result<String> {
```
malbarbo: I could not implement something similar to the reqwest implementation, but I not sure that the implementation is really necessary.

pkgw:
Thanks — I appreciate the work you're doing to increase the system's flexibility here!

The implementation looks solid. The only comment I have is that there should be a bit of documentation. Would you mind adding some text to README.md describing the feature, and CHANGELOG.md listing the change for the next release notes?

XVilka:
Would be nice to have this also in the oxidized fork of Tectonic - https://github.com/crlf0710/tectonic

pkgw:
@malbarbo I've pushed a commit that merges current `master` into your branch here, which pulls in the updated CI/CD infrastructure that I've put together.

pkgw:
It took a year, but I have finally merged this functionality in #726! 

```
cargo build --no-default-features --features "serialization geturl-curl"
```

should do it. This is one of the the things that I feel much better implementing using the new sub-crate architecture, made possible by cbindgen, a better understanding of how Rust's sys crates work, and [Cranko](https://pkgw.github.io/cranko/).

malbarbo:
Thanks @pkgw for moving this forward.

Issue #475, opened 2019-10-01
Wire in support for `-Z` options


pkgw:
Now that we have structopt (#465), the next thing I'd like to do is add support for `-Z` options as used by programs like `rustc` and `cargo`. The basic idea is to give us a way to let people activate unusual engine features, like shell-escape (#38) or min-crossrefs (#175), without proliferating an endless series of long command-line options:

```
tectonic -Z bibtex-min-crossrefs=1000
```

In particular, I think that Rust folks have gotten a sense that `-Z` conveys that features are unstable and might go away. I certainly don't expect us to have a nice stable/unstable split any time soon but we might eventually get there.

The last time I looked at this, I got stuck since it was going to take a little bit of effort to figure out how to get the CLI frontend to control the backend driver, but fundamentally there shouldn't be anything too complicated here.

ralismark:
I've been experimenting with this on the [experiments branch of my fork](https://github.com/ralismark/tectonic/tree/experiments). My current setup is to have an [`UnstableOptions` struct](https://github.com/ralismark/tectonic/blob/experiments/src/unstable_opts.rs) which gets passed to the various engines in src/engines/. The individual engines then interpret the relevant options from the struct.

My current method of parsing -Z arguments is to prepend `--` to each value and parse it into UnstableOptions using structopt ([see here](https://github.com/ralismark/tectonic/blob/experiments/src/unstable_opts.rs#L34-L44)). It works, but the error messages produces are very unhelpful - do you know of a cleaner way to parse these options that still has decent error messages?

ratmice:
My inclination when looking at the change [here](https://github.com/ralismark/tectonic/blob/experiments/src/bin/tectonic.rs#L78-L84) is whether this could use the [possible_values](https://stackoverflow.com/questions/59115832/how-to-attach-possible-values-to-a-struct-with-structopt) instead of going from `Vec<String>`, using that to go directly to `UnstableOpts` in the top-level struct.

Perhaps theres a reason that doesn't work though?

ralismark:
Some options take a value e.g. `-Z bibtex-min-crossrefs=1000` so I don't think we can use `possible_values`. I suppose a [custom string parser](https://docs.rs/structopt/0.3.18/structopt/#custom-string-parsers) could be used, but that doesn't address the error message issue.

pkgw:
[...]

Issue #175, opened 2018-05-01
bibtex min crossrefs


andronat:
I've been heavily relying in the `-min-crossrefs=1000` flag of `bibtex`? How can I pass this flag to tectonic? I can't find a way.

pkgw:
The version of `bibtex` included in Tectonic does support this feature internally, but we don't currently have any way for the user to change the setting. I'm not familiar with its purpose — can you please explain to me why you sometimes use this option? Thanks.

andronat:
Sorry for the late response! We use the flag in my research group extensively as we have a huge database of references and cross-references to make the setup as modular as possible. Due to the nature of our architecture we use `-min-crossrefs` to produce "flat" references instead of cross-references. You can see an example [here](https://tex.stackexchange.com/questions/191231/expand-citations-with-bibtex-or-disable-nested-citations). Do you think you can allows access to this flag in the next version? It's a deal breaker for me right now :(

andronat:
ping. any news about this?

pkgw:
I'm afraid not:

Part of the reason is that this particular feature request ties in to a bigger architectural design question that I've been struggling with. I see the need to control parameters like these, but I don't want the `tectonic` CLI program to gain a ridiculous number of command line arguments. Instead I tend to think that it should act more like the Rust `cargo` tool, where the build settings are stored in a file named something like `Tectonic.toml` and all the user needs to do is run something like `tectonic build`.

That change would totally break the command line experience, and I think it would do so in a way that would be *very* annoying to users. On the other hand, I think it's the right thing to do, and if that's the case, the earlier it's done, the better.

Anyway, the basic "problem" is that I'm not sure how to add a UI knob to control this option in a way that will be good for the project long-term. Here are two ideas:

1. If you can compile Tectonic yourself, I believe it is literally a one-line patch to change the default value of this setting. If you need it to stay at zero for other uses, you could have a special copy of the executable named something like `tectonic_mcr1k` used for the relevant documents.
2. Oh! I just thought of this: I could add a flag along the lines of the `-Z` family used by the Rust toolchain. These activate special experimental features. In Rust, they are only allowed on "nightly" compilers, not the "stable" ones, so people don't accidentally start using them until they're formally blessed. Tectonic doesn't have the infrastructure to go that far, but it could at least emit a big warning that your flag might do bad things or go away in the future.

[...]

PR #534, opened 2019-12-28
Set min_crossrefs to default to 2, not 0.

jneem:
In the original bibtex.web, the default value for min_crossrefs is 2, not 0. Thia matches that behaviour.

The bibtex distributed with texlive also defaults to min_crossrefs = 2 (and it also makes it configurable at run-time, with a CLI arg, which this patch doesn't do).


pkgw:
Thanks! See #175, #475 for some related discussion.

If I had any actual time to work on this, I'd create a prototype new CLI client with a "multitool" style interface (e.g. `tectonic init`, `tectonic build`, `tectonic show-cached-file`) and a `Tectonic.toml` file defining the document in which one could set low-level parameters such as `min_crossrefs`.

PR #536, opened 2020-01-02
Fixing memory leak of shaperList

elliott-wen:
A leak is detected in Engine->shaperList. A simple and safe fix.


pkgw:
Great, thanks!

Issue #539, opened 2020-01-07
Filenames in output should be quoted


alexreg:
This is motivated by the fact I get output like the following in the `LaTeX Compiler` log window in VSCode:

```
Writing /Users/alex/Documents/A filename with spaces.tex (123 bytes)
```

and only `/Users/alex/Documents/A` gets highlighted as a filename. In any case, the general practice is to enclose filenames in console output as `` `/a/b/c` ``, I believe. This would solve the issue.

I don't know if it covers all cases, but the `note_highlighted` fn (second parameter) seems to be responsible for this sort of printing. The enclosing of filenames in backticks is probably best done at call-site however (using a helper function?).

While I'm writing this, a bonus would be comma-separating the number of bytes, e.g., `123,456,789 bytes` -- or better yet, `123 MB` (also `kB` and `B` of course). All for readability of course, since this is nitpicking!

pkgw:
There is almost surely a small Rust crate out there that implements the algorithm of taking a number of bytes and textifying it as `123 MiB` or whatever else as needed. Both of these suggestions are good and would make great pull requests for someone interested in making their first project contribution!

alexreg:
@pkgw I'm really busy at the moment, but if someone doesn't get around to this in the next 2-3 weeks I should be able to do this! (Feel free to ping me around then.)

ralismark:
Can this be closed? Looks like a fix has been merged.

pkgw:
Sure! We can always reopen, or make a new issue, if there's more to do.

alexreg:
Thanks, glad to see this has been fixed!

PR #559, opened 2020-03-21
Improve log formatting

as-f:
Hi!
This pr addresses issue #539 

- Paths and filenames are now quoted in backticks
- File sizes are now expressed in appropriate units (KiB/MiB)

| Before | After |
| --- | --- |
| Writing **file name.pdf** (47203 bytes) | Writing **\`file name.pdf\`** (46.10 KiB) |

When the output has size 0, the "Not writing **\`file name\`**: it would be empty" message also gets backticks.

This adds a new dependency (byte-unit) for expressing file sizes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,6 +104,11 @@ version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "byte-unit"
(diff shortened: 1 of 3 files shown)


pkgw:
CI failures are known issues unrelated to this patch. Thanks!

Closes #539.

PR #579, opened 2020-04-29
Guard against potential segfault

### LINKED ISSUE #574: Segmentation fault on repeated failure under async executor
I am using tectonic as a library for generating pdfs on an [actix-web server](https://actix.rs/).

This only seems to happen when running on different threads under an async executor. The way to trigger a segfault is to repeat an api call that causes a failed latex compilation:

After a first call to `latex_to_pdf` (for example, calling with invalid latex), on a second invalid call [XeTeXFontMgr_FC::initialize](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/xetex-XeTeXFontMgr_FC.cpp#L310) does not get called, and therefore when it tries to terminate [FcFontSetDestroy](https://github.com/tectonic-typesetting/tectonic/blob/master/tectonic/xetex-XeTeXFontMgr_FC.cpp#L338) is called with NULL, segfaulting.
[Here is the full stack trace](https://github.com/tectonic-typesetting/tectonic/files/4504028/segfault.txt)

Placing a guard on the call

```cpp
    if (allFonts != NULL) {
        FcFontSetDestroy(allFonts);
        allFonts = NULL;
    }
```
appears to prevent this, but I am not familiar with this codebase so I can't tell if it is just a patch on the symptoms rather than a fix for the root cause.

fmgoncalves:
Fix #574

### DIFF
--- a/tectonic/xetex-XeTeXFontMgr_FC.cpp
+++ b/tectonic/xetex-XeTeXFontMgr_FC.cpp
@@ -335,8 +335,10 @@ XeTeXFontMgr_FC::initialize()
 void
 XeTeXFontMgr_FC::terminate()
 {
-    FcFontSetDestroy(allFonts);
-    allFonts = NULL;
+    if (allFonts != NULL) {
+        FcFontSetDestroy(allFonts);
+        allFonts = NULL;
+    }
 
     if (macRomanConv != NULL) {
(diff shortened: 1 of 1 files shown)


pkgw:
Thanks! CI failures are a longstanding, unrelated problem with harfbuzz in vcpkg.

PR #587, opened 2020-05-31
Specify target directory implied by the `package` function

thomaseizinger:
By default, cargo places build artifacts in a `target` directory in the root of the current workspace. This behaviour can be overridden through various means like ENV variables: https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir

If that is the case, installation of the package will fail because the tectonic binary will not be in the path assumed by the `package` function.

We fix this by specifying the path of the `target` directory directly for the `build` and `check` functions.


pkgw:
Thanks!

PR #593, opened 2020-06-03
add vcpkg metadata to Cargo.toml

mcgoo:
Hi. I have been using tectonic as a test crate for some vcpkg related work I have been doing. If you were to merge this PR, the following would be sufficient to get the required dependencies installed and run a build.

```
PS> cargo install cargo-vcpkg
PS> cargo vcpkg build
PS> $Env:TECTONIC_DEP_BACKEND="vcpkg"
PS> $Env:RUSTFLAGS="-Ctarget-feature=+crt-static"
PS> cargo build
```

It works the same on Linux and macOS also. This is out of the blue so thanks for taking a look at this PR.


pkgw:
Interesting, thanks! Do you see a place where we should document this functionality? And maybe even add it to the CI test suite? (Although I'll admit that I'm feeling a bit burned about the vcpkg harfbuzz package being broken for several months, causing all of our CI to go red for a while.)

mcgoo:
I'm not game to mess with your CIs, but I did add a Github Actions set, because who doesn't need another one? As with all of this, I'll take no offence if you don't want to merge it.

I've found the vcpkg stuff to be brittle when I track master, too. It is going to be interesting to see if it is much better pinning to a specific revision.

pkgw:
Apologies for taking so long to get back to this. I think we're ready to merge!

I replaced the Github Action with a new Azure pipeline run — we already have a whole bunch of different CI services being used and I'd like to avoid adding another one. I've added an Azure Pipelines run that demos the concept. Should probably go ahead and add variants for other platforms but I want to just get this merged.

PR #606, opened 2020-06-30
Prevent segfault on latex compilation error

fmgoncalves:
This is a followup of sorts to #574, and it turns out the async environment was totally unnecessary. In fact, this snippet is all that is necessary to trigger the segfault (the DejaVu Sans font needs to be installed):
```rust
use tectonic::latex_to_pdf;

fn main(){
        for _ in 0..2 {
            latex_to_pdf(r"\documentclass{article}
\usepackage{fontspec}
\setmainfont{DejaVu Sans}

\begin{document}
\invalidcommand{}
\end{document}");
        }
}
```

This is a legacy issue, of sorts. Since XeTeX is meant to be used as an executable binary, it does not clean up its intermediate compilation data in case of a fatal failure: no further compilations will be tried and the OS will clear up the process' memory. However, tectonic uses the XeTeX as a library, and some variables get abandoned/leaked or its memory locations become invalid in case of a compilation failure. This resulted in segfaults on the font handling code when accessing structures that had been partially invalidated, and in memory leaks when compilation failed but no invalid memory accesses were made.

This tiny refactor tries to force the cleanup code on failures, preventing segfaults and memory leaks.

### DIFF
--- a/tectonic/xetex-errors.c
+++ b/tectonic/xetex-errors.c
@@ -45,6 +45,7 @@ post_error_message(int need_to_print_it)
 
     history = HISTORY_FATAL_ERROR;
     close_files_and_terminate();
+    tt_cleanup();
     ttstub_output_flush(rust_stdout);
 }
 
@@ -102,6 +103,7 @@ fatal_error(const char* s)
     print_cstr("Emergency stop");
     print_nl_cstr(s);
     close_files_and_terminate();
+    tt_cleanup();
     ttstub_output_flush(rust_stdout);
     _tt_abort("%s", s);
 }
--- a/tectonic/xetex-ini.c
+++ b/tectonic/xetex-ini.c
@@ -3900,6 +3900,89 @@ get_strings_started(void)
 }
 /*:1001*/
 
+void
+tt_cleanup(void) {
+    /*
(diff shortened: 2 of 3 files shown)


pkgw:
Thank you for this submission! I am always interested in improvements to this aspect of Tectonic ... it's not a part of my regular usage of the software but it's something that I would like the project to handle really well.

Would you be able to add a program to the test suite implementing this test, so that we can avoid regressions in the future?

If you've got lots of bandwidth to spend on this topic, note that the fuzzing infrastructure in the `fuzz/` directory is very good at detecting and triggering memory leaks.

pkgw:
PS. With this patch, though, I can run the fuzzer for much much longer without getting any crashes — it seems to be making a big difference!

pkgw:
Well, it would be nice to add test coverage here, but this patch seems to work well and I don't want to stall it.

fmgoncalves:
Sorry for taking so long, time is scarce sometimes. I've opened PR #608 with the relevant test, hopefully I've placed it where it makes some sense.

I hadn't noticed the fuzzing infrastructure but I am definitely looking forward to give it a try and report back any bugs (hopefully with accompanying fixes)

PR #608, opened 2020-07-04
Add regression test for previously fixed crash

fmgoncalves:
Test for #606, sorry for the delay


pkgw:
Thanks, much appreciated!

PR #625, opened 2020-08-06
[WIP] Capture warnings into diagnostic objects

### LINKED ISSUE #577: Improve the message about warnings
When warnings are issued during compilation by the TeX engine, `tectonic` prints the following message:

```
warning: warnings were issued by the TeX engine; use --print and/or --keep-logs for details.
```

It seems to me that `tectonic` should be printing out more information about the warning.
Seeing that `tectonic` can point out that warnings have occurred, it must be parsing or otherwise searching the output of the TeX engine. In this case, wouldn't it be quite easy to print out the line that makes `tectonic` think that a warning has occurred?

Of course, one can easily find the warning by using `--keep-logs` and then searching for "Warning" in the generated log but there are at least two reasons why this is not ideal:

1. It's much more convenient to not have to look at the log file.
2. Sometimes, `tectonic` tells me that warnings were issued but I cannot find anything involving a warning in the log file. Therefore I would like to understand what makes `tectonic` decide that a warning has been issued.

In fact, I have recently been compiling a document with `tectonic` during the compilation of which I keep getting the message about warnings despite the log not mentioning any warning. I made [a post about this problem](https://tectonic.newton.cx/t/tectonic-says-warnings-were-issued-but-i-cannot-find-those-warnings/231) in the forum.

ralismark:
This is meant to resolve #577. The general design of this is have each warning/error be reported via a `Diagnostic` object (`diagnostic_t` on the C side) -- doing this will allow extra information to be added later (e.g. location of error). I've also added a sort of hack in `xetex-output.c` to be able to capture print messages into a diagnostic. The main thing left now is to wrap all the places where warnings/errors are emitted with

```c
diagnostic_t warning = ttstub_warn_begin(); // or ttstub_error_begin()
capture_to_diagnostic(warning);
```
and 
```c
capture_to_diagnostic(0);
ttstub_diag_finish(warning);
```

### DIFF
--- a/src/engines/mod.rs
+++ b/src/engines/mod.rs
@@ -467,6 +467,9 @@ impl<'a, I: 'a + IoProvider> ExecutionState<'a, I> {
 #[repr(C)]
 struct TectonicBridgeApi {
     context: *const libc::c_void,
+    warn_begin: *const libc::c_void,
+    warn_finish: *const libc::c_void,
+    warn_append: *const libc::c_void,
     issue_warning: *const libc::c_void,
     issue_error: *const libc::c_void,
     get_file_md5: *const libc::c_void,
@@ -512,6 +515,37 @@ extern "C" {
 }
 
 // Entry points for the C/C++ API functions.
+
+struct Warning {
+    message: String,
+}
+
+extern "C" fn warn_begin() -> *mut Warning {
+    let warning = Box::new(Warning {
+        message: String::new(),
+    });
+    Box::into_raw(warning)
+}
+
+extern "C" fn warn_finish<'a, I: 'a + IoProvider>(
+    es: *mut ExecutionState<'a, I>,
+    warnp: *mut Warning,
+) {
+    let warning = unsafe { Box::from_raw(warnp as *mut Warning) };
+    let es = unsafe { &mut *es };
+
+    tt_warning!(es.status, "{}", warning.message);
+}
+
+extern "C" fn warn_append(
+    warning: *mut Warning,
+    text: *const libc::c_char,
+) {
+    let rwarn = unsafe { &mut *warning };
+    let rtext = unsafe { CStr::from_ptr(text) };
+
+    rwarn.message.push_str(&rtext.to_string_lossy());
+}
 
 extern "C" fn issue_warning<'a, I: 'a + IoProvider>(
     es: *mut ExecutionState<'a, I>,
@@ -810,6 +844,9 @@ impl TectonicBridgeApi {
     fn new<'a, I: 'a + IoProvider>(exec_state: &ExecutionState<'a, I>) -> TectonicBridgeApi {
         TectonicBridgeApi {
             context: (exec_state as *const ExecutionState<'a, I>) as *const libc::c_void,
+            warn_begin: warn_begin as *const libc::c_void,
+            warn_finish: warn_finish::<'a, I> as *const libc::c_void,
+            warn_append: warn_append as *const libc::c_void,
             issue_warning: issue_warning::<'a, I> as *const libc::c_void,
             issue_error: issue_error::<'a, I> as *const libc::c_void,
             get_file_md5: get_file_md5::<'a, I> as *const libc::c_void,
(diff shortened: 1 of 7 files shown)


pkgw:
I really like the look of this! I think you're right that it will be helpful to use a small structure to allow for the capture of line-number information. (The line-numbering option is turned on in Tectonic by default, but I have to admit that I feel like I can never understand what it reports anyway.)

pkgw:
The current Travis failure is due to formatting — our CI requires that `cargo fmt` and `cargo clippy` both be satisfied. Something to keep an eye out for before updating the branch.

tectonic/core-bridge.c
```diff
@@ -110,6 +110,47 @@ bibtex_simple_main(tt_bridge_api_t *api, char *aux_file_name)
...
+
+void
+ttstub_diag_vprintf(diagnostic_t diag, const char *format, va_list ap)
+{
+    vsnprintf(error_buf, BUF_SIZE, format, ap); /* Not ideal to (ab)use error_buf here */
+    TGB->diag_append(diag, error_buf);
+}
+
+void
+ttstub_diag_printf(diagnostic_t diag, const char *format, ...)
```
pkgw, line 145: Can you please annotate this function with the `PRINTF_FUNC` macro? I see that you've used it in the header; TBH I'm not sure if that suffices or whether it also needs to go here, but it is already present for `ttstub_issue_warning` and friends.

src/engines/mod.rs
```diff
@@ -513,6 +517,45 @@ extern "C" {
 
 // Entry points for the C/C++ API functions.
 
+struct Diagnostic {
+    message: String,
+    kind: MessageKind,
+}
+
+extern "C" fn warn_begin() -> *mut Diagnostic {
```
pkgw, line 525: For consistent nomenclature, I think maybe this function should be called `diag_begin_warn`, and an analogous change for `error_begin`. How does that sound to you?

tectonic/xetex-ini.c
```diff
@@ -1198,6 +1220,9 @@ new_hyph_exceptions(void)
                     hc[0] = trie_tro[hyph_index + cur_chr];
 
                 if (hc[0] == 0) {
+                    ttstub_diag_finish(diagnostic_error_here(
+                        "Not a letter"));
+
```
pkgw, line 1225: I'm seeing a lot of cases where you're duplicating the message text (i.e. "Not a letter"). I'd prefer to avoid that if possible. From a quick browser, it looks like it might be helpful to add a helper function of the form:

```
{
  ttstub_diag_finish(diagnostic_error_here(message));
  if (file_line_error_style_p)
    print_file_line();
  else
    print_nl_cstr("! ");
  print_cstr(message);
}
```

tectonic/core-bridge.c
ralismark: Whoops, looks like I just forgot. Thanks for catching this.

src/engines/mod.rs
ralismark: Sounds good, I've committed a fix for this.

tectonic/xetex-ini.c
ralismark: I've added `error_here_with_diagnostic` in xetex-output.c that creates a diagnostic, prints the line number, and sends a message to both. I'm not sure about the name though.

pkgw:
For some reason, codecov seems to be stuck, but CI is looking green otherwise. Is this ready for review?

ralismark:
I've only added the diagnostics to a portion of the places where warnings or errors get emitted - and there's quite a lot of these. Should I do that in this PR or separate those into a new one? Either way I'll clean up the commits before mark this as ready for review.

pkgw:
@ralismark I think it makes sense to have this PR work on the foundation without trying to cover every single message in the codebase. If nothing else, that leaves some nice low-hanging-fruit tasks for new contributors.

ralismark:
This PR should be ready for review & merging now.

PR #643, opened 2020-09-20
Use cbindgen to generate parts of core-bridge header file

ralismark:
This is essentially an updated #246. I encountered a lot of merge conflicts trying to rebase that onto upstream master (it was ~750 commits behind), so I've rewritten it from scratch.

This generates a C header file (`tectonic/core-bindgen.h`) using [cbindgen](https://github.com/eqrion/cbindgen). This header file contains `#[no_mangle] pub extern "C"` functions, as well as type declarations that they use. This makes the API bridge safer by reducing duplicate/unsafe type declaration (in particular, `TectonicBridgeApi`/`tt_bridge_api_t`, which had to be manually synchronised) and removing some unsafe typecasts.

Additionally, setting up cbindgen would also allow the rendering engines to be _gradually_ oxidised in the future, by replacing C functions with exported Rust ones. This change would also allow Tectonic to be used as a library, though that's not the main focus of this PR.

Remaining issues:

- `cargo build` doesn't regenerate `tectonic/core-bindgen.h` when exported function declarations change. Currently, I've set it up to update whenever `src/engines/mod.rs` changes (like in #246 - that file contains almost all exported functions), but this won't catch changes in other files. I'm not that familiar with `build.rs`, so I'm not sure how to handle this.
- ~~(from #246) `cbindgen` introduces `serde` as a build dependency. This breaks the `alpine-linux-musl` target.~~ Musl CI is green, so this is likely no longer an issue.


pkgw:
Thanks, this is exciting! I have long wanted to get cbindgen going to be able to start oxidizing the lower lays as you mention.

Re: your final comment about serde breaking the Alpine/musl target ... Looks like the CI builds are all green, and I *think* that my updates to the musl build approach should mean that this will work OK these days. If I understand correctly your comment was propagated forward from #246 but just to check ... if the CI is green we can cross off that worry, right?

ralismark:
I'm not sure if it's possible to have cbindgen always run, but the compiler invocations only run when the C/C++ files change _or_ the output of cbindgen is different - the `cargo:rerun-if-` instructions only determine whether to run the whole build script.

The easiest solution seems to be to have the build script always run, by removing all `cargo:rerun-if-` instructions.

pkgw:
Sorry, Codecov, I'm going to ignore you again.

pkgw:
Sorry for the delay in processing this. I made one change to output the generated header file in the Cargo build $OUT_DIR, but otherwise it looks great! Thanks again!

PR #646, opened 2020-10-02
tectonic/dpx-cidtype0.c: attempt to fix loop condition

### LINKED ISSUE #639: unreachable
https://github.com/tectonic-typesetting/tectonic/blob/64db8d70ebc76f5d5b29cb7d492e077621a1768f/tectonic/dpx-cidtype0.c#L213

pkgw:
The `if` statement inside the loop body could never succeed. Presuming that the `vertOriginYMetrics` are a sorted list, it looks like we need a `>=` condition.

Closes #639.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,7 +87,7 @@ tokio = "0.1.22"
 
 [package.metadata.vcpkg]
 git = "https://github.com/microsoft/vcpkg"
-rev = "ff1d20fd9a72d72d51a549c27b669cd6040fb7d9"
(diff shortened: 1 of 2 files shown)


pkgw:
Windows vcpkg build failing due to transient error in vcpkg:

https://github.com/microsoft/vcpkg/issues/13852

pkgw:
Codecov is sad because we don't actually have test coverage that demonstrates this fix. This isn't a bad point but I'm gonna YOLO this one anyway.

PR #657, opened 2020-10-14
Support for unstable -Z options

### LINKED ISSUE #475: Wire in support for `-Z` options

ralismark:
This resolves #475 by introducing `UnstableArg` and `UnstableOptions` (in `src/unstable_opts.rs`) which can be used to easily pass options directly to the tex engines. I've also added 3 example options:

- `-Z min-crossrefs=<num>` to replicate the `-min-crossrefs=<num>` flag to bibtex
- `-Z paper-size=<spec>` to set the default paper size (e.g. to `a4`)
- `-Z shell-escape` to enable `\write18` (though the actual command execution isn't implemented yet - I might leave this to another PR)

Remaining issues:

- Running `tectonic -Z help` shows an error about `<input>` not being provided, instead of listing the unstable options.
- We should probably add some tests that use these options.


src/engines/bibtex.rs
```diff
@@ -9,6 +9,12 @@ use super::{ExecutionState, IoEventBackend, TectonicBridgeApi};
 use crate::errors::{ErrorKind, Result};
 use crate::io::IoStack;
 use crate::status::StatusBackend;
+use crate::unstable_opts::UnstableOptions;
+
+#[no_mangle]
+extern "C" {
+    static mut min_crossrefs: i32;
+}
```
pkgw, line 17: I have been trying to minimize the number of direct symbol references with `extern "C"` names and such. Thus far, I've wanted to control these kinds of options through APIs such as `tt_xetex_set_int_variable`. One approach that I'd prefer here would be to add a `tt_bibtex_set_int_variable` here.

Another approach that is now possible would be to start moving this kind of state to the Rust side, and expose it to the C through the new cbindgen support. There could be a `tectonic_get_min_crossrefs()` function or something along those lines. This is appealing to me since it keeps more of the control logic on the safe Rust side.

Would it be too much of a pain to try one of these approaches instead of the direct variable linkage?

src/unstable_opts.rs
```diff
@@ -0,0 +1,99 @@
...
+use std::default::Default;
+use std::str::FromStr;
+
+const HELPMSG: &str = r#"Available unstable options:
+
+    -Z help                     Lists all unstable options
+    -Z min-crossrefs=<num>      Equivalent to bibtex's -min-crossrefs flag - "include after <num>
+                                    crossrefs" [default: 2]
+    -Z paper-size=<spec>        Change the default paper size [default: letter]
+    -Z shell-escape             Enable \write18
```
pkgw, line 21: My biggest comment: I do not want to expose a `shell-escape` option until it is actually implemented. I always find it *super* frustrating when a piece of software gives the impression that it supports something, then you try to use the feature and it doesn't actually work. We should keep the infrastructure that you've put together so far, but make sure not to give the impression that \write18 is available yet.

pkgw (commented):
Yay! Very nice implementation. Besides the issues that you mentioned, I have two main comments.

It would be nice to get some test coverage and documentation for these new options. I can take a look at that side of things once we iterate on the implementation.

src/unstable_opts.rs
ralismark: Sounds good. I've removed it from the help and UnstableArg parsing.

ralismark:
I've added a test for -Z paper-size=a4, but I'm not sure how to generate the output files to check - directly running `xetex -papersize=a4 a4paper.tex` still adds a header for the xetex version to `a4paper.log` and `a4paper.xdv`.

pkgw:
Great work!

pkgw:
This has been released in [Tectonic 0.2.0](https://github.com/tectonic-typesetting/tectonic/releases/tag/tectonic%400.2.0).

PR #664, opened 2020-10-21
also emit cargo metadata after tectonic's own when using vcpkg

mcgoo:
This changes building with the vcpkg dep backend to emit the cargo metadata after tectonic's own in the same way that it does when using pkg-config.


pkgw:
Just out of curiosity, is there a known issue that this PR addresse?

mcgoo:
Yes, I'm just commenting on https://github.com/mcgoo/vcpkg-rs/issues/21, but I wanted to get the PR in first.

PR #681, opened 2020-11-15
src/engines: incorporate remaining data into digest when closing input files

### LINKED ISSUE #679: Tectonic is always running 6 times and reaching the maximum number of runs
When compiling my latex projects, tectonic always runs 6 times (`TeX rerun seems needed, but stopping at 6 passes`), justifying each re-run with
```
Rerunning TeX because "aom.run.xml" changed ...
```

I have isolated 3 of those files and computed their hashes, and they all look identical:
```
$ openssl dgst -sha256 < aom.run.xml       
(stdin)= 21f268eb669dce0181d75b69c8fabe9f7b34874df1adfeb1b34ef220cbef35eb
$ openssl dgst -sha256 < aom.run.xml.1 
(stdin)= 21f268eb669dce0181d75b69c8fabe9f7b34874df1adfeb1b34ef220cbef35eb
$ openssl dgst -sha256 < aom.run.xml.2
(stdin)= 21f268eb669dce0181d75b69c8fabe9f7b34874df1adfeb1b34ef220cbef35eb
```

This happens with both 0.3.0 and 0.3.1, I never used earlier versions, so I don't know if it is a regression.
I run `tectonic` on macOS 10.15.7 using the binaries downloaded from the release page.

The document that I used for this report is available at https://github.com/mseri/aom (to avoid font issues I suggest running `sed -i.bak '8,10d' aom.tex`)

pkgw:
It seems that in some esoteric cases, there are TeX files that are read-then-written from one run to another, but they are only partially read. Because our rerun-detection logic compares what was read to what was written, this sort of incomplete reading triggers a rerun loop.

Here we attempt to fix that by scanning through any remaining unread data when closing an input file and updating the digest if needed. This would also yield the right results for the common TeX pattern of opening an input file and immediately closing it, which we had to special-case before.

Closes #679.

### DIFF
--- a/src/engines/mod.rs
+++ b/src/engines/mod.rs
@@ -469,12 +469,20 @@ impl<'a> ExecutionState<'a> {
             let p: *const InputHandle = &*self.input_handles[i];
 
             if p == handle {
-                let ih = self.input_handles.swap_remove(i);
+                let mut ih = self.input_handles.swap_remove(i);
+                let mut rv = false;
+
+                if let Err(e) = ih.scan_remainder() {
+                    tt_warning!(self.status, "error closing out input {}", ih.name().to_string_lossy(); e);
+                    rv = true;
+                }
+
                 let (name, digest_opt) = ih.into_name_digest();
                 self.events.input_closed(name, digest_opt);
-                return false;
+                return rv;
             }
         }
+
         // TODO: Handle the error better. This indicates a bug in the engine.
         tt_error!(
             self.status,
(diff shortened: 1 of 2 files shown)


pkgw:
NB, the Arxiv regression test flagged that this tweak introduced an error when the engine tries to open a directory. Thanks, testing! Should be fixed now.

pkgw:
Testing against the TeXLive 2020 bundle tests also shows no regressions.

PR #682, opened 2020-11-17
tectonic/xetex-output.c: handle USV 0xFFFF in print()

### LINKED ISSUE #678: print BIGGEST_CHAR
https://github.com/tectonic-typesetting/tectonic/blob/5fbf5be3769ec843ae20353f3f05e0291dbf7c2b/tectonic/xetex-output.c#L250

`s == BIGGEST_CHAR` (0xffff) is not covered here.

I think `s <= BIGGEST_CHAR` should be here.

pkgw:
Closes #678.

### DIFF
--- a/tectonic/xetex-output.c
+++ b/tectonic/xetex-output.c
@@ -247,7 +247,7 @@ print(int32_t s)
 
     if (s >= str_ptr)
         return print_cstr("???");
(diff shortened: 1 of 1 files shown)


pkgw:
The Arxiv test flags a failure because of changed output, but the new output is better because the 0xFFFF character is getting printed out correctly.

PR #690, opened 2020-11-28
Dokerize

snowinmars:
Dockerization is good)

Publish `typesetting/tectonic` into hub.docker.com before merge.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,36 @@
+FROM archlinux
+
+RUN pacman -Syy
+# tectonic build required packages only. Install custom packages below to avoid docker recompiling
+RUN pacman --noconfirm -S \
+        pacman-contrib \
+        wget \
+        tar \
+        sudo \
+        binutils \
+        fakeroot \
+        pkg-config \
+        gcc \
+        coreutils
+
+RUN useradd -m notroot
+RUN echo "notroot ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/notroot
+
+USER notroot
+WORKDIR /home/notroot
+RUN wget -q https://aur.archlinux.org/cgit/aur.git/snapshot/tectonic.tar.gz 2>/dev/null
+RUN tar -zxvf tectonic.tar.gz
+WORKDIR /home/notroot/tectonic
+RUN makepkg -sric --noconfirm
+
+USER root
+# install additional packages here to avoid docker recompiling
+RUN pacman -S --noconfirm ttf-liberation
+RUN pacman -R --noconfirm wget sudo gcc
+RUN paccache -r
+RUN pacman -Scc
+RUN pacman -Qtdq --noconfirm | pacman -Rns --noconfirm -
+
+WORKDIR /data
+VOLUME ["/data"]
+
--- a/README.md
+++ b/README.md
@@ -15,6 +15,19 @@ If you just want to compile TeX documents, you should probably **click through t
 [the main Tectonic website](https://tectonic-typesetting.github.io/)**. This
 page is primarily aimed at folks interested in how Tectonic works “under the hood.”
 
+## Dockerization
+We provide `typesetting/tectonic` docker image that was build from `Dockerfile` in this repo.
+
+You can run it as `docker run -v ./host/dir:/data typesetting/tectonic`
+
+It contains everything but starting command, extend it with
+```dockerfile
+# based on arch linux
+FROM typesetting/tectonic
+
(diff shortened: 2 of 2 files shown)


snowinmars:
And hi again.\
The question is below, this is an intro.

I want to add docker volume support to make tex compiling as fast as possible.\
A lot of time consumes by tex package downloading.\
So I have to mount host directories as tectonic package/cache directories. But I literally can't find any trace of tectonic packages.

```
homk ~/tmp $ ls
tmp.tex

homk ~/tmp $ docker run -it -v ~+:/data snowinmars/ariph-tex bash 
[root@66514c1a00f8 data]# whoami
root
[root@66514c1a00f8 data]# tectonic tmp.tex 
note: this is a BETA release; ask questions and report bugs at https://tectonic.newton.cx/
note: connecting to https://archive.org/services/purl/net/pkgwpub/tectonic-default
note: resolved to https://ttassets.z13.web.core.windows.net/tlextras-2020.0r0.tar
note: downloading index https://ttassets.z13.web.core.windows.net/tlextras-2020.0r0.tar.index.gz
note: downloading SHA256SUM
note: generating format "latex"
note: downloading tectonic-format-latex.tex
note: downloading xelatex.ini
note: downloading latex.ltx
...
...
note: downloading glyphlist.txt
Writing `tmp.pdf` (6.72 KiB)
Skipped writing 2 intermediate files (use --keep-intermediates to keep them)

[root@66514c1a00f8 data]# find / -name "*xelatex*"
[root@66514c1a00f8 data]#
[root@66514c1a00f8 data]# find / -name "*tlextras*"
[root@66514c1a00f8 data]#
```

I don't believe in magic flows, the packages should exist.\
So I went further and compare `/` before and after package downloading. No new files was added to the system.

```diff
# I remove /proc changes
> /data/after.txt
> /data/tmp.pdf
< /usr/include/linux/netfilter_ipv4/ipt_TTL.h
> /usr/include/linux/netfilter_ipv4/ipt_TTL.h
< /usr/share/terminfo/E/Eterm-color
> /usr/share/terminfo/E/Eterm-color
< /usr/share/terminfo/L/LFT-PC850
> /usr/share/terminfo/L/LFT-PC850
< /usr/share/terminfo/N/NCRVT100WPP
> /usr/share/terminfo/N/NCRVT100WPP
< /usr/share/terminfo/P/P9-W
> /usr/share/terminfo/P/P9-W
```

So, that means that each and every change has been written down to `/usr/include/linux/netfilter_ipv4` and `/usr/share/terminfo` directories.

Am I right, that tectonic works only with these directories? All the stuff that tectonic wraps, ends here?

pkgw:
Hi,

Thanks for asking. Tectonic's caching layer does not store files under their TeX names, which is why you're not finding them. The main cache directory is in `~/.cache/Tectonic`, which has several subdirectories for different categories of cached items.

The actual cache files are in the `files` subdirectory, with a Git-like directory hierarchy where each file is stored in a location depending on the SHA256 digest of its contents.

The mapping from TeX filename to SHA256 digest is recorded in the index file(s) stored in the `manifests` subdirectory.

The manifest name is another SHA256 digest derived from the total contents of the TeX file collection "bundle". The cache files in the `urls` subdirectory tell the engine how to map from bundle URL to manifest SHA256, so that the engine can avoid initiating any network connections unless it needs to find new files.

This should all be documented but it is not, currently :-(

Anyway, I am not sure if it makes sense to distribute an "official" Tectonic Docker image, since the most useful aspect of such an image is the pre-caching step, and it's going to be application-dependent as to what needs to be cached. And it isn't very hard to prepare a custom Tectonic Docker container. I think it might be better to add some documentation show people how to create their own Tectonic-enabled Docker images.

jnnksdev:
> The actual cache files are in the files subdirectory, ...

Why so complicated? Seems like a fairly intransparent implementation.

pkgw:
@jnnksdev The design goals are as follows:

- No need to touch the network if no new resource files are needed
- Able to function in a degraded mode if the network is unavailable
- No failures if a new bundle is selected, or if its contents change
- No failures when multiple instances of Tectonic are modifying the cache simultaneously
- And it is *not* a design goal that the on-disk cache structure be human-friendly.

Given all of the above, the current scheme is the best one I could come up with.

pkgw:
Closing due to inactivity. I don't think there's anything wrong with the idea of providing official Docker images, but there's some design work to think about what those should look like, given that installing Tectonic into a custom Docker image is (or at least ought to be!) very easy.

PR #691, opened 2020-11-28
Fix arch linux makedepends

snowinmars:
#690 fails without these dependencies.

I set them in Dockerfile manually, but I think this pr is the right way.

`gcc` is required for `proc-macro2`.
`pkg-config` is required for `rust-openssl`.


pkgw:
Thanks!

Issue #700, opened 2020-12-09
Auto-update tectonic-bin on AUR as part of release automation


oblitum:
Hi. I've commented this on AUR, but I'm afraid that maybe no one hears comments there as it seems the maintainer is automated, so I'm copying it here:

> Hi. This package is listed on the [Pre-built binary packages docs section](https://tectonic-typesetting.github.io/en-US/install.html#pre-built-binary-packages), but it's not a binary package. Sure, it's not a `-bin` package, but docs made me think otherwise, so I used yay to install it right away, and now it's still building on my low-end laptop....
>
> Will there be a `-bin` package?
>
> FWIW, the GitHub release assets does provide prebuilt binaries for Linux.

Rust packages on AUR often provide a  `-bin` version because Rust build can often be processor and time consuming.

pkgw:
Thanks for reporting. I'm afraid that I'm not an Arch user and I'm not familiar with its packaging ecosystem, but I'm happy to merge any improvements to our automated system. @snowinmars recently submitted some updates in this area and might understand the relevant issues better?

snowinmars:
Let me investigate it a bit.

oblitum:
@snowinmars example Rust packages which provide `-bin` versions are, for example, `ytop-bin`, `zenith-bin`, `dust-bin`, etc.

pkgw:
@oblitum OK, I am learning a bit about how this all works. I guess the basic thing is that our installation page is erroneous because I thought that our Arch package resulted in pre-built binaries. I thought that Arch Linux provided those? Is there a way to improve the existing package so that most users can obtain pre-built binaries using it? I can see how we could add a `-bin` package deriving from our official release artifacts (see discussion in #702) but there's no reason that we shouldn't be able to build Tectonic binaries within the Arch framework.

oblitum:
>  I thought that Arch Linux provided those?

Not in such a way. The Arch Linux repositories are divided into two fields, [official repositories](https://wiki.archlinux.org/index.php/Official_repositories) (binaries), and AUR, which are completely user provided, non-official, not verified, these may or may not provide binaries, source version, or both.

> Is there a way to improve the existing package so that most users can obtain pre-built binaries using it?

[...]

PR #702, opened 2020-12-13
Add bin PKGBUILD for Arch Linux

snowinmars:
Made an example of arch linux bin packaging.
Seems ok on my machine, but I didn't test in anywhere else.

The package source link should be replaced with something meaningful. That's just my host from disk.yandex.ru

### DIFF
--- a/dist/arch/PKGBUILD.bin
+++ b/dist/arch/PKGBUILD.bin
@@ -1,0 +1,27 @@
+# Maintainer: tectonic-deploy <sasha+tectonic@hackafe.net>
+# Maintainer: Daniel M. Capella <polyzen@archlinux.org>
+# Contributor: Jan Tojnar <jtojnar@gmail.com>
+
+# The master version of this file is maintained here:
+#
+#   https://github.com/tectonic-typesetting/tectonic/blob/master/dist/arch/PKGBUILD.in
+#
+# The version on aur.archlinux.org is updated automatically by the Tectonic
+# CI/CD system # when new versions are released.
+
+pkgname=tectonic
+pkgver=0.3.3
+pkgrel=0
+arch=('x86_64')
+pkgdesc='Modernized, complete, self-contained TeX/LaTeX engine, powered by XeTeX and TeXLive'
+url=https://tectonic-typesetting.github.io/
+license=('MIT')
+depends=('fontconfig' 'harfbuzz-icu' 'openssl')
+makedepends=('rust')
+source=("$pkgname-$pkgver.tar.gz::https://downloader.disk.yandex.ru/disk/47cc21dcf1946440ca098f101f8c61ea6b77c3f37439e6f1ac992ab1ac4e86b8/5fd6ce61/jHPLzTQRj2952-DqNlFxRk9eBLuCYStdc4wyNjIVb3eh77r6YG0uMG2zGEVJ99kyhUP1VzhfqYavpZ6qmPZg3Q%3D%3D?uid=0&filename=tectonic.tar.gz&disposition=attachment&hash=16VTQ6nekxWAR4TtQQd8LgvUv9uvZSm3pJxeOT1faC4f8OpzJEw/mTJnz6lHQj5%2Bq/J6bpmRyOJonT3VoXnDag%3D%3D&limit=0&content_type=application%2Fx-gzip&owner_uid=441928204&fsize=2962135&hid=a66f4bc3cc29805051e529361a8b475a&media_type=compressed&tknv=v2")
+sha512sums=('d555604788ec10c6db21f9cd346e7a52cdc2f044ffffdaf3bc08451f9fb269fe01be5cbd64835897421300e013edb97583aac44798df721f64761dbc7d8e0b88')
+
+package() {
+	install -Dm755 usr/bin/tectonic "$pkgdir"/usr/bin/tectonic
+	install -Dm644 usr/share/licenses/"$pkgname"/LICENSE "$pkgdir"/usr/share/licenses/$pkgname/LICENSE
+}


snowinmars:
#700

dist/arch/PKGBUILD.bin
```diff
@@ -0,0 +1,27 @@
...
+pkgname=tectonic
+pkgver=0.3.3
+pkgrel=0
+arch=('x86_64')
+pkgdesc='Modernized, complete, self-contained TeX/LaTeX engine, powered by XeTeX and TeXLive'
+url=https://tectonic-typesetting.github.io/
+license=('MIT')
+depends=('fontconfig' 'harfbuzz-icu' 'openssl')
+makedepends=('rust')
+source=("$pkgname-$pkgver.tar.gz::https://downloader.disk.yandex.ru/disk/47cc21dcf1946440ca098f101f8c61ea6b77c3f37439e6f1ac992ab1ac4e86b8/5fd6ce61/jHPLzTQRj2952-DqNlFxRk9eBLuCYStdc4wyNjIVb3eh77r6YG0uMG2zGEVJ99kyhUP1VzhfqYavpZ6qmPZg3Q%3D%3D?uid=0&filename=tectonic.tar.gz&disposition=attachment&hash=16VTQ6nekxWAR4TtQQd8LgvUv9uvZSm3pJxeOT1faC4f8OpzJEw/mTJnz6lHQj5%2Bq/J6bpmRyOJonT3VoXnDag%3D%3D&limit=0&content_type=application%2Fx-gzip&owner_uid=441928204&fsize=2962135&hid=a66f4bc3cc29805051e529361a8b475a&media_type=compressed&tknv=v2")
```
snowinmars, line 21: This link should be updated. The archive is the result of `makepkg -s && cd pkg && tar -zcvf tectonic.tar.gz .` command.

Ping me if something will go wrong.

dist/arch/PKGBUILD.bin
```diff
@@ -0,0 +1,27 @@
+# Maintainer: tectonic-deploy <sasha+tectonic@hackafe.net>
```
pkgw, line 1: This file should be named something like `PKGBUILD.bin.in`.

dist/arch/PKGBUILD.bin
```diff
@@ -0,0 +1,27 @@
...
+# Maintainer: Daniel M. Capella <polyzen@archlinux.org>
+# Contributor: Jan Tojnar <jtojnar@gmail.com>
+
+# The master version of this file is maintained here:
+#
+#   https://github.com/tectonic-typesetting/tectonic/blob/master/dist/arch/PKGBUILD.in
```
pkgw, line 7: This link needs updating.

dist/arch/PKGBUILD.bin
```diff
@@ -0,0 +1,27 @@
...
+#
+# The version on aur.archlinux.org is updated automatically by the Tectonic
+# CI/CD system # when new versions are released.
+
+pkgname=tectonic
+pkgver=0.3.3
```
pkgw, line 13: The version should be a replaceable template `@version@`, as in the `PKGBUILD.in` file, since we'll need to update this file during the release processing stage.

dist/arch/PKGBUILD.bin
```diff
@@ -0,0 +1,27 @@
...
+pkgrel=0
+arch=('x86_64')
+pkgdesc='Modernized, complete, self-contained TeX/LaTeX engine, powered by XeTeX and TeXLive'
```
pkgw, line 16: The description should probably mention that this is the pre-built binary package variant.

dist/arch/PKGBUILD.bin
pkgw: I think this part needs to follow [the example of dust](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=dust-bin) and use the official Linux/MUSL release binary as the "source" package. That way we are linking into the existing release automation infrastructure.

dist/arch/PKGBUILD.bin
```diff
@@ -0,0 +1,27 @@
...
+url=https://tectonic-typesetting.github.io/
+license=('MIT')
+depends=('fontconfig' 'harfbuzz-icu' 'openssl')
```
pkgw, line 19: If this package is created from the official MUSL builds (see below), then it will have no dependencies because these libraries will all be statically linked.

pkgw (changes requested):
Thanks for submitting this. I think I get the basic idea of how a binary Arch package should work, and from that I think there are some changes that need to be made here.

The main thing is that this needs to be integrated into our release automation pipeline, specifically the [Arch linux deployment stage](https://github.com/tectonic-typesetting/tectonic/blob/master/dist/azure-deployment.yml#L163) which invokes [the arch/deploy.sh script](https://github.com/tectonic-typesetting/tectonic/blob/master/dist/arch/deploy.sh). These would need to fill in a template and push an update to AUR using similar techniques as the existing non-binary package.

Do you have a link to any documentation about the Arch Linux policy for binary packages? Usually it's discouraged to package prebuilt binaries unless absolutely necessary, and Tectonic should be pretty easy to build in-situ.

pkgw:
Also, this branch should be updated against the latest Tectonic `master` to fix the Windows/pkgconfig build failure.

Markus00000:
> Do you have a link to any documentation about the Arch Linux policy for binary packages?

From [AUR submission guidelines](https://wiki.archlinux.org/index.php/AUR_submission_guidelines):

*Packages that use prebuilt deliverables, when the sources are available, must use the -bin suffix. An exception to this is with Java. The AUR should not contain the binary tarball created by makepkg, nor should it contain the filelist.*

> Usually it's discouraged to package prebuilt binaries unless absolutely necessary

Note [this related change](https://wiki.archlinux.org/index.php?title=Arch_User_Repository&diff=572792&oldid=570351), which removed *Submitting binaries should be avoided.* If I’m not mistaken, there are 2311 `-bin` packages in the AUR at the moment. Here is [the list](https://aur.archlinux.org/packages/?O=0&SeB=n&K=-bin&outdated=&SB=n&SO=a&PP=50&do_Search=Go) with a few false positives. Many aren’t absolutely necessary. My impression is that avoiding long compile times (`thunderbird-beta-bin`) or large build dependencies (`rust` is 491MB) are common reasons.

Also consider this part of another rule:

*If more than a few people would find this package useful, it is appropriate for submission.*

Users have asked for a `-bin` package on GitHub and in the [AUR comments](https://aur.archlinux.org/packages/tectonic/). As [new Arch users discover tectonic](https://old.reddit.com/r/archlinux/comments/kerj38/why_are_the_latex_core_utilities_so_huge/gg4chm3/), I’d guess *more than a few people* has already been achieved. Coincidentally that post exists because the typical TeX Live installation (`texlive-most`) takes about 2GB of disk space. A fourth of that would be required to compile `tectonic`.

Finally, I’d consider the relatively high frequency of tectonic releases.

pkgw:
@Markus00000 Thanks, that's all very helpful.

Once again, I'm a bit unfamiliar with how Arch works ... but now that I think about it, probably Step 1 here should be for someone to create a good `tectonic-bin` package on AUR, independently of anything in this repo? Then when that's shaken down, we can update Tectonic's release automation to update that package along with the non-bin one.

snowinmars:
Just saw: https://aur.archlinux.org/packages/tectonic-bin/

pkgw:
OK, with the new binary package that seems to have been created by @lmartinez-mirror, I'm going to close this pull request because it's been superseded by that work. If @lmartinez-mirror would be amenable, we could work on adding infrastructure to automatically update the binary package upon release, in the same way as we automatically update the non-binary package, as I've described above.

Thanks for your contribution!

PR #709, opened 2021-01-01
Add support for aarch64-apple-darwin building with vcpkg dep backend

mcgoo:
This adds support for building on an `aarch64-apple-darwin` machine using dependencies from `vcpkg`.

It updates `vcpkg` to the most recent commit - the `vcpkg` tree has pretty recently fixed building `graphite2` for `arm64-osx`. (I'm pretty sure the fix is that`vcpkg` now uses `cmake` 3.19 which sets `CMAKE_SYSTEM_PROCESSOR` correctly.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,10 +88,11 @@ tokio = "0.1.22"
 
 [package.metadata.vcpkg]
 git = "https://github.com/microsoft/vcpkg"
-rev = "b7056e9f1f34f18b6648b1f1d9c4e9d31f04e91c"
+rev = "50ea8c0ab7aca3bb9245bba7fc877ad2f2a4464c"
 
 [package.metadata.vcpkg.target]
 x86_64-apple-darwin = { install = ["freetype","harfbuzz[icu,graphite2]"] }
+aarch64-apple-darwin = { triplet = "arm64-osx", install = ["freetype","harfbuzz[icu,graphite2]"] }
 x86_64-unknown-linux-gnu = { install = ["fontconfig","freetype","harfbuzz[icu,graphite2]"] }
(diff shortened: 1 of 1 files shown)


pkgw:
Awesome! The Windows vcpkg build failure looks spurious.

I'd like to get this building in the Azure Pipelines CI suite before merging. I'll research how that might happen.

pkgw:
@mcgoo I think that in order for this to be supported in the CI, the vcpkg build would need to cross-compile, creating binaries for the `aarch64-apple-darwin` target while running on the `x86_64-apple-darwin` target. Does `cargo vcpkg` support that?

mcgoo:
I tried it for cross testing like you are suggesting but no go. (I don't know of any CI with M1 hosts yet.)

I think it probably will work in general because 64 bit host to 32 bit target builds work on Windows. It looks like it will work for tectonic, but `vcpkg install --triplet arm64-osx graphite2` cross build on x64 mac fails in the way that it failed on an arm64 machine running cmake 3.18. (The same issue with `CMAKE_SYSTEM_PROCESSOR` being wrong, causing `-msse2` to be set.) 

I think the fix in cmake 3.19.2 hosted on the arm64 machine is that 3.19 is a universal binary so it runs now arm64 cmake on arm64 hosts and you get `CMAKE_SYSTEM_PROCESSOR` set to arm-something. It seems like running on an intel machine results in `CMAKE_SYSTEM_PROCESSOR`  being wrong. 

I think cmake will get fixed to set the target cpu correctly, at which point
```
cargo vcpkg build --target aarch64-apple-darwin
cargo test --target aarch64-apple-darwin
```
should work.

pkgw:
Yeah, it looks like it will take a little while for the all of the pieces to come together. So I'll merge this and let's keep an eye on the CI / cross-compile situation.

PR #716, opened 2021-01-05
Start working on breaking the code into sub-crates

pkgw:
We're getting serious about this now. The main goal is to get to the point where the Rust <=> C bridge can be separated into its own crate.

# API Breaks

While this change aims to preserve API compatibility, the break into sub-crate means that we have to start using a boxed error type for our errors, because we need to abstract across various library error types that aren't yet defined. This means that there is inevitable churn in the APIs.

- New crate `tectonic_errors` provides a boxed error type based on `anyhow`, and new macros for error context such as `atry!`
  - `tectonic::errors` still provides its `error_chain` error type as migration occurs
  - "New" `tectonic_errors::Error` errors can be converted into `tectonic::errors::Error` errors with `.into()`
  - "Old" errors must be wrapped in a new type, `tectonic::errors::SyncError`, since `anyhow` requires that errors implement `Sync`. Use: `SyncError::new(old_err).into()`.
  -  New extension trait `tectonic::errors::ChainErrCompatExt` helps provide compatibility for `chain_err` calls on Result types using the new error types
- New crate `tectonic_status_base` provides most of the base status-reporting types that originated in `tectonic::status`.
  - `tectonic::status` now reexports these types, which means that their error types change from old to new
  - `NoopStatusBackend` has its `::new()` constructor replaced with `Default::default()`, since the former is totally superfluous.
  - Backtrace printouts removed from error-reporting implementations since the APIs are not stable
- New crate `tectonic_io_base` now provides most of the base I/O types that originated in `tectonic::I`.
  - `tectonic::io` now reexports these type, with new error types
  - `tectonic::digest` now reexports `tectonic_io_base::digest`
  - `tectonic::io::filesystem` moved to `tectonic_io_base::filesystem`
  - `tectonic::io::stack` moved to `tectonic_io_base::stack`
  - `tectonic::io::stdstreams` moved to `tectonic_io_base::stdstreams`


pkgw:
CI: Windows timed out again (argh) but it has succeeded before.

PR #718, opened 2021-01-06
Support CARGO_TARGET_DIR in PKGBUILD

bi4k8:
The src PKGBUILD previously forced a target directory of `./target`, presumably so that installation could copy the `tectonic` binary from `target`. It's preferable to let the target directory vary with `CARGO_TARGET_DIR` and then reach into `CARGO_TARGET_DIR` if defined when installing, because a shared target directory enables less cache duplication across rust programs and can decrease build time if dependencies are already cached.

### DIFF
--- a/dist/arch/PKGBUILD.src.in
+++ b/dist/arch/PKGBUILD.src.in
@@ -24,16 +24,16 @@ sha512sums=('@source_sha512@')
 
 build() {
 	cd $pkgname-$pkgver
-	cargo build --release --target-dir "./target"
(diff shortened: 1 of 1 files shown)


pkgw:
Thanks! Has this change been validated with an Arch build in some fashion? Our CI won't test to that depth.

The CI failure here is an timeout that we sometimes hit.

pkgw:
Well, this seems straightforward enough, so I'll go ahead and merge.

PR #734, opened 2021-01-12
Add basic watching ability to v2 cli

### LINKED ISSUE #719: Add a watch mode
It would be nice to have the ability to run something like `tectonic -X build --watch` with maybe `-w` for short.

I'd expect this to watch all input files (e.g. working out those from `\input`s) used in compilation and then rebuild when those change.

Probably fits more with the new cargo-like cli.

jeffa5:
Fixes #719

This adds the `watch` subcommand, based off of `cargo-watch`. This design should allow for future addition of subcommands such as `check` or `format` which may run after or before a build stage or in place of one. It is very dumb at the moment and just watches all files in the current directory, besides the output `build` dir.

Tested that it works for me on a setup from `tectonic -X new`.

Haven't really got a good idea of how to go about testing this kind of thing but most of the work is done by the `watchexec` crate anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -134,6 +134,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "473fc6b38233f9af7baa94fb5852dca389e3d95b8e21c8e3719301462c5d9faf"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -230,6 +239,7 @@ dependencies = [
  "atty",
  "bitflags",
  "strsim 0.8.0",
+ "term_size",
  "textwrap",
  "unicode-width",
  "vec_map",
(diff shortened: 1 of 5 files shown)


pkgw:
Thanks for submitting this! I am a little wary about using `watchexec` for this functionality, since it adds some potential failure points (needing to be able to re-execute the program under the name `tectonic`, for instance), and we can in principle do a much better job — for instance, the "makefile rules" functionality yields a precise list of the files that need to be watched to know when to rebuild.

I'm a bit tied up for the rest of this week but I'll try to review this in more depth soon. `watchexec` might be a fine choice in terms of getting the functionality without needing to put in a lot of effort.

jeffa5:
I've updated this to use the current executable name (prettified) for the command now so it also works when not running `tectonic`.

I think `watchexec` is pretty useful for this as it allows the flexibility of using multiple commands, adhering to ignore files and provides this cross-platform.

I'd expect that the list of files to watch from the Makefile rules would be able to be used as included paths to `watchexec` later.

pkgw:
Hi, I just wanted to apologize for the slow follow-up here — I've had a lot going on preventing me from spending time on Tectonic lately. But I haven't forgotten about this!

pkgw:
I'd be interested in a fancier implementation that avoids re-execing the Tectonic executable, but let's merge this now and get the functionality it. The implementation can always be reworked later if/when needed.

PR #739, opened 2021-01-18
Break out the core "bridge" functionality and port bibtex to use it

pkgw:
At long last, let's try to see if we can get the engines really modularized.

## Draft status

- [x] Create new crates
- [x] Integrate them into main build


pkgw:
The `tectonic-on-arXiv` changes appear to be improvements related to TeX path normalization now being applied in more situations.

Issue #745, opened 2021-02-08
Tectonic.toml multiple outputs for the same document


jeffa5:
I assume there are already ideas for the multiple outputs section in `Tectonic.toml` but I thought it would be useful to maybe begin outlining some features or set some expectations.

One thing I'd like is the ability to specify the preamble, index, and postamble file locations in this config file. Then we'd be able to have one build manifest for different versions of a document, e.g. a draft version and a final one. If these aren't specified then they should just default to what they are now.

e.g.
```toml
[[output]]
name = "draft"
type = "pdf"
preamble = "_preamble_draft.tex"

[[output]]
name = "final"
type = "pdf"
preamble = "_preamble_final.tex"
```

I'm sure there are other situations when specifying these would come in useful. I'd be interested to hear any thoughts.

pkgw:
Yup, that is exactly one of the things that I have envisioned!

The existing infrastructure is *really* just a sketch of what's possible, but the intention is that it shouldn't require anything very fancy in order to build out the functionality of the CLI commands. I would very much welcome assistance in doing so.

PR #746, opened 2021-02-09
Tectonic.toml: preamble, index and postamble

jeffa5:
This enables the use of preamble, index and postamble keys in an output.
These get built together to form the input buffer and default to the
current defaults too.

There isn't any fancy checking here as we defer it to the latex engine.
In theory this could enable injection but I'm not aware that would be a
concern at the moment.

Fixes https://github.com/tectonic-typesetting/tectonic/issues/745

### DIFF
--- a/docs/src/ref/documents.md
+++ b/docs/src/ref/documents.md
@@ -23,9 +23,10 @@ multiple documents inside a single workspace.
 [workspace]: ./workspaces.md
 
 The TeX sources are stored in a `src` subdirectory of the document root. This
-directory should contain at least three files: `index.tex`, `_preamble.tex`, and
-`_postamble.tex`. The [`build` command][cli-build] will process these
-files in the following order:
+directory should contain at least three files: `index.tex`, `_preamble.tex`,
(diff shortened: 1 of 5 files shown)


pkgw:
Thank you! At first blush this looks good, and I really appreciate you adding some test coverage.

Would you be able to go even a step farther and document the new settings in the reference? `docs/src/ref/tectonic-toml.md` would be the file.

I'm pretty oversubscribed this week so it might be a bit before I can do a full review. Please ping me if it's been a little while.

jeffa5:
Thanks for taking a quick look.

I've now added some documentation, happy for comments :)

pkgw:
Sorry, this is taking me even longer than expected to get to. If anyone else wants to do a code review they are more than welcome to offer their opinion!

pkgw:
After such a long wait ... this looks great! Thank you! My apologies for taking so long to review and merge this .. if it's any consolation, it's not like I've been working on anything else Tectonic-related in the meantime ...

Issue #743, opened 2021-01-31
Github Action for Tectonic (with biber/cache support)


WtfJoke:
Hi there,

I recently created a Github Action to compile tex files using tectonic, see https://github.com/wtfjoke/setup-tectonic
I did that, because I wasnt happy enough with all the existing (docker) actions.
It has some major advantages to all existing ones:

| Pro                       | Description                                                               |
| --------------------------|:--------------------------------------------------------------------------|
| :zap: Performance         | - Supports caching </br> - Native Javascript Action's are faster than docker (:whale:) based actions  |
| :robot:  Future proofed   | New tectonic versions right on release without code changes                                 |

The action downloads tectonic right from the [release page here](https://github.com/tectonic-typesetting/tectonic/releases). This means I dont need to update any code that users can enjoy the newest tectonic version. 

Tectonic runs directly on the github runner,  docker is not involved in any moment (this also means it supports linux, windows and mac). 
The great thing about that, is that its possible to cache the downloaded packages through actions like [actions/cache](https://github.com/actions/cache), which speeds up the execution even more :)

It also optionally downloads [biber right from sourceforge](https://sourceforge.net/projects/biblatex-biber/files/biblatex-biber/)

I was hoping it could be mentioned in the readme/website, so that more people know about it. 
If you wish I can create PR for the change.

pkgw:
Hi, sure, I'm happy to add a mention of it. I don't feel qualified to judge the different merits of different implementations here (I haven't used GitHub Actions), but I think it is good if we can provide a menu of options. If you could create one or more PRs that would be great.

Thanks!

WtfJoke:
Thank you for the fast merge

PR #750, opened 2021-02-14
Add setup-tectonic github action reference

WtfJoke:
As discussed in https://github.com/tectonic-typesetting/tectonic/issues/743 will add a reference to the setup tectonic github action


pkgw:
Thanks! For completeness I'm rerunning the Windows CI step, which intermittently times out on Azure Pipelines.

Issue #749, opened 2021-02-12
Test failure in i686


ericonr:
Hi, as part of packaging `tectonic` for Void Linux, we run testuites for the package, which led us to notice an issue with the testuite on i686 (glibc, so nothing out of the ordinary):

```
Running target/i686-unknown-linux-gnu/release/deps/tex_outputs-8a48d0397862c30c

running 20 tests
test a4paper ... ok
test md5_of_hello ... ok
test file_encoding ... ok
test pdfimages ... ok
test pdfoutput ... ok
test prim_creationdate ... ok
test prim_filedump ... ok
test prim_filemoddate ... ok
test prim_filesize ... ok
test issue393_ungetc ... ok
test redbox_png ... ok
test synctex ... ok
test tectoniccodatokens_errinside ... ok
test otf_basic ... FAILED
test tectoniccodatokens_noend ... ok
test tectoniccodatokens_ok ... ok
test tex_logo ... ok
test the_letter_a ... ok
test negative_roman_numeral ... ok
test unicode_file_name ... ok

failures:

---- otf_basic stdout ----
thread 'otf_basic' panicked at 'difference in otf_basic.xdv; contents saved to disk', tests/util/mod.rs:198:9
stack backtrace:
   0: rust_begin_unwind
   1: std::panicking::begin_panic_fmt
   2: tex_outputs::util::ExpectedInfo::test_data
   3: tex_outputs::util::ExpectedInfo::test_from_collection
   4: tex_outputs::TestCase::go
   5: core::ops::function::FnOnce::call_once
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    otf_basic

test result: FAILED. 19 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

This happens only with i686, x86_64 is fine. Could be i686 only, or could be 32-bit related.

pkgw:
Thank you for reporting. It's not obvious to me what's happening here but I wouldn't be shocked if something 32-bit-related has snuck in.

A good first step would be adding a CI build that reproduces the problem. This can be a bit annoying to do for Tectonic since we rely on external libraries, so you can't use a bare-bones cross-compilation environment, but for a platform like i686 there shouldn't be anything tricky to do, just a matter of getting the pieces together.

pkgw:
Using [dviasm](https://github.com/aminophen/dviasm), I can get an analysis of the differences between the expected XDV file and the observed result on i686.

[...]

PR #758, opened 2021-03-20
Get an i686 build going

pkgw:
Re #749. The first commit adds i686 to the test framework, so the CI should fail due to the issue being reported in #749. That's what I get locally as well, at least.


pkgw:
The new i686 test is failing in the "right" place: in the `otf_basic` test after successfully building the code.

pkgw:
i686 is working now! Current CI failures are something that seems to have started spontaneously, unrelated to this PR.

pkgw:
Awwww *heck yeah* waiting actually worked!

Issue #754, opened 2021-02-18
Hard-to-interpret error if a directory/file named "latex" exists in the source directory


manfredlotz:
This is Fedora 33 and tectonic 0.4.1

I run `tectonic x.tex` and get

```
note: this is a BETA release; ask questions and report bugs at https://tectonic.newton.cx/
Running TeX ...
warning: 4-byte read failed
caused by: Is a directory (os error 21)
error: something bad happened inside TeX; its output follows:

===============================================================================
===============================================================================
error: the TeX engine had an unrecoverable error
caused by: could not undump 1 4-byte item(s) from latex
```

When I move the file into a different directory and run `tectonic x.tex` no error at all, i.e. the pdf gets built.

I have no idea what's going on.

pkgw:
So if I'm understanding correctly, the build succeeds in one directory, but fails in another? Maybe the bad directory is read-only or something?

Would you be comfortable obtaining an attaching an `strace` log to the failing run? That might help in this case.

manfredlotz:
It was definitely not a r/o file system. Now, it is working again. No idea what happened. Next time I will run `strace` rightaway.

Nevertheless, do you have any idea what those error messages tell me?

dbuscher:
I have the exact same problem, but on OS-X/brew (did not realise that moving the files to a new directory would work but it does!). Is there some equivalent of strace I can run?

dbuscher:
OK, I have worked out the immediate cause: if I remove the directory called "latex" from the working directory, it works, creating a new directory called "latex" causes it to fail. Latex file as follows:

\documentclass{article}
\begin{document}
This is a test
\end{document}

manfredlotz:
Yep, **good catch**. Perhaps I had a directory `latex` where things did not work some days ago. 

At least, when creating a directory `latex` I get the error which caused me to open the issue.

ralismark:
Looks like this issue has been fixed -- this could probably be closed now.

manfredlotz:
When I clone the repository and build tectonic in master I see the same error as almost 2 months ago. What could I do to verify that it has been fixed?

pkgw:
@manfredlotz The issue identified here should be fixed in master, yes. To confirm that you're running the latest code, I guess you could try creating a minimal dummy document and verifying that it builds (1) as a minumum working example and (2) when you create a `latex` directory to trigger the bug that I believe should be fixed?

[...]

PR #759, opened 2021-03-27
Various updates

pkgw:
This PR addresses #754 and also touches things up for the new Rust release.


pkgw:
The tectonic-on-arXiv changes here seem to all be due to the recent rounding fix (#758). It would be nice to be able to check my solution to #754 separately but I don't see any new build failures and I'm inclined to say that's good enough for me.

PR #762, opened 2021-04-03
Use absolute paths in synctex

### LINKED ISSUE #720: Reverse search fails: paths in .synctex.gz should be absolute
I was sent to report to you from https://github.com/James-Yu/LaTeX-Workshop/issues/2467. Reverse search fails for VS-code with the LaTeX-Workshop extension , and the problem appears to be that the path created in in .synctex.gz is not absolute.

### LINKED ISSUE #744: synctex doesn't handle \input without .tex extension
If I `\input{foo}` (rather than `\input{foo.tex}`), `tectonic --synctex` generates a file with `foo` as the filename rather than `foo.tex`. This breaks the synctex-compatible viewers I've tried.

```shell
$ cat test.tex
\documentclass{article}
\begin{document}
\input{test-input}
\end{document}
$ cat test-input.tex
Hello world!
$ pdflatex -synctex=1 test.tex > /dev/null; gunzip -c test.synctex.gz  | grep test-input
Input:6:/home/zjn/git/spectrum-paper/./test-input.tex
$ tectonic --synctex test.tex > /dev/null; gunzip -c test.synctex.gz | grep input
Input:5:test-input
```

Thanks for Tectonic; I use it for every project I can!

hulloanson:
Fixes #720.

Edit: I haven't fixed #744


pkgw:
Thank you for your contribution @hulloanson ! I imagine that you probably had to navigate around my rearrangements of the source tree, so thanks for your patience with that.

It looks like Azure Pipelines has expunged the logs from its initial run here, so I'm going to close and reopen this PR to rerun the CI and see where we stand.

hulloanson:
It looks like using absolute paths for synctex breaks its tests as the original one expects relative path. Here I am at a lost though: how do we test absolute paths?

pkgw:
@hulloanson That will be a bit tricky, yes. The first thing I can think of is that we could add a layer to the test program that postprocesses the `synctex.gz` file created by the engine and rewrites it to replace the current absolute path prefix with some invariant string such as `$ROOT`. Then we can still use the existing machinery to compare the output files on a byte-for-byte basis.

pkgw:
I have an idea for how to approach this that's a bit more complex, but it ought to be a bit more robust and address #744 as well. I'll push some updates to this branch sometime this week, hopefully.

pkgw:
@coxackie @znewman01 @hulloanson — I've pushed an update to this branch that I think should correct the SyncTeX output. Are any of you in a position to test the code in this PR and see if it seems to be doing the right thing?

znewman01:
I don't have a test case for #720, sorry. (Still happy to test for #744).

pkgw:
@znewman01 I think #744 is a superset of #720 anyway. Just general clicking-around with a SyncTeX-enabled viewer should suffice — I don't have one handy.

znewman01:
ACK I can try it sometime in the next day or so then

pkgw:
Thanks! As per the comment in #744, and my looking at the data, I think this PR seems to doing the right thing. Thanks for kicking this off @hulloanson !

aterenin:
Any chance we can get a release with this in?

pkgw:
Yes, this will happen soon. I want to make a release featuring the new V2 CLI commands added in #786, but there's just one more security-related API tweak I want to make before publishing the new crates.

PR #763, opened 2021-04-03
dist/azure-coverage.yml: oh hey, let's unbreak coverage

pkgw:
Trying to rework the arguments that we use so that the coverage analysis will be more robust to rearrangements of the source tree.


pkgw:
OK, it was fun while it lasted, but it looks like we're back to getting correct coverage stats.

PR #770, opened 2021-04-27
Fix build with watchexec 0.15

pkgw:
The new watchexec changed its error type, I think, and our build broke. Require the new version, track changes to deprecated APIs, and work around the change.

Also update our deps in Cargo.lock in general.


pkgw:
Sorry, Codecov.

Issue #747, opened 2021-02-10
If `shell-escape` is implemented one day, `no-shell-escape` should be also


kpym:
We can see [in the code](https://github.com/tectonic-typesetting/tectonic/blob/045dbb5350e6bdb33bbf89d01942add49a1484d4/src/unstable_opts.rs#L24-L25) that may be one day `shell-escape` will be implemented. 

I put this issue here to say that `shell-escape` goes together with `no-shell-escape` and if one is implemented, the other should be also.

pkgw:
I can say with high confidence that the default setting will always be to have shell-escape disabled. In that context, do you think that there are use cases where an inverted `no-shell-escape` flag would be useful?

kpym:
IMO, the situation is similar as other tex compilers : we should specify `shell-escape` to use `\write18`, but if we use `no-shell-escape` even with `shell-escape` we can't have system access. In this way if you have in a script `xelatex -no-shell-escale $otherarguments` you are sure that even if `$otherarguments` contains `-shell-escape` the program will stay "secure".

pkgw:
That makes sense. And `-Z shell-escape` is now implemented as of 0.5.0.

pkgw:
Actually I think this should be generalized a bit. In #785 I added a `disable_insecure()` method to the `ProcessingSessionBuilder`, with intended semantics that any kind of insecure feature will be prevented from subsequent activation. We could provide a `--disable-insecure` argument that would not only disable shell-escape, but also turn off any other known-insecure behaviors that might be added in the future.

You could imagine some kind of baroque use case where maybe someone wants to disable Insecure Feature X but *does* want to enable shell-escape, but I think the blanket approach is likely to be what people are seeking 99% of the time.

kpym:
I agree that `-no-shell-escape` is here only for security reasons, so it could be included in `--disable-insecure` without problem. The reasons, which come to my mind, for having both are:
- I have no idea what the other insecure features are, but may be someone would like to disable `shell-escape` and to enable one of this features.
- If `--disable-insecure` is just a shortcut to other `--disable-` and `--no-` flags, it will be clear what `tectonic` consider as "insecure".
- If `--disable-insecure` is a shortcut to other flags, probably the testing will be simpler, because you could test all this flags separately. 
- `--disable-insecure` can always be introduced later, when more than one insecure flags are present, as shortcut.

pkgw:
[...]

PR #787, opened 2021-06-14
Add a new centralized security API

pkgw:
In the wake of adding shell-escape support, I felt like I wanted to offer a more coherent story for users to be able to control security settings. We do this with some new security settings APIs, changing the default behavior to "untrusted" mode where known problematic features are disabled.

On the user front, in the regular `tectonic` program, insecure features will still be disabled by default. You can enable, say, shell-escape with `-Z shell-escape` as before. But now there's a `--untrusted` option that overrides any other settings and prevents shell-escape from being turned on (cf. #747).

Furthermore, the security APIs are set up in a centralized fashion such that the environment variable `$TECTONIC_UNTRUSTED_MODE` will always be honored, with the same effect as `--untrusted`.

Note that in a certain sense it's not great to have `--untrusted` be an extra option, not the default, but from a usability standpoint people are going to be spending most of their time building trusted documents.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -125,7 +125,7 @@ x86_64-unknown-linux-gnu = { install = ["fontconfig","freetype","harfbuzz[icu,gr
 x86_64-pc-windows-msvc = { triplet = "x64-windows-static", install = ["fontconfig","freetype","harfbuzz[icu,graphite2]"] }
 
 [package.metadata.internal_dep_versions]
-tectonic_bridge_core = "4e16bf963700aae59772a6fb223981ceaa9b5f57"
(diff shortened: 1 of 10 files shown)


ratmice:
Nice I like the sound of --untrusted a lot better than the double negative in --disable-insecure, but didn't mention for fear of bikeshedding and having not looked if --disable-x is a general convention, suffice to say I like `--untrusted` a lot better.

pkgw:
Tectonic-on-Arxiv changes are more expected Synctex changes.

@ratmice Yeah, I was a little dissatisfied with the terminology until I landed on that one. I try to be careful about this stuff!

PR #794, opened 2021-06-19
Add biber support

pkgw:
Let's do this! I think it should work.


pkgw:
The Arxiv changes are related to the SyncTeX updates. @Mrmaxmeier would it be possible to update the reference results to catch them up with the SyncTeX fixes?

Mrmaxmeier:
> The Arxiv changes are related to the SyncTeX updates. @Mrmaxmeier would it be possible to update the reference results to catch them up with the SyncTeX fixes?

Those should be picked up automatically. The bot compares between the head and base of each pull request. :slightly_smiling_face:

pkgw:
@Mrmaxmeier That's what I thought, but the SyncTeX fixes have been in place for a little while. Are the builds done in a temporary directory? The SyncTeX files now embed the absolute paths of the inputs, so if those paths change from one run to another, they'll show up as changes. In the test suite we turn the variable path prefix into `$ROOT` or something like that. (There's also a byte offset that becomes variable if the length of the value of $ROOT can vary.)

Mrmaxmeier:
Yep, builds are run in fresh temp dirs and these paths seem to be the only changes in the SyncTeX output files. I guess an unstable option that switches back to relative paths would be nice (similar to the `SOURCE_DATE_EPOCH` env var, or maybe even a catch-all like `-Z reproducible`).

pkgw:
@Mrmaxmeier Interesting suggestion! Something like a `-Z reproducible` does sound useful, so I'll think about what we could do there. But if your system is able to postprocess the `.synctex` output a little bit, that also ought to work, I think.

PR #802, opened 2021-07-02
Fix the build on Rust 1.46

pkgw:
Needed to keep the conda-forge package building on macOS, which is stuck on an older Rust version due to an (accidental) increase in the minimum required OS version of later releases.

I ought to define a Minimum Supported Rust Version and set up CI to monitor it, but ... not right now.


pkgw:
xref https://github.com/conda-forge/tectonic-feedstock/pull/40

pkgw:
tectonic-on-arXiv failures are known issues with reproducibility in the SyncTeX output.

PR #803, opened 2021-07-03
Don't wrap diagnostic warnings

### LINKED ISSUE #684: option: don't add newline
https://github.com/tectonic-typesetting/tectonic/blob/0f1553e5a82e5ae5bfb7e5986576baaae7c8b490/tectonic/xetex-output.c#L149-L151

As modern terminals can line break long string themselves, I think we should add `-Z` option to disable tectonic do this.

ralismark:
The `print_ln()` calls in the original code would also output newlines to the current diagnostic. Changing it to directly print newlines fixes this issue (though interestingly, the code for the `SELECTOR_TERM_AND_LOG` already did this). This change shouldn't affect the contents of the tex log, only the nicer-looking warnings that get printed to the terminal.

Fixes #684, I think, though that issue might be referring to the actual tex log that's being printed to `rust_stdout`.

### DIFF
--- a/crates/engine_xetex/xetex/xetex-output.c
+++ b/crates/engine_xetex/xetex/xetex-output.c
@@ -139,16 +139,20 @@ print_raw_char(UTF16_code s, bool incr_offset)
         ttstub_output_putc(log_file, s);
         if (incr_offset)
             file_offset++;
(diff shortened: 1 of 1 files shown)


pkgw:
Thanks!

Issue #796, opened 2021-06-21
Biber error: "cannot find _____.bib"


aterenin:
Posting a standalone issue following Biber support added in #35. The following works if `backend=bibtex` option is used but not if it is removed.

MWE.tex
```tex
\documentclass[12pt]{book}

\usepackage{biblatex}
\bibliography{citations}

\begin{document}

\title{MWE}

\maketitle

\textcite{example}

\printbibliography

\end{document}
```

citations.bib
```
@inproceedings{example,
	author = {Example Example},
	booktitle = {Example Example},
	title = {Éxample Example},
	year = {2021}}
```

Output
```
aterenin@MacBook-Pro MWE % tectonic mwe.tex
Running TeX ...
Running external tool biber ...
error: the external tool exited with an error code; its stdout was:

===============================================================================
INFO - This is Biber 2.16
INFO - Logfile is 'mwe.blg'
INFO - Reading 'mwe.bcf'
INFO - Found 1 citekeys in bib section 0
INFO - Processing section 0
INFO - Looking for bibtex format file 'citations.bib' for section 0
ERROR - Cannot find 'citations.bib'!
INFO - ERRORS: 1
===============================================================================
error: its stderr was:

===============================================================================
===============================================================================
error: the external tool exited with error code 2

```

pkgw:
Thanks for reporting! I'll have to do a bit of digging to identify the "right" approach but I anticipate that this should be straightforward to fix.

WtfJoke:
Same problem with this sample repo
https://github.com/WtfJoke/setup-tectonic-sample

zzamboni:
I'm seeing the same error with a document in which I use `\addbibresource`, even when specifying the `backend=biber` option to `biblatex`. My minimum example would be very similar to the repo posted by @WtfJoke.

BTW: I think it's AWESOME that Tectonic now supports biber!

Siborgium:
@zzamboni it seems to work with absolute paths.

zzamboni:
@Siborgium thanks for the tip! I can confirm that using the absolute path of the `.bib` file in `\addbibresource` works.

This is not ideal since I would like the filenames to be relative to the current file and not dependent on where the directory is checked out, but it's good to have a workaround until this bug get fixed :)

pkgw:
[...]

PR #804, opened 2021-07-03
Improve biber handling by parsing the `.run.xml` file

pkgw:
It turns out that in some cases, biber needs access to various `.bib` files that may be associated with the document build. Fortunately, the `.run.xml` file contains exactly the information we need to populate the temporary tool directly properly. We have to link in an XML parser, but so be it.

I wasn't aware of the `.run.xml` file before and we potentially could use it a lot more aggressively. We don't do that here, but the support is written to try to be future-conscious.

Addresses #796.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1654,6 +1654,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "quick-xml"
+version = "0.22.0"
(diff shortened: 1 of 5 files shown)


pkgw:
tectonic-on-arxiv failures are the expected SyncTeX reproducibility issues.

Issue #769, opened 2021-04-23
--hide does work with absolute paths/folders


ralismark:
Passing an absolute path to `--hide` does not stop tectonic from reading that file. If a relative path is used, this still works.

e.g. `tectonic b.tex --hide "$PWD/a.tex"` with a.tex as
```tex
hello world
```
and b.tex as
```tex
\documentclass{article}
\begin{document}
\input{a.tex}
\end{document}
```

`tectonic b.tex --hide a.tex` correctly fails with "File a.tex not found".

The cause of this is that FilesystemIo doesn't convert paths to absolute paths, and so might try to compare an absolute path with a relative one.

A related issue is that `--hide` doesn't work with folders, and requires you to specify each file individually.

pkgw:
Thanks for reporting this. I think the basic issue here is that the `--hide` mechanism operates on "TeX paths" as the TeX engine sees them, rather than filesystem paths of actual files. For relative paths one can fudge the difference, but here it matters.

It would be quite reasonable to try to improve things here. I don't think that I'm going to have the bandwidth to work on this personally but I'd welcome contributions in this area!

PR #806, opened 2021-07-12
Improve path processing in FilesystemIo

ralismark:
This addresses #8 and #769 by adding path checks to `input_open_name_with_abspath`. The hidden path checking is a bit complex, but hopefully it correctly handles edge cases involving symlinks.


pkgw:
Thanks for tackling this! I'm a bit busy this week but I'll try to find the time to sit down and review the approach you're taking.

pkgw:
I'm superseding this PR with #911 since it's been idle for a while. That one will include the absolute path warning but not the changes to hidden-path processing, due to the concerns I gave above.

PR #808, opened 2021-07-13
bridge_harfbuzz: bump harfbuzz to 2.8.2

pkgw:
Finally remove that annoying deprecation warning!


pkgw:
This does cause a change in the 1702.04877 testcase in the Arxiv corpus. It's a relatively long math paper and I can't perceive any important difference, though.

PR #814, opened 2021-08-26
Support custom search directories

### LINKED ISSUE #755: User-configurable search path(s) for .sty files, etc. ("-Z searchpath=..."?)
Hi, is there a recommended way to get Tectonic to locate my own .sty files when building (i.e., environment variable, flag, etc.)? My use case is with a "homework.sty" file that I'll be reusing in several documents in different locations, so copying the .sty multiple times doesn't seem like a great solution. Thanks for reading!

ralismark:
This add the `-Z search-path` option, which make tectonic look in additional paths for files.

Resolves #755.

I'm not that much of a fan of that option name (maybe something like `-Zinclude` is better?), so if you have any suggestions I'm happy to change this PR.

Todo:

- [x] `--hide` support
- [x] untrusted mode support

### DIFF
--- a/crates/bridge_core/src/lib.rs
+++ b/crates/bridge_core/src/lib.rs
@@ -764,6 +764,11 @@ impl SecuritySettings {
     pub fn allow_shell_escape(&self) -> bool {
         !self.disable_insecures
     }
+
+    /// Query whether we're allowed to specify extra paths to read files from.
+    pub fn allow_extra_search_paths(&self) -> bool {
+        !self.disable_insecures
+    }
 }
 
 impl Default for SecuritySettings {
--- a/docs/src/v2cli/compile.md
+++ b/docs/src/v2cli/compile.md
@@ -126,4 +126,4 @@ the set of unstable options is subject to change at any time.
 | `-Z min-crossrefs=<num>` | Equivalent to bibtex's `-min-crossrefs` flag. Default vaue: 2 |
 | `-Z paper-size=<spec>`   | Change the initial paper size. Default: `letter` |
 | `-Z shell-escape`        | Enable `\write18` (unless `--untrusted` has been specified) |
-
+| `-Z search-path=<path>`  | Also look in <path> for files, like TEXINPUTS. Can be specified multiple times |
--- a/docs/src/v2cli/compile.md
+++ b/docs/src/v2cli/compile.md
@@ -18,6 +18,8 @@ tectonic -X compile myfile.tex
 
 #### Usage Synopsis
 
+<!-- Keep options alphabetized -->
+
 ```sh
 tectonic -X compile  # full form
   [--bundle PATH] [-b PATH]
@@ -90,6 +92,8 @@ can trivially defeat this by explicitly clearing the environment variable.
 
 The following are the available flags.
 
+<!-- Keep alphabetized by full name: -->
+
 | Short | Full                      | Explanation                                                                                    |
 |:------|:--------------------------|:-----------------------------------------------------------------------------------------------|
 | `-b`  | `--bundle <PATH>`         | Use this Zip-format bundle file to find resource files instead of the default |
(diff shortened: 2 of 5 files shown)


ralismark:
Fixed those two things. There's a couple more things that might be worth fixing

1. Making this option work with untrusted mode (i.e. by forbidding extra include directories). Currently, we don't have the ability to completely forbid looking up absolute paths so it won't affect anything, but the ability to arbitrarily add search paths seems like the kind of thing we'd want to forbid in a locked-down mode.
2. Making this work with path hiding i.e. `--hide`.

pkgw:
Thank you! I've pushed a tiny tweak to keep the unstable options alphabetized in the docs, and will merge once CI passes (module the arxiv test expected failure).

(Whenever I have a list of things in source that doesn't have any inherent ordering, I try to keep it alphabetized for tidiness.)

ralismark:
The use case for `--untrusted` is that the *input document* is untrusted, but how we invoke Tectonic is trusted?

In that case, it's relatively easy to make it always allowed, without completely stripping out security support:
```diff
diff --git a/crates/bridge_core/src/lib.rs b/crates/bridge_core/src/lib.rs
index 7aa496b0..1eaeacd1 100644
--- a/crates/bridge_core/src/lib.rs
+++ b/crates/bridge_core/src/lib.rs
@@ -767,7 +767,7 @@ impl SecuritySettings {

     /// Query whether we're allowed to specify extra paths to read files from.
     pub fn allow_extra_search_paths(&self) -> bool {
-        !self.disable_insecures
+        true
     }
 }
```

pkgw:
@ralismark That's right. There basically isn't any choice to trusting the user that's invoking the code, because they can alter the environment and set up `LD_PRELOAD` or just compile a new executable with hacked source code, or whatever. Things would be a little different if Tectonic was installed in some kind of multi-user setuid setup or anything, but that's not the case (and it's hard to see how it ever would be).

The Windows vcpkg build has an error that I think must have been introduced with Rust 1.55, which was released today, since the build worked before. I'll merge this and iron that out in another PR.

ralismark:
@pkgw just a note that [what's in master](https://github.com/tectonic-typesetting/tectonic/blob/master/crates/bridge_core/src/lib.rs#L770) *doesn't* have that above patch to allow extra search paths in `--untrusted`.

pkgw:
@ralismark Right. Creating a nice opportunity for someone to get their feet wet with a small pull request :-) (Basically I didn't want to wait for CI to run again, which isn't a very good reason but we're being too careful, not too sloppy, so I don't feel too bad about leaving this loose end dangling.)

ratmice:
Just some questions/observations, about this as well as untrusted now doing two things.

1. search-path is an unstable option because it affects reproducibility?  I.e. a document requiring the option won't build without that option specified, but paths may differ between systems requiring different paths to be set between different systems?
2. if search-path was also a Tectonic.toml thing there suddenly *would* be trusted/untrusted security implications (I could actually use this now with git submodules) this would also seem to make it reproducible assuming only paths relative to the Tectonic.toml were allowed (Which sounds like a job for https://github.com/bytecodealliance/cap-std).

For now i'll just say if 2 becomes a thing the behavior of the `untrusted` bit is somewhat nuanced.
It might be expected that relative paths is allowed in non-hidden directories during untrusted or something.

There is a worry here that untrusted disables both shell-escape and search-paths, when one might want to disable one without the other. It is a bit concerning that `tectonic -Z shell-escape=false foo.tex` enables shell-escape (It is actually ignored entirely, see `-Z shell-escape=foo`, we might want to make that an error (rather than extend shell-escape to accept a bool?).

pkgw:
@ratmice Thanks for your comments! Here's a smattering of responses ...

- I think search-path should be unstable mainly because of its reproducibility implications — not so much that the CLI interface might evolve, but frankly I hope that the "unstable" terminology discourages people from using it a bit. (That being said, we're not like Rust where you have to be on a "nightly" train in order to use unstable options at all, so it's a very low level of discouragement.) I want to prevent the CLI from growing a profusion of command-line options and modes, and I believe the best way to do that is to encode that sort of configuration into `Tectonic.toml`, and I think that the search-path functionality would need to become more sophisticated before exposing it in `Tectonic.toml`. Because ...
- I agree with what you say about the implications of search-path going into Tectonic.toml. The key is that this changes the search path from something the user fully controls, to something that the input document can affect. As you say, it would make a ton of sense to allow this *if* the search path is something relative to the source document directory. (Or the overall multi-document workspace, if/when we flesh out that concept.)
- I tried to design the security model so that it can be a bit more fine-grained than a boolean trusted/untrusted if needed. The vision is that there should always be a global untrusted mode that means "prevent this document from doing anything naughty that we can possibly think of", but if there's a use-case for allowing a subset of potentially-dangerous options, we should make it possible.
- I agree that if we accept `-Z shell-escape=false` with a meaning of "enable shell-escape", that's bad.

So in terms of working on things:

- A fix to prevent surprises of the form  `-Z shell-escape=false` would definitely be welcome
- Addition of support for search-path in `Tectonic.toml` would also be very welcome, with the restriction that in untrusted mode the search path needs to be a subdirectory of the `Tectonic.toml` file. (Or maybe one level of parent directory since we don't support a "workspace" concept yet, but you want to prevent processing of `../../../../../../../../../../etc/shadow`. And note that the default is trusted mode!)

Finally, I'll mention a related but distinct idea. I'd like to eventually add a `reproducible = True` flag in `Tectonic.toml` that would indicate that the document is intended to be fully reproducible. This would, e.g., cause the engine to reject an absolute search-path setting even if the document is trusted. Eventually we'd also reject uses of system fonts not available in the bundle, require that locale-dependent settings like the paper size be specified explicitly, and so on. I think this would be a nice way to promote reproducible documents without being too restrictive. In principle reproducibility and trust are separate issues, but often non-reproducible = system-dependent = access outside of the document sandbox = trust implications.

PR #820, opened 2021-09-10
Miscellaneous maintenance updates

pkgw:
Mostly motivated by the breakage of our Windows/vcpkg CI builds.

- General deps update
- Update vendored Harfbuzz to 2.9.1
- Update to latest vcpkg

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,15 +185,15 @@ dependencies = [
 
 [[package]]
 name = "bytes"
-version = "1.0.1"
+version = "1.1.0"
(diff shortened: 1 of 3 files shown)


pkgw:
The build issue here seems to stem from a change in Rust 1.55 where Cargo is no longer exporting RUSTFLAGS to build.rs scripts, breaking the logic we use to decide whether to define `GRAPHITE2_STATIC`. We should use `CARGO_ENCODED_RUSTFLAGS` instead.

PR #822, opened 2021-09-16
pdf_io: prevent fallthrough when reading EXIF data

### LINKED ISSUE #821: False warning when including JPEGs
While compiling a document containing a JPEG image, I encountered the warning: `JPEG: Invalid data for ResolutionUnit in Exif chunk.` After checking my JPEG, I'm convinced that this warning is false.

After checking the code, I think that [this line](https://github.com/tectonic-typesetting/tectonic/blob/3e6fc3b811e7697056221b118f87bd2395c2ca13/crates/pdf_io/pdf_io/dpx-jpegimage.c#L721) is missing a break, but I'm not sure. Could someone verify this?

korrat:
This unintended fallthrough caused the JPEG decoder to issue a warning
when including a JPEG image.

Fixes #821.


pkgw:
Thank you for your investigation and solution of this issue!

PR #823, opened 2021-09-18
Shell escape arg error

ratmice:
This is a quick hack, to error on shell-escape if an argument is given, currently it ignores any value given via `-Z shell-escape=foo`, which means if you call `-Z shell-escape=false` it is still set to true.

Rather than try and set this up to toggle the value, since that would have backwards compatibility implications, throw an error if a value is set.

I had some troubles trying to figure out here what was needed for the errmsg! macro, and ended up hacking it to just use `Err(format!`, so that might need another iteration.

The test added in the first commit fails with the current master, and should pass by the second commit.

### DIFF
--- a/src/unstable_opts.rs
+++ b/src/unstable_opts.rs
@@ -61,6 +61,18 @@ impl FromStr for UnstableArg {
             })
         };
 
+        let require_no_value = |unwanted_value: Option<&str>, builtin_value: UnstableArg| {
+            if let Some(value) = unwanted_value {
+                Err(format!(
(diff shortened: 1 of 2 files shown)


pkgw:
Thanks! There were some hopefully-transient CI errors; rerunning now.

pkgw:
Hm, OK, the timing of the CI failure was juuust so that I'd have to relaunch everything, but I'm confident that this PR isn't going to introduce any new issues, so I'll just merge it now. Thanks again!

PR #828, opened 2021-10-03
fix: v2 cli watch on windows failed 

### LINKED ISSUE #827: v2 cli watch on windows failed
```
[Running `\scoop\apps\tectonic\current\tectonic.exe -X build`]
At line:1 char:66
+ ... \scoop\apps\tectonic\current\tectonic.exe -X build & echo [F ...
+                                                                 ~
The ampersand (&) character is not allowed. The & operator is reserved for future use; wrap an ampersand in double quot
ation marks ("&") to pass it as part of a string.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : AmpersandNotAllowed
```

On Windows 11 (Build 22468). According to this snippet below it evals `tectonic -X build & echo ...` while `watchexec` under this condition it by default requires `powershell` to run that command.
https://github.com/tectonic-typesetting/tectonic/blob/master/src/bin/tectonic/v2cli.rs#L536

I think it's better explicitly request cmd to be the shell under the circumstance.

Sinofine:
this fixes #827

### DIFF
--- a/src/bin/tectonic/v2cli.rs
+++ b/src/bin/tectonic/v2cli.rs
@@ -530,15 +530,18 @@ impl WatchCommand {
 
         let command = cmds.join(" && ");
 
+        let mut args = watchexec::config::ConfigBuilder::default();
         let mut final_command = command.clone();
         #[cfg(unix)]
         final_command.push_str("; echo [Finished running. Exit status: $?]");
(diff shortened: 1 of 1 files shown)


pkgw:
Thanks!

Issue #776, opened 2021-06-06
0.5.1 AUR package build fails with redbox_png test suite error


roland-rollo:
Hi,

I tried now to upgrade tectonic to version 0.5.0 with paru (Arch Linux AUR package manager) and it throwing me this:
```
   Compiling flate2 v1.0.20
error: failed to run custom build command for `tectonic_bridge_harfbuzz v0.2.0`

Caused by:
  process didn't exit successfully: `/home/roland/.cache/paru/clone/tectonic/src/tectonic-0.5.0/target/release/build/tectonic_bridge_harfbuzz-d360d82d05dc3c96/build-script-build` (exit status: 1)
  --- stderr
  error: no such file /home/roland/.cargo/registry/src/github.com-1ecc6299db9ec823/tectonic_bridge_harfbuzz-0.2.0/harfbuzz/src/harfbuzz.cc
     if you have checked out from Git, you probably need to fetch submodules:
         git submodule update --init
     This is needed because we are attempting to compile a local copy of Harfbuzz.
warning: build failed, waiting for other jobs to finish...
error: build failed
==> ERROR: A failure occurred in build().
    Aborting...
error: failed to build 'tectonic-0.5.0-0':
```
So I thought it's AUR package error, but when I try cargo install tectonic its the same song.
```
   Compiling flate2 v1.0.20
error: failed to run custom build command for `tectonic_bridge_harfbuzz v0.2.0`

Caused by:
  process didn't exit successfully: `/tmp/cargo-installmqIeKw/release/build/tectonic_bridge_harfbuzz-7d943ccf3c5cc12b/build-script-build` (exit status: 1)
  --- stderr
  error: no such file /home/roland/.cargo/registry/src/github.com-1ecc6299db9ec823/tectonic_bridge_harfbuzz-0.2.0/harfbuzz/src/harfbuzz.cc
     if you have checked out from Git, you probably need to fetch submodules:
         git submodule update --init
     This is needed because we are attempting to compile a local copy of Harfbuzz.
warning: build failed, waiting for other jobs to finish...
error: failed to compile `tectonic v0.5.0`, intermediate artifacts can be found at `/tmp/cargo-installmqIeKw`

Caused by:
  build failed
```

pacman -Ss harfbuzz
```
extra/harfbuzz 2.8.1-1 [installed]
    OpenType text shaping engine
extra/harfbuzz-icu 2.8.1-1 [installed]
    OpenType text shaping engine (ICU integration)
```

[...]

PR #847, opened 2021-12-07
engine_xetex: attempt to fix `redbox_png` test issue

pkgw:
We would get results that varied depending on whether we were compiled in debug or release mode. It looks like the root cause is some single-precision math in computing image sizes.

Cf. #845, possibly #776.


wisp3rwind:
This does indeed seem to fix #776 for me.

---
Random thoughts below this line, because I was quite curious on what the issue might be here, but don't know what the compiler is allowed to do:

> /* Tectonic customization: if we use single-precision math, we can
        bounds->wd *= 72.27;
         * sometimes get numerical results that vary depending on whether we're
        bounds->ht *= 72.27;
         * compiled in debug or release mode (?!) */

I also don't really understand this... Might this even be a compiler bug, where on `--release`, `get_image_size_in_inches` is inlined, and on constant-folding `1 / 72 * 72.27` (is constant-folding this even allowed without `fast_math`, given that it changes associativity?) something goes wrong with the `double -> float` type conversions?

pkgw:
I believe your diagnosis is about correct. In release mode, the behavior changed when I added a printf of bounds->wd between the function call and the multiplication. That implies to me that the two statements were being fused in a way that changed their behavior subtly.

On the other hand, with this fix, the *release* mode behavior turns out to have been more correct: the behavior of the *debug* mode changed and I had to update the test files as a result. So I think the inlining actually helped to compiler maintain better precision, while in debug mode the conversion to float must happen "too early" and lose some precision.

PR #870, opened 2022-03-01
Avoid using patched Pinot

pkgw:
OK, based on feedback, I think we have to temporarily stop using the patched Pinot.

The first version of this PR should fail with a new CI test that aims to catch the issue that I encountered.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1770,8 +1770,9 @@ checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
 [[package]]
 name = "pinot"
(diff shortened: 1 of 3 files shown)


pkgw:
xref #869 and https://github.com/rust-lang/cargo/issues/10440

pkgw:
Updates:

- pinot 0.1.4 was kindly released with the math changes, so the whole issue of crate patching is now gone
- It turns out that the `cargo package` step I was hoping to do isn't going to work, because it pulls all dependencies from crates.io during the packaging. So if you add a new API and use it in another crate, the package step won't work. So for now I guess we'll just keep things at the status quo.

Issue #874, opened 2022-03-16
V2 interface is undocumented in `tectonic --help`


clbarnes:
That's the interface the docs recommend, so presumably that's what people should use. But there is no reference to it in the tectonic help text.

pkgw:
That's a good point! I think this should be pretty easy to fix with a little customization to the `clap`/`structopt` usage. Is this something you might be willing to try implementing yourself?

clbarnes:
Resolved by #877

PR #877, opened 2022-03-22
src/bin/tectonic/main.rs: -X option in CLI docs

clbarnes:
See #874 . Hope this is enough!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -120,7 +120,7 @@ tokio = "0.1.22"
 
 [package.metadata.vcpkg]
 git = "https://github.com/microsoft/vcpkg"
-rev = "997cd412fa4dff71bec6c99dea95f0b9a3c2c240"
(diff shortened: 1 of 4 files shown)


pkgw:
Thanks, I like this implementation! The Windows CI failed with what I hope was a transient issue ... if not, I probably need to push an update for our vcpkg integration.

If you're hungry for "extra credit" we could add an `executable` test case for the `-X` option in non-initial position, but I'm definitely happy to merge without.

pkgw:
OK, we may need a vcpkg bump here — I've pushed a commit to your branch.

pkgw:
Thanks!

Looks like the vcpkg update broke the macOS build :-( Please be patient since it might take a few more tries to sort this out. No need for you to do anything here, I just need to figure out how to unstick this piece of the CI system.

PR #884, opened 2022-03-30
Make cache location customizable

### LINKED ISSUE #880: Allow the user to change the cache directory
Currently the cache directory is automatically determined by tectonic, which is nice to get things going quickly but I'm currenlty in a situation (VM environment with Windows) where I'm trying to change the cache directory (because `LOCALAPPDATA` is resetted when the service restarts). I tried to change the environment variable `%LOCALAPPDATA%` but that doesn't seem to work. Probably because the rust crate that provides the directories uses the WinAPI and not the environment variable?

Is there a way to change the directory already and maybe not documented?

Thanks in advance.

wischi-chr:
Make the cache location customizable by allowing the user to overwrite it using an environment variable called `TECTONIC_CACHE_DIR`.

This fixes #880

### DIFF
--- a/crates/bundles/src/cache.rs
+++ b/crates/bundles/src/cache.rs
@@ -19,6 +19,7 @@
 use fs2::FileExt;
 use std::{
     collections::HashMap,
+    env,
     fs::{self, File},
     io::{BufRead, BufReader, Error as IoError, ErrorKind as IoErrorKind, Read, Write},
     path::{Path, PathBuf},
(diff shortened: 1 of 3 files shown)


pkgw:
Thanks for working on this! Your implementation looks great to me.

Would you have the chance to add some supporting pieces for this new feature?

- A test case, hopefully boosting the coverage enough to make Codecov happy
- A mention of this variable in the documentation (`docs/`), although I will fully admit that the docs are very partial and so there might not be an ideal place to add text

wischi-chr:
Sure thing. It might take a little while because I'm AFK the next few days. Should I make separate commits or squash it into a single commit?

pkgw:
@wischi-chr In this case I think I'd prefer squashing everything into one commit, but it's really not a big deal either way.

(If it were a larger change in a series of commits I think it might help the readability of the history to group the different aspects in different commits, but since this one is so straightforward I think it is nice for one commit to contain all of the relevant changes — and I do want to encourage the idea that new features call for updates to code+docs+testing as a group.)

pkgw (approved):
Thanks for following up here! This all looks great. I fixed a typo in one of the documentation files, but this should be good to go once the CI reruns.

PR #899, opened 2022-05-13
Augment error message when opening files for external tools

matz-e:
I missed a publication file when compiling a very old document using
biber.  The only error message I got was

    error: not found

which isn't very helpful.  I've tried to augment the error, now it shows

    error: can't open path `pub.bib`: not found

Hope this is helpful.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,15 +37,15 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.57"
+version = "1.0.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08f9b8508dccb7687a1d6c4ce66b2b0ecef467c94667de27d8d7fe1f8d2a9cdc"
+checksum = "bb07d2053ccdbe10e2af2995a2f116c1330396493dc1269f6a91d0ae82e19704"
 
 [[package]]
 name = "app_dirs2"
-version = "2.4.0"
+version = "2.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d6f731ef6e0f345c835c86724f619d3a73d8be0a9505b83d5e2fa34fd1d50261"
+checksum = "ed3f1233fbc4700ed960dbaa89fff962a7b77e614bf11570e7aba2493c9a41b7"
 dependencies = [
  "jni",
  "ndk-context",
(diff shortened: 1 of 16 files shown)


pkgw:
As for the implementation, Tectonic has the macro `ctry!` ("contextualized try!") for providing a bit more context in error reporting output, building on the `anyhow` crate. Would you be able to use that instead of `map_err`? There are some examples of the usage in the file in question; the basic syntax is:

```
ctry!($fallible_expr; "format string {}", args...)
```

matz-e:
Thank you for the comments! I'll switch to the macro.

matz-e:
Thanks, much nicer that way. Now gives me the following:
```
error: can't open path `pub.bib`
caused by: not found
```

pkgw:
D'oh, now it looks like the new release of Rust 1.61 has broken some of our cross-build infrastructure, which happens occasionally. I will hope to get that fixed up here, but I'm about to go offline for two weeks, and I may not be able to come up with a fix before that happens. So this PR may have to sit idle for a little while — the cross-build setup is *very* gnarly and I'm definitely not going to ask anyone else to try to understand what the heck is going on in its guts.

matz-e:
No worries, hope you can enjoy your time offline! Looks to me like a bunch of linking issues, but I'm also no expert on `cross`.

pkgw:
I was hoping to take a look at this this week, but I've been tied up getting ready for a big astronomy conference that's next week, so I'm afraid I won't be able to do that for a little while longer ...

pkgw:
Cool, I think I've identified the issue here. There was a linker behavior change introduced by:

https://github.com/rust-lang/rust/pull/93901/

That looks like it has broken the extra static libraries added here in our cross-builder images:

https://github.com/tectonic-typesetting/tectonic-ci-support/blob/master/cross-images/06_setup_env.sh#L24

Looking at the logs, these libraries are appearing on the linker command line well ahead of the static libraries like FreeType that depend on them. Ideally we would make sure that the extra deps appeared at the end of the link line, but it's not obvious to me how to make that happen. But the new `+whole-archive` static library linker option should restore the old behavior that worked.

I'll try creating some new cross-builder images, which should magically fix this problem if the diagnosis and solution are correct.

pkgw:
I think I finally fixed the cross-build issues!

pkgw:
Finally!

matz-e:
Thank you for the merge! Sorry, been very busy myself!

PR #907, opened 2022-06-16
fix #906: In Bibtex pass, Process all .aux files generated by the .tex with Bibtex engine.

### LINKED ISSUE #906: When a .tex file generates multiple .aux files, only one .aux file can be processed by Bibtex
### Descriptions
Sometimes a `.tex` file may generate more than one `.aux` files. For example, in [the thu-thesis project](https://github.com/tuna/thuthesis), multiple `.aux` files is created for multiple isolated environments which have dedicated bioblography.

However, currently in Tectonics, only the `.aux` file with the same basename as the `.tex` file would be processed by Bibtex, which causes incorrect compilation result when there are multiple `.aux` files generated by the first `Running Tex` pass.

### Reproduction procedures
1. Download [this project](https://github.com/tuna/thuthesis), run `xetex thuthesis.ins` to generate `.cls`, and uncomment the line `\input{data/appendix-translation}` in `thuthesis-example.tex`, for the extra `.aux` files to be generated.
2. Compile without `tectonic`. Run the following commands:
```shell
xelatex thuthesis-example.tex
bibtex thuthesis-example.aux
bibtex thuthesis-example-translation.aux
bibtex thuthesis-example-index.aux
xelatex thuthesis-example.tex
xelatex thuthesis-example.tex
```
3. Compile with `tectonic`. Run the following commands:
```shell
tectonic --synctex --outfmt=pdf --outdir=./out -k thuthesis-example.tex
```

### Expected behaviour
The `table of contents` of compiling with `xelatex` and three `bibtex` commands (as step 2) is the following:
image

And in the working directory, all three `.bbl` file are generated:
image

### Actual behaviour
The `table of contents` of compiling with `tectonic` (as step 3) is the following:
image

And in the output directory, all three `.aux` file are generated, but only one `.bbl` file is generated:
image

### Analysis
https://github.com/tectonic-typesetting/tectonic/blob/c77ef784564b26114c129c3d49349911835eb822/src/driver.rs#L1143-L1144
In `driver.rs`, the variable `aux_file`  is assigned by replacing the `.tex` with `.aux` in the input file name, which is lately assigned to `ProcessingSession.tex_aux_path`. 
https://github.com/tectonic-typesetting/tectonic/blob/c77ef784564b26114c129c3d49349911835eb822/src/driver.rs#L1809-L1816
And in function `bibtex_pass`, only the `tex_aux_file` mentioned above is processed by the bibtex engine.

**Thanks all the developers for the amazing TeX engine!** 
I will try to raise a PR to fix the problem in the following hours.

Starrah:
Fix #906 : When a .tex file generates multiple .aux files, only one .aux file can be processed by Bibtex.
For the motivation of this PR, please refer to #906 for details.

Instead of only process the `.aux` file with the same basename as the main input file, Now in function `bibtex_pass`, all the `.aux` files generated by previous passes are processed by BibtexEngine.
To implement this, all intermediate files generated by previous passed should be listed.

The intermediate files is listed by listing files in the `MemoryIo`, refers to function `BridgeState.get_intermediate_file_names`. I'm not sure if there is a better way.

**Thank you for time in reviewing this!**  Please feel free to edit the PR if needed.


pkgw:
Thanks for submitting! FYI, I have been away from the computer a great deal in the past month, so it will probably take me a little while to get to reviewing this. You are welcome to ping me in a message if I lose track of things.

pkgw:
OK, step 1: I've pushed an update to your branch to catch it up to the main development branch, which just received some fixes to the CI system.

pkgw:
Thank you very much for your contribution!

I have no comments on the implementation — that looks great.

The only CI error is arising because there is not any test coverage of the new code. Would you be able to create a test case that exercises this functionality? The test suite does *not* have access to the full LaTeX stack (since it would be so large), so the test case would have to manually create some sample files. There is a `bibtex` test that exercises, well, BibTeX, but you might need to adapt it into the `executable` test in order to get the coverage of your new code.

Thanks again, this is a very nice and clearly-explained contribution!

Starrah:
> Thank you very much for your contribution!
> 
> I have no comments on the implementation — that looks great.
> 
> The only CI error is arising because there is not any test coverage of the new code. Would you be able to create a test case that exercises this functionality? The test suite does _not_ have access to the full LaTeX stack (since it would be so large), so the test case would have to manually create some sample files. There is a `bibtex` test that exercises, well, BibTeX, but you might need to adapt it into the `executable` test in order to get the coverage of your new code.
> 
> Thanks again, this is a very nice and clearly-explained contribution!

@pkgw Sorry, but I find it too hard for me to adapt the test case in `bibtex` test into `executable` test. As far as I noticed, in `bibtex` test, we **directly create the `MemoryIo` and `CoreBridgeLauncher` instance** and direcly use the `BibtexEngine` to process the `.aux` file, in which the `.aux` file is read from `FilesystemIo`. 
Since there is no full LaTeX stack in the test suite, I cannot make the `.aux` generated by the TeX engine, instead I have to provide `.aux` files in the test suite and see whether `.bbl` and `.blg` files are generated. However, **only files in `MemoryIo` will be found, as below**
https://github.com/tectonic-typesetting/tectonic/blob/a1d1212a529e9432accc688adc09ca540e51c797/src/driver.rs#L1675-L1683
**when trying to process with Bibtex. Also, the generated `.bbl` and `.blg` files are saved in the same `MemoryIo`.**
However, in `executable` test, tectonic is called as an isolated process, in which I cannot have access to the inner program data such as `MemoryIo` in the test case. For the reasons above, I cannot think out of a way to test my code in `executable` test.
I think maybe we can test it in a similar way as `bibtex`, i.e. to directly operate the `MemoryIo` instance and save our pre-generated `.aux` file into it. **So is there a way to put files into the `MemoryIo`** of a `ProcessingSession` before running the `ProcessingSession`? If there is not, how can I test the `fn bibtex_pass()` in `src/driver.rs`? In fact, I noticed that the `fn bibtex_pass()` is *not* covered *before* my PR.

pkgw:
@Starrah Hmm, yes. To test this will require a different approach with some custom TeX code that manually creates the necessary `.aux` file. The `biber` tests in the `executable` harness show how to do this in a similar situation — it is somewhat tedious. If you don't have the time to try to get something working, I can try to set up a test myself.

pkgw:
I've updated your branch with a new test that ought to cover the relevant code. It's not trivial to do so!

Issue #835, opened 2021-10-16
`\inputminted` support: files missing from shell-escape working directory


kwshi:
#708 introduced shell-escape support, which (as motivated in #38) is needed for the exceedingly popular `minted` package for typesetting syntax-highlighted source code.

However, shell-escape support alone is not enough to support `minted`--in particular, the `\inputminted` command does not work correctly.  Here is a simple project demonstrating the issue: [tectonic-minted-test.tar.gz](https://github.com/tectonic-typesetting/tectonic/files/7358064/tectonic-minted-test.tar.gz).  Its contents are:
- `hello.tex`
  ```latex
  \documentclass{article}
  \usepackage{minted}
  \begin{document}
  \inputminted{python}{world.py}
  \end{document}
  ```
- `world.py`
  ```python
  print("hello world")
  ```

When running `tectonic -Z shell-escape` in the above project, the following error is returned:
```
Error: cannot read infile: [Errno 2] No such file or directory: 'world.py'
warning: command exited with error code 1
error: failed to execute the shell-escape command "pygmentize -l python -f latex -P commandprefix=PYG -F tokenmerge -o _minted-hello/72214DB4E1E543018D1BAFE86EA3B444BA1FB043E1554A6696271B99677258E5.pygtex world.py": execution of the request failed
error: hello.tex:4: Package minted Error: Missing Pygments output; \inputminted was
probably given a file that does not exist--otherwise, you may need
the outputdir package option, or may be using an incompatible build tool,
or may be using frozencache with a missing file.

See the minted package documentation for explanation.
Type  H <return>  for immediate help
error: halted on potentially-recoverable error as specified
```

After some digging around, I think I've figured the reason for this error.
- Unlike the "original" LaTeX compilers, which run shell-escape commands from the same working directory as the project, Tectonic **runs shell-escape commands from fresh, empty temporary directories** (I checked this by making a blank LaTeX file containing some dummy text and `\write18{pwd}`).
- **Other files in the project directory are not copied over to this temporary directory** (`\write18{ls -l}` shows an empty list).
- As such, when `minted` invokes the `pygmentize` shell command, it tries to load the source file (expecting it to be in the current _project_ working directory) but fails to find it because its working directory is actually set to something like `/tmp/.tmpVGQl7T`.

This being the cause gives rise to a temporary workaround: passing `\inputminted` an _absolute_ rather than relative path makes Tectonic run correctly.

[...]

PR #909, opened 2022-06-25
Add shell escape working directory option

0x00002a:
Hi

This adds support for a `-Z shell-escape-cwd=<dir>` which sets a working directory for `shell-escape` (`-Z shell-escape-cwd=$(pwd)` gives essentially the "normal" behaviour of engines for `shell-escape`). This allows using stuff like `\inputminted` (see: #835) to work the same as with other engines.

I have tested this by hand for my own use-case (#835) and it seems to work. I have not tested the toml format (because I don't use the v2 interface and couldn't get it working). I couldn't get a unit test case for it working properly though, if I run `tectonic` normally `\write18{pwd > file}` will work and send `pwd` to `file`, but when run as part of the tests it wouldn't do anything, even if I gave it an absolute path. If a test case for this is needed, I'd appreciate some pointers as to what the issue could be.

fixes: #835

### DIFF
--- a/crates/docmodel/src/document.rs
+++ b/crates/docmodel/src/document.rs
@@ -229,7 +229,6 @@ pub struct OutputProfile {
     ///
     /// Directory is not managed and any files created in it will not be deleted.
     ///
-    /// Only meaningful when shell_escape = true.
     pub shell_escape_cwd: Option<String>,
 }
 
--- a/crates/docmodel/src/document.rs
+++ b/crates/docmodel/src/document.rs
@@ -221,6 +221,16 @@ pub struct OutputProfile {
     /// shell-escape opens enormous security holes. It should only ever be
     /// activated with fully trusted input.
     pub shell_escape: bool,
+
+    /// Directory to use as the cwd for shell escaped execution.
+    ///
+    /// Setting this to $(pwd) gives the same relative path shell-escape behaviour
+    /// (e.g. for \inputminted), as other engines, such as xelatex
+    ///
+    /// Directory is not managed and any files created in it will not be deleted.
+    ///
+    /// Only meaningful when shell_escape = true.
+    pub shell_escape_cwd: Option<String>,
 }
 
 /// The output target type of a document build.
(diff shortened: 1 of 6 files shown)


pkgw:
Thank you for your submission! I think this could be a good approach to helping shell-escape work in more contexts when people really need it. I hope to be able to review this promptly, but please feel free to ping be if you don't hear anything for a little while. As ever, things are busy.

pkgw:
OK, I finally was able to take a look at this. The implementation looks very nice and straightforward.

How do things work out if the CWD parameter is a relative path? In particular, for the TOML serialization, if someone is going to specify the setting at all it will probably be a value like ".", so we should check that works. (And conversely, we issue a warning when the value is an absolute path, since that indicates a non-portable build.)

Also, what will happen if the CWD parameter is the source directory, and one of the external tool inputs is a local source file? I'm worried about the tool-launching code trying to effectively copy a file onto itself, which I think will destroy the file in most cases.

0x00002a:
> Also, what will happen if the CWD parameter is the source directory, and one of the external tool inputs is a local source file? I'm worried about the tool-launching code trying to effectively copy a file onto itself, which I think will destroy the file in most cases.

@pkgw I'm not exactly sure what you mean by this? External tools expect cwd to be the source dir usually afaik, because thats the behaviour of every other engine I know of with shell-escape?

0x00002a:
> How do things work out if the CWD parameter is a relative path? In particular, for the TOML serialization, if someone is going to specify the setting at all it will probably be a value like ".", so we should check that works. (And conversely, we issue a warning when the value is an absolute path, since that indicates a non-portable build.)

Did a quick test, seems to take the path relative to the pwd when `tectonic` is invoked. So `cd prefix && tectonic ./dir/doc.tex -Z shell-escape-cwd=.` will put the shell cwd at `prefix`. Does tectonic support portable builds with tectonic invoked "out of source" as it were? if so I suggest we make it so relative paths are relative to the toml config file and make them absolute internally at runtime

pkgw:
I'm sorry for letting this sit for so long. In the interest of moving things along, I've merged as-is, and we can touch up behavior in any corner cases if/when they arise. Thank for your contribution! Once again, my apologies for taking so long to merge.

9SMTM6:
This works with the V2 CLI as expected btw.

One side effect though is that intermediate files will be placed in the directory thats marked as working-directory, depending on the commands that were used. I dont know how that could be solved generally without potentially problematic approaches.

pkgw:
@9SMTM6 Right — that's why I view this approach as a workaround that ideally we'd avoid. And I still suspect that we can improve the encapsulated shell-escape mode to succeed more often.

Issue #8, opened 2017-03-15
Complain if a filesystem file with an absolute path is accessed


pkgw:
The idea is that we want to identify when the document compilation requires "external" ("system"?) files that may make the document build not reproducible. I think there's a way to do this that's helpful but doesn't make the user experience too annoying. We probably just want to warn on the first system file rather than each one, since I expect that the usual case for this will be system fonts, which usually involves accessing a whole bunch of related files in a group.

pkgw:
Finally fixed in #911!

malbarbo:
@pkgw How can one avoid such warnings?

I'm getting warning such as:

```
warning: accessing absolute path `/nix/store/91qbwa97mgiwanvgaihc2rykxy6lcfd4-source-sans-3.046/share/fonts/truetype/SourceSans3-Semibold.ttf`; build may not be reproducible in other environments
```

pkgw:
@malbarbo Currently there isn't a way to disable the warning, but one could certainly be added.

I'm curious about your use case. I presume that you're doing builds in some Nix setup where absolute paths are working their way into the system? (I hardly know anything about Nix, I'm afraid.) Do you think it is *possible* (even if annoying/undesirable) to reorganize your document and/or build process to avoid the absolute paths?

malbarbo:
@pkgw I do not have any specific configuration or use case, the warning show up with a minimal example.

To reproduce the issue start the nixos [iso](https://channels.nixos.org/nixos-22.05/latest-nixos-gnome-x86_64-linux.iso) in a virtual machine, download the latest [tectonic](https://github.com/tectonic-typesetting/tectonic/releases/download/tectonic%400.12.0/tectonic-0.12.0-x86_64-unknown-linux-musl.tar.gz) and build the following example

```tex
\documentclass{beamer}
\usetheme{metropolis}

\begin{document}
\end{document}
```

Note that the warnings are only generated when using metropolis theme.

vlasakm:
> @pkgw I do not have any specific configuration or use case, the warning show up with a minimal example.

Thank you for the minimal example @malbarbo. The issue isn't NixOS specific, it's #9 (see #965, in particular https://github.com/tectonic-typesetting/tectonic/issues/965#issuecomment-1317865762).

The only problem is that for example in the case of `metropolis` you are not loading the fonts yourself, but the package does. In this case the culprit is in the file `beamerfontthememetropolis.sty`.

[...]

PR #911, opened 2022-06-28
Warn when opening absolute paths (pared down version of #806)

pkgw:
I would like to get some of the changes proposed by @ralismark in #806 into the codebase, but that PR has been idle for a while. I'm going to supersede it with this pared-down version.


pkgw:
Closes #8!

Issue #916, opened 2022-07-15
Can halt_on_error be disabled?


jooooscha:
I have a document which causes 
```
Running TeX ...
error: pdflscape.sty:105: Package pdflscape Error: The driver option `' is not compatible
(pdflscape)                with XeTeX, using driver for XeTeX (dvipdfm) instead.

See the pdflscape package documentation for explanation.
Type  H <return>  for immediate help
error: halted on potentially-recoverable error as specified
```
when compiled with tectonic.

Assuming the error is, indeed, recoverable I dug around in the code and found the *message* to be caused by the `halt_on_error` variable in the `TexEngine` struct.
I see that it is set to false in some tests but I did not found out how to disable this for compiling.

Did I miss something?
How can I specify that I do not want to halt on `potentially-recoverable` errors?

Thanks in advance

pkgw:
Yes, there's an "unstable option" `-Z continue-on-errors` that should do what you want. It's mentioned here in the documentation although I'll admit that it's not super easy to find:

https://tectonic-typesetting.github.io/book/latest/v2cli/compile.html#unstable-options

jooooscha:
Ah, yes. Thank you, I did miss that. 

However, enabling this does not seem to change anything. I get the same error. I tried `tectonic file.tex -Z continue-on-errors` and a bunch of other permutations. I also tried the V2 interface.
Anything I could have missed? 

I am sorry, if this is a dump mistake.

For the record, I am on version 0.9.0.

pkgw:
That's odd. Maybe I'm overlooking something simple here. Would you mind providing a minimal example document that fails for you?

jooooscha:
Umm, that could actually take a while. 
The document is based on a huge template I have to use.

I will investigate what's causing the error and come back to you.

Thanks, anyway, for you time :)

jooooscha:
Ok, I am not sure this is helpful, but this is the smallest example I get the same error.

```
\documentclass{article}

\usepackage{pdfpages}
\pdfoutput=1

\begin{document}
Hello World.
\end{document}
```

Again the `-Z continue-on-errors` option does not seem to change.

If this does not help, I accept that the template will not work with tectonic, If I remove the `\pdfoutput=1` in the template some other control sequence is not found. Debugging that could take for ever 😅

vlasakm:
[...]

PR #917, opened 2022-07-16
Use the value of the continue-on-errors option

vlasakm:
The actual effect of the continue-on-errors option probably got lost in refactoring, see https://github.com/tectonic-typesetting/tectonic/issues/916.

A bit off topic: While investigeting, using [`into()` for `bool` to `c_int` in `xetex_engine`](https://github.com/tectonic-typesetting/tectonic/commit/bda8f7e1ec82a185bbbac55d5ddd4fd4d185eec4) popped into my mind, but I am not sure if it is actually an improvement, so I left it out of this PR. It seems to me that both my "compaction" and the previous version assume that `libc::c_int == i32`, which is mostly true, I am not sure it is a problem we should worry about.

### DIFF
--- a/crates/engine_xetex/src/lib.rs
+++ b/crates/engine_xetex/src/lib.rs
@@ -191,37 +191,30 @@ impl TexEngine {
             // Note that we have to do all of this setup while holding the
             // lock, because we're modifying static state variables.
 
-            let v = if self.shell_escape_enabled { 1 } else { 0 };
-            unsafe {
-                c_api::tt_xetex_set_int_variable(b"shell_escape_enabled\0".as_ptr() as _, v);
-            }
-
-            let v = if self.halt_on_error { 1 } else { 0 };
-            unsafe {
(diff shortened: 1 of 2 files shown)


pkgw:
The `into()` change seems like a nice one. The way that the `libc` crate is implemented, there shouldn't be any improper assumptions about the size of `c_int`, but stuff like the docs will show a concrete type like `i32` if that was the correct value for the system where the docs were being built.

It looks like some of the cross builds have an unrelated failure ... which is a bit odd, since they've been working with the current version of Rust, 1.62, in recent builds. But `rustup` just had an update, so I guess maybe that's caused a problem.

vlasakm:
Hm, as long as `c_int` is a type alias to some primitive type that has an `Into<bool>`, `into()` should work, so it seems better than the current state where integer literals get inferred as `i32` and it could potentially not compile on some platforms.

AFAICT `c_int` is currently `i32` on all `libc` platforms.

So I'll add that commit as is.

I hope the CI failure is undeed unrelated, and unfortunately due to lack of experience can't offer any help in this regard.

Issue #923, opened 2022-08-16
-Zhelp still requires input file


timvisee:
`tectonic -h` suggests to use `tectonic -Zhelp` to view unstable options.

This command doesn't work though, as an `<input>` file is required.

```
$ tectonic -Zhelp
error: The following required arguments were not provided:
    <input>

USAGE:
    tectonic <input> --outfmt <format> --chatter <level> -Z <option>... --pass <pass> --format <path> --color <when>

For more information try --help
```

```
$ tectonic -Zhelp -
Available unstable options:

    -Z help                     Lists all unstable options
    -Z continue-on-errors       Keep compiling even when severe errors occur
    -Z min-crossrefs=<num>      Equivalent to bibtex's -min-crossrefs flag - "include after <num>
                                    crossrefs" [default: 2]
    -Z paper-size=<spec>        Change the default paper size [default: letter]
    -Z search-path=<path>       Also look in <path> for files, like TEXINPUTS. Can be specified
                                    multiple times.
    -Z shell-escape             Enable \write18
```

PR #929, opened 2022-09-10
Make `tectonic -Zhelp` work

pkgw:
Cf. #923. `tectonic -Zhelp -` is enough, and this is a hack, but it should give a nice little quality-of-life improvement.

Also, fix a long-standing thing that's bothered me and stop my PDF viewer from popping open every time I run the test suite.


pkgw:
Windows builds were having some Cranko dirtiness problem so I've gone ahead and updated the CI to use my now-standard three-stage workflow. This will also be useful to have in place if/when we start depositing to Zenodo and registering DOIs, since that workflow requires a single release-commit creation stage.

PR #945, opened 2022-10-04
Update the book to have a guide on setup with `AucTeX`

KaranAhlawat:
This PR adds a How-To section to the Tectonic book on setup of AucTex and Tectonic together.
Only a single MD file has been added to the docs source, containing all the relevant information.


KaranAhlawat:
Just a quick question, do you mind adding the `hacktoberfest-accepted` label to this PR so that I can earn some brownie points with Hacktoberfest?

pkgw:
Sorry about taking so long to review this! This looks great overall. I made a few changes to your PR:

- added hard linewraps
- suggested `tectonic -X` rather than `nextonic` everywhere
- fixed a typo

The `tectonic -X` way of using the V2 CLI is what's generally recommended everywhere else so I think it should be the default here. Based on your docs, I don't expect that the two-word command should cause any problems with your integration, right?

KaranAhlawat:
It shouldn't cause any problems, no. Thanks for merging! Keep up the good work :)

RobertSDThomas:
I am an experienced user of emacs and LaTeX but not of their mechanics. Having a new OS, Ubunto 22.04, I have naively set up emacs with AUCTeX and LaTeX with Tectonic. I hope less naively I have followed the above directions to create and fill my init.el file, but I still get the following with C-c C-c in a .tex file:
"ERROR: AUCTeX cannot find a working TeX distribution.
Make sure you have one and that TeX binaries are in PATH enviromnent variable."
I have no idea what "PATH environment variable" even means. Help?

KaranAhlawat:
Have you installed Tectonic on your system? If so, how did you do that?
The PATH environment variable is used to figure out which binaries are available to run anywhere on the system (such as when you do `cat`, or `vim` in a shell, those commands are present on the PATH).

Emacs has it's own PATH variable, which needs to be set appropriately to find the tectonic binary.

Also, if you can share the actual relevant changes you made in your `init.el`, that will be helpful in debugging if anything is wrong in the code itself.

@RobertSDThomas

KaranAhlawat:
Good morning to you too, and first off, that's one of the best written GitHub comments I've ever read, so thank you for that. 

> Yes, I have installed Tectonic, using the utility supplied as part of Ubuntu 22.04, which I just had installed on a new hard drive. It asks whether one wants TeXlive whole, basic TeXlive, or Tectonic, which I had not heard of but sounded when I read about it like the coming thing.
Can you please share the name of the utility? Furthermore, we should test if you can actually run Tectonic from the terminal. Please open a terminal, and type `tectonic --help` and share the output it displays.

Coming to the PATH environment variable, it is one of the many environment variables in your system. Consider an environment variable as a key-value pair (the key being PATH, in this case
and the value being locations on your filesystem where the system looks for executable programs). Programs can read the value of these environment variables for various purposes, such as Emacs using PATH to find out where a particular executable (tectonic) is. 
A typical PATH environment variables can look like the following:
```
$ echo $PATH
/usr/bin:/usr/local/bin:/bin:/home/<username>/.local/bin:
```
Then, the system looks at all of these paths one by one to find where the executable you wanted to run is, and if it can't find it, it says it couldn't find it on the PATH environment variable.
I hope that makes things clearer.

For listing files, `ls` is indeed the command, but that does not list the "hidden" files, which are files and directories beginning with a `.`. For that, you can run `ls -la`, which should list your `.emacs.d` directory, and you can move into it with `cd .emacs.d`. Doing an `ls` in this directory should list your `init.el` file.

> how to deal with your remark
Not sure which remark you are referring to, but happy to help 😄 

> All I've thought of for this is to put the tex file to be compiled in the build directory so that when the pdf file (what I want) is put there, AucTeX will find it because it's the same directory. But I don't know where the build directory is or whether my idea is sound.
This is a solved problem, the AucTex section in the Tectonic book provides you with the relevant code snippet, although please do let us know if it doesn't work by creating a new issue in this repository (you can tag me there so I can make a PR fixing the issue).

Lastly, your `init.el` looks good except one thing. You need to actually have the Auctex package installed for this to work. For that you can run the following command in emacs
`M-x package-install RET auctex RET`. That should sort it out for you, and then your `init.el` should behave as expected.

Best wishes,
Karan

RobertSDThomas:
At the foot of your message, there is this text:
Reply to this email directly, view it on GitHub, or unsubscribe.
where the latter are links but no hint is given on how to do the first. Is it really customary to put such correspondence as this as comments?

Thank you for your reply.

The software with which I installed Tectonic (21.7 MB according to it) is called only Ubuntu Software in 22.04. I have the same icon, name, and offerings on my 20.04 machine, where it is offering an update of device firmware from version 1.20.0 to 1.28.0. I can't get to an 'about' window for it.

When I ran 'tectonic --help' in the terminal, I got
Tectonic 0.14.1
Process a (La)TeX document

USAGE:...
FLAGS:...
OPTIONS:... and
ARGS:...
I presume as you would expect. I have all this but hesitate to include it. 

I note that the default for where to put output is where the input was. Is this a change from when you wrote your original material some time ago? It would deal with my question about your remark (which you did not realize was then quoted), namely
"AucTeX expects the output PDF after compiling to be in the same directory as the input file. So it will error out when that is not the case, since Tectonic places the output in a build directory."
The problem worrying me has gone away, as you say later in your message.

I take it from your explanation of PATH environment variable that I don't need to do anything about it.

 Thank you for the explanation, and also for showing me how to list 'hidden' files. Having sent you the listing, obtained otherwise, we know what's in init.el.

It looked as though the goal was within sight and that installing AucTeX would be simple. When emacs looks at a file with extension .tex it puts Latex in its menu bar, and Auctex things like its manual are available, but when I followed your instruction 'M-x package-install RET auctex RET' I got '[No match]' as the response in the minibuffer. Using 'M-x package-list-packages' however shows auctex version 13.3.0 as installed.

We have now verified a number of things but not changed anything. Accordingly I was sceptical that the result would be any different. To verify that, I typed C-c C-c with a Latex file open and got the query about Latex being the default and then the same error message as before warning that auctex could find no TeX in those binaries. We are now more familiar with the territory, but we are in the same position as before. The one thing not verified was the contents of the file init.el; so I learned what the command is to view its contents and checked that it is indeed there with the contents that emacs showed as there yesterday.

Thinking that it might be good for you to know just where tectonic is in the file system, I learned how and used 'command -v tectonic' to get '/snap/bin/tectonic'.

KaranAhlawat:
If you wish to reply via email, you can simply view this thread in an email client, and then use the "reply" feature of said client. That will fill in the necessary details, and you just need to write your message accordingly.

If you rather wish for this to be a private correspondance, you can email me at ahlawatkaran12@gmail.com.

KaranAhlawat:
Also, just to answer your first question, it's not customary to put it as comments in a closed Pull Request. Ideally you'd open up a new issue in the appropriate place, or reach out for help via some other medium (email, Discord, etc). But this is fine too, no harm done.

PR #953, opened 2022-10-15
Fix low level dvipdfmx issues

### LINKED ISSUE #904: warning: Tectonic does not support `config' special. Ignored.
I get this warning message each time I invoke the `tectonic -X build` command for compilation. Unfortunately, I couldn't figure out if there's anything wrong with the compiled PDF nor do I see any mention of it anywhere. So, I was wondering why does `tectonic` such a warning? Is it something I should investigate further?

vlasakm:
This code change is bigger than I anticipated, but nonetheless, here it is. I appreciate any feedback and testing (I don't use LaTeX much, so don't have real documents to test with).

Both commit messages are reproduced below.

In upstream dvipdfmx the `dvidpfmx:config` DVI special can be used to
pass arguments to dvipdfmx. This has seen some use in XeTeX packages.

For example, the ubiqitous package "hyperref" uses:

    \special{dvipdfmx:config C 0x0010}

This is normally parsed by dvipdfmx as command line arguments, i.e. as
if xdvipdfmx was ran as:

    xdvipdfmx -C 0x0010

The `C` option is a bitset of compatibility flags. In this case hyperref
needs the option `OPT_PDFDOC_NO_DEST_REMOVE`, which makes dvipdfmx
handle PDF destinations differently (I didn't check thoroughly how).

The `opt_flags` global variable used to hold these bitflags and the
`dvipdfmx:config` special could be used to set them. #92 removed that
option.

While passing arbitrary command line arguments to dvidpfmx is probably
not what we want, I definitely know, that we at least need the
compatibility flags to work, and there is probably no other way packages
can set them.

The code that does the option parsing is mostly port from the 'C' option
parsing from dvipdfmx's `read_config_special` and `do_args_second_pass`.
IMHO the way `opt_flags` gets modified is undefined behaviour and does
other weird things with negative values, but the purpose of this commit
is to increase compatibility with upstream, not to fix things (which
should preferably be done upstream IMO).

Like with other options, instead of having a global variable, the
pointer to the value in `dvipdfmx_main` is passed around instead.

While at it, I also added `z` and `g` options which also seem to be used
by LaTeX, so this should be enough for a while.

Hopefully fixes #904.

A previous change left the `fullname` variable unconditionally set to
`NULL`, which meant that all attempts to find streams would fail.

The `fullname` variable is however no longer needed, since Tectonic
doesn't have the two step Kpathsea file workflow (find full name +
open), but it suffices to just open the file with `ttstub_input_open`.

### DIFF
--- a/crates/engine_xdvipdfmx/Cargo.toml
+++ b/crates/engine_xdvipdfmx/Cargo.toml
@@ -1,4 +1,4 @@
-# Copyright 2021 the Tectonic Project
+# Copyright 2021-2022 the Tectonic Project
 # Licensed under the MIT License.
 
 # See README.md for discussion of features (or lack thereof) in this crate.
@@ -30,4 +30,4 @@ cc = "^1.0.66"
 [package.metadata.internal_dep_versions]
 tectonic_bridge_core = "thiscommit:2021-01-17:fohCh1sh"
 tectonic_errors = "5c9ba661edf5ef669f24f9904f99cca369d999e7"
-tectonic_pdf_io = "thiscommit:2022-03-29:6nizWic"
+tectonic_pdf_io = "thiscommit:2022-10-21:pkYKcMI"
--- a/crates/engine_xdvipdfmx/xdvipdfmx/dvipdfmx.c
+++ b/crates/engine_xdvipdfmx/xdvipdfmx/dvipdfmx.c
@@ -339,6 +339,11 @@ dvipdfmx_main (
   int has_id = 0;
   unsigned char id1[16], id2[16];
   struct pdf_setting settings;
+  int pdf_version_major = 1;
+  int pdf_version_minor = 5;
+  int compression_level = 9;
+  double annot_grow_x = 0;
+  double annot_grow_y = 0;
 
   assert(pdf_filename);
   assert(dvi_filename);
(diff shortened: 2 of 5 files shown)


vlasakm:
I forced pushed a rebase with one my mistake fixed and the global variables commit (which fixed the Windows CI issue) incorporated into the first one.

pkgw:
Thank you very much for working on this! I've been meaning to look into the `config not supported` warning for a while, but I didn't realize that it could have a tangible impact on behavior.

My GitHub backlog has gotten pretty deep but I'll try to review this soon. In the meantime, if you're able to find the time to add some test cases exercising this functionality, that would be awesome, but that might end up being its own can of worms ...

9SMTM6:
How can I as someone not overly familiar with tectonic help with this?

Should I pull, try to build it, and see if I sopt any issues with files that were throwing that error?

How would I check if the fox works as intended? Apparently when it comes to hyperref without this working external links to sections dont work properly?

pkgw:
@9SMTM6 Thank you for your interest! Yes, something like that would be a way to test. The `config not supported` message isn't an error per se, but if we're handling things properly, the targets for links within a PDF should work more reliably, I think?

@vlasakm I made the time to review this and it looks great! I pushed a couple of small tweaks to your branch but overall I don't have any suggestions besides what I mentioned before about how it would be nice to add some test cases for some of this.

Issue #954, opened 2022-10-16
tectonic 0.11.0 generating html fails


ville-h:
using tectonic 0.11.0 I was trying to generate html from:
```
\documentclass{article}
\begin{document}
    xxx
\end{document}
```
with the command: `tectonic --outdir build --outfmt html src/xxx.tex`, but it errors out with:
```
Running spx2html ...
warning: imbalanced tags; had to close 0 to find `div`
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', crates/engine_spx2html/src/lib.rs:1660:57
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Generating a pdf works fine.

pkgw:
Well that's embarrassing! You should know that as things currently stand, a great deal of special options are needed to get HTML output (I should probably add a warning), but we certainly shouldn't panic.

ville-h:
I did have a look at the commit, but based on [the message](https://github.com/pkgw/tectonic/commit/462bb92cb9536daf6776227bc2cab5214b111974#diff-86eb9aca9675a44d1e7afd0ad70f4664fbe2ec8e7de58b9d413bc3a897bdf460R1988) I wouldn't know what to do. The word `template` appears to be key, but how do I specify one? What does it mean concretely to be "html-compatible"? The comment above doesn't make the steps I need to take any more concrete either. I thought tectonic was supposed to download everything as-needed?

pkgw:
Those are all good questions. I think that, unfortunately, there's just too much context to pack into one error message, and what you read is the best I could come up with.

The longer explanation is that producing HTML requires the engine, the TeX code, and HTML/JS/CSS templates that all work together. The engine has core support, which you can turn on with the `--outfmt` option, but right now the only existing example of the "full stack" is the specialized [tt-weave] framework that produces [XeTeX: A Pseudoprogram][xap]. So I'm afraid that the `--outfmt` option is more or less useless unless one is willing to read the HTML output code to understand the various `\special`s and then write some TeX, HTML, CSS, and JS that stitches everything together. The vision is that one day we'll include all of the necessary templates and customizations to the standard document classes and packages so that what you've attempted will Just Work, but sadly that's still far off. My plan is to shake down the details with specialized cases like [XAP][xap] and build up the "transparent" support as we better understand the problem space.

[tt-weave]: https://github.com/tectonic-typesetting/tt-weave/
[...]

PR #955, opened 2022-10-21
engine_spx2html: fix up behavior when compiling un-customized documents

pkgw:
See issue #954. We shouldn't panic! And we also now try to print a warning that will hopefully be helpful.


pkgw:
As discussed in #954, this new error message is not exactly crystal-clear, but I can't think of anything better right now :slightly_frowning_face:.

PR #983, opened 2022-12-30
V2 CLI: separate `new` and `init` commands

caiogeraldes:
This is a simple quality of life change I would like to suggest, as it standardizes the CLI of tectonic with that of tools such as `cargo` or `poetry`. Not much else to say, tbh.

### DIFF
--- a/docs/src/v2cli/init.md
+++ b/docs/src/v2cli/init.md
@@ -1,6 +1,6 @@
 # tectonic -X new
 
-Create a new Tectonic workspace.
+Initializes a new Tectonic workspace in the current directory.
 
 **_This is a [V2 CLI][v2cli-ref] command. For information on the original (“V1”
 CLI), see [its reference page][v1cli-ref]._**
@@ -11,11 +11,8 @@ CLI), see [its reference page][v1cli-ref]._**
 #### Usage Synopsis
 
 ```sh
-tectonic -X new [DIRECTORY]
+tectonic -X init
 ```
-
-If `{DIRECTORY}` is unspecified, the workspace is created in the current
-directory.
 
 #### Remarks
 
--- a/docs/src/v2cli/init.md
+++ b/docs/src/v2cli/init.md
@@ -1,4 +1,4 @@
-# tectonic -X new
+# tectonic -X init
 
 Initializes a new Tectonic workspace in the current directory.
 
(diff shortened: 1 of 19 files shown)


pkgw:
Thank you for the submission! This seems like a nice improvement to me. I have two requests:

1. Can you please add a commit fix the clippy error that the CI is flagging? It's unrelated to your change but also should be a trivial fix. (Sometimes when new versions of Rust come out, we start getting new errors without changing our code, because Clippy gets stricter.)
2. Can you please update the documentation as well? The files are in `docs/src/v2cli/` and `docs/src/SUMMARY.md`.

Thanks again!

caiogeraldes:
The last two commits should hopefully fix the warnings clippy was throwing, most of them concerning the pattern `variable_X as type_Y`.

However, I could not fix one of them without setting `#[allow(clippy::write_literal)]` in the following lines:

[tests/executable.rs, l.741ff. ](https://github.com/tectonic-typesetting/tectonic/blob/92261fae4fe77b7f1d6b2ea28307effd1435b6a2/tests/executable.rs)

```rust
        #[allow(clippy::write_literal)]
        writeln!(
            file,
            "{}", // <= works around {} fussiness in Rust format strings
            r#"\newwrite\w
\immediate\openout\w=first.demo\relax
\immediate\write\w{content-un}
\immediate\closeout\w
\immediate\openout\w=second.demo\relax
\immediate\write\w{content-deux}
\immediate\closeout\w
"#
```

Without it, clippy went into a kind of a loop, asking for more and more curly braces around `content-un` and `content-deux`. Considering it is part of a test which is passing, I assumed the warning could be ignored without any issues. I think the issue is with clippy in this case, and I reckon that the comment near `"{}",` is related.

docs/src/v2cli/init.md
```diff
@@ -0,0 +1,35 @@
+# tectonic -X new
```
pkgw, line 1: The title here should be `init`, not `new`.

pkgw (commented):
Thanks! There's one small thing to fix in the new docs.

As for the Clippy stuff, thanks — when I initially asked there was only one small error, so you ended up doing a lot more work than I wanted to ask of you!

I'll look into the MSVC issue. I think it's something that might not be too hard to fix/work around. I might fix that doc title issue while I'm at it.

PR #994, opened 2023-01-28
Fix everything for Rust 1.67's Clippy

pkgw:
It wants format strings to include variables inline, which is a change that I can get behind.


pkgw:
Sorry, Codecov.

Issue #748, opened 2021-02-10
Implement pdfcrop as single executable (example of tectonic as library)


kpym:
I do not know `rust` so I can't do this, but it will be nice if [pdfcrop](https://github.com/ho-tex/pdfcrop) can be implemented as single executable using tectonic (with precompiled binaries for all major platforms). In this way we will be able to use it without heavy latex distribution installed.

And I think that this well be a good example of usage of tectonic as library.

By the way I opened [an issue](https://github.com/ho-tex/pdfcrop/issues/6) to ask the support of `tectonic` in `pdfcrop`.

pkgw:
Yes, I find `pdfcrop` to be a very useful tool, and I've often wished I could install it in a more standalone fashion.

That being said, Tectonic's PDF manipulation code (from xdvipdfmx) is all in C, so this isn't something that can be done straightforwardly in Rust right now. I would like to one day switch all of the PDF routines over to something written in Rust but that would/will be a *major* undertaking because there are so many features to implement.

Unless I am misunderstanding how `pdfcrop` works — does it primarily operate by running a TeX engine under the hood?

kpym:
> That being said, Tectonic's PDF manipulation code (from xdvipdfmx) is all in C, so this isn't something that can be done straightforwardly in Rust right now. I would like to one day switch all of the PDF routines over to something written in Rust but that would/will be a _major_ undertaking because there are so many features to implement.

Is it a problem that part of the library is in C ? You manage to create `tectonic` CLI, so it should be possible to create a "custom version" of tectonic that do the `pdfcrop` job, no ?
   
> Unless I am misunderstanding how `pdfcrop` works — does it primarily operate by running a TeX engine under the hood?

`pdfcrop` perl script do the following, as far as I understand : 
1) It runs `gs` to evaluate the minimal media/cropbox possible.
2) It creates a temp `.tex` file that `crop` the the `pdf`.
3) It runs PDF/Xe/Lua/TeX to compile this temporary `.tex` file.

So `tectonic` is needed only for the final stage.

pkgw:
Oh, then yes, this probably wouldn't be too hard to recreate as a small Rust tool. The dependency on the external `gs` tool is important to note, but this could indeed make for a nice example. If you're interested in learning Rust this could be a great way to start!

kpym:
@pkgw Yes, I would like but I'm learning `go` now and I have no such time for fun these days :( Sorry. I understand if you have no time either.

[...]

PR #995, opened 2023-01-28
Add pdf crop #748

AlongWY:
look as #748;

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1014,6 +1014,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "hex"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
+
+[[package]]
+name = "home"
+version = "0.5.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "747309b4b440c06d57b0b25f2aee03ee9b5e5397d288c60e21fc709bb98a7408"
+dependencies = [
+ "winapi 0.3.9",
+]
+
+[[package]]
 name = "html-escape"
 version = "0.2.13"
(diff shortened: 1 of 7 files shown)


pkgw:
Thanks for submitting this! This is a very interesting idea.

I want to think a bit about the design here and different ways it could be approached. In the issue you wrote that you made two versions, "It is a standalone one, but it's too big." Can you explain a bit more about what you mean by that? I am not sure that I understand.

AlongWY:
The standalone version use tectonic as a library, the size of compiled binary is same as tectonic, about 11MB.
I think it’s not worth that, so I think maybe the pdfcrop can be a part of tectonic.

AlongWY:
And I will rebase the repo later.

pkgw:
OK, I think I understand now. Thank you for the explanation.

After thinking about this for a bit, the main thing on my mind is that I really do not want Tectonic to depend on GhostScript. This is because that *any* external dependency limits Tectonic's ability to run in different environments (e.g., WASM), and also because I am very worried about the security qualities of GhostScript in particular. In this case, GhostScript is being used in a controlled way so I think that the security implications are not problematic, but I don't want to encourage people to depend on it in their workflows.

Because the functionality in this pull request is not essential to running the Tectonic engine, I think it would be better to keep it in a separate piece of code — either option (1) or (2) that you listed above. I agree that the binary size becomes pretty unreasonable when you link against Tectonic directly, and that's unfortunate.

For use cases like this, it might be very helpful for Tectonic to provide some kind of `tectonic_client` crate that can provide a Rust interface for driving the `tectonic` executable in a structured way. That would help people take advantage of Tectonic typesetting capabilities without having to compile and link in the entire software stack directly, which, as we are seeing here, is a lot to ask for many applications. It wouldn't be able to handle as many use cases as direct linking, but perhaps it would meet a lot of needs in a much more convenient way. Would you be interested in working together to create such a crate?

AlongWY:
I think this is a good idea.
I'll probably finish this in the next few weeks.

Issue #876, opened 2022-03-21
V2 layout and tools like latexdiff


clbarnes:
Many tools, including latexdiff, require you to point them at the root file of the document. As of the V2 recommended/ automatic layout, tectonic doesn't have one, and (presumably) implicitly concatenates the preamble, index, and postamble files. How can this be explained to tools which need a root file? What benefit does this layout provide if it is internal to tectonic and breaks all other latex workflows?

pkgw:
I'll start with the very small-bore answer. One way to get this workflow to work in the V2 paradigm is to make `_preamble.tex` and `_postamble.tex` empty, and just put everything in `index.tex`.

In terms of how tools like `latexdiff` might be supported in the future while "leaning in" to the V2 paradigm, Tectonic might gain a mode similar to `latexpand` that would reduce the input file structure to a single homogenized text stream. The key thing here is that if/as document source structures gets more sophisticated (e.g. #872), Tectonic should really be the source of authority for how to resolve `\input` inclusions. Also, this mode would come with some guarantees about the superficial parse-ability of the input, which is kind of a tangent for this issue but is something I think is super important. (By "superficial" I mean that you can parse its syntax correctly without executing a full TeX engine, which is *impossible* in full generality.)

None of the above justifies the use of the V2 layout in the first place, though. I'll point you to [this comment][x] which attempts to explain the vision behind the V2 approach. In a certain sense I believe that it is best that V2 documents are *not* easily put into existing LaTeX workflows, because in some very profound ways the whole reason that Tectonic exists is that it is *not* going to be maximally compatible with the existing ecosystem.

As for the more specific motivation, I want to (1) encourage people to separate their setup code and actual document text, with a vision of (1b) potentially being able to parse and auto-format the different files in different styles appropriate for code and marked-up prose, and (2) open the window for Tectonic to automatically traverse and compile whole trees of input files without needing explicit `\input` statements, which is a more natural fit for web-based outputs and I believe will open up some major usability wins for certain document types.

[x]: https://github.com/tectonic-typesetting/tectonic/issues/846#issuecomment-985576778

clbarnes:
[...]

Issue #846, opened 2021-12-02
tectonic -X new harmful 


dougransom:
Tectonic creates projects that won't compile with other tex programs.  This is going to drive users mad over the long run, there is no benefit, it can only create problems for us down the road.  

These three files are implicitly concatenated together by tectonic.  
`src/_preamble.tex
src/index.tex
src/_postamble.tex`

If there is some advantage to a workspace, why not just have index.tex import the other two files through some latex command?

I suggest  this behavior be changed immediately, to avoid users creating these projects with the implicit concentation, and providing an error message requiring a fix in index.tex when compiling existing projects.

pkgw:
OK — I'm going to attempt to convince you that this design choice is in fact the correct one.

Big picture: In a certain sense, the whole reason that the Tectonic project exists is to break compatibility with existing TeX engines. I believe that there are enormous usability/UX wins that can be unlocked if we allow ourselves to break certain things, and symmetrically that you *have* to break certain things in order to solve certain big problems. So, while I don't want to be gratuitously incompatible with the ecosystem, interchangeability with other TeX engines is an explicit *non*-goal — it's just too constraining. That's why I'm not just working to improve XeTeX.

Given that context, there's certainly some tension on the compatibility question. It's exactly the concern that you voiced: it's going to be unpleasant if people create documents using Tectonic with the expectation that they can be compiled with other engines, only to discover that they've used some functionality unique to Tectonic that prevents this. I want people to "opt in" to creating Tectonic-specific documents that can take advantage of unique features, but may not be compilable with other engines.

That's exactly what `tectonic -X new` is. While you could imagine a similar approach that instead prioritizes compatibility, the contract for this command (implicit only, I'll admit) is the new document is only going to be portable to other engines "at your own risk". Precisely because the `Tectonic.toml` "document model" approach is unique to Tectonic and so different than traditional TeX tools, I hope that it signals to people that their documents won't necessarily just drop into another engine successfully. The issue you're raising here suggests that it might be good to make this more explicit.

One day, it might be nice to have a `compatible` flag in `Tectonic.toml` that users could set to explicitly indicate that they want their documents to be compatible with other engines, causing errors to be raised if they use features unique to Tectonic. This would be associated with a requirement that the preamble and postamble files be empty. On the other hand, so much of "compatibility" is an emergent property coming from the hundreds of LaTeX class and style files that might underlie the document compilation, so that it's really hard to say with confidence whether a given document will compile as intended on another platform.

[...]

PR #1023, opened 2023-04-14
Remove preamble and postamble.

gtsiam:
I don't really expect this to be merged as-is, but I would definitely be happy if it was. I'm using my fork for now, but I figured making it into a pr would be a good way to start a discussion at the very least.

# Why

I _despise_ the idea of an implicit preamble/postamble. Not because of compatibility; I'm perfectly fine with breaking compatibility. But because I don't like implicit dependencies.

If I want to split the preamble into a separate file, I will. But I don't want to be forced to do it.

I know you've attempted to justify this choice [here](https://github.com/tectonic-typesetting/tectonic/issues/846#issuecomment-985576778) and [here](https://github.com/tectonic-typesetting/tectonic/issues/876#issuecomment-1074696932). While I agree with everything you've said in those posts, I don't agree with the solution. Maybe some form of templating (#881) could achieve what you're aiming, but a hard-coded preamble/postamble seems clumsy at best to me.

### DIFF
--- a/crates/docmodel/src/document.rs
+++ b/crates/docmodel/src/document.rs
@@ -21,23 +21,11 @@ use tectonic_errors::prelude::*;
 
 use crate::workspace::WorkspaceCreator;
 
-/// The default filesystem name for the "preamble" file of a document.
-///
-/// This default can be overridden on an output-by-output basis in
-/// `Tectonic.toml`.
-pub const DEFAULT_PREAMBLE_FILE: &str = "_preamble.tex";
-
 /// The default filesystem name for the main "index" file of a document.
 ///
 /// This default can be overridden on an output-by-output basis in
 /// `Tectonic.toml`.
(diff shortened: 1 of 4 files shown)


rm-dr:
Possible alternative [here](https://github.com/tectonic-typesetting/tectonic/discussions/1100).

pkgw:
Hi @gtsiam —

I really do appreciate your constructive engagement on this topic! I'm sorry that I've taken so long to actually reply to your PR. Due to various things going on with me, I've been really unresponsive with this project for far too long :-( I'm trying to at least give all of the open pull requests their due as best I can.

For the time being, I think I'll suggest trying to continue a discussion in #1100 as suggested by @rm-dr — as you might expect, I would prefer not to fully remove this feature, but it's definitely been a persistent sticking point. I always like to think that there's a way to meet in the middle that makes people happy, and hopefully we can converge on a design there.

rm-dr:
Closing in favor of #1123

PR #1031, opened 2023-05-09
Bump pinned vcpkg version to fix fallout of microsoft/vcpkg#30542

### LINKED ISSUE #1030: Out of date vcpkg causes msys2 download to fail
See https://github.com/microsoft/vcpkg/issues/30542 for context. In short, since around March, an old msys2 build was pulled from various mirrors. A fix was released, but requires a newer vcpkg revision than is pinned in the Cargo.toml.

CraftSpider:
Fixes #1030
Update to latest head revision - seems to build correctly locally.


pkgw:
Thanks!

PR #1032, opened 2023-05-09
Begin rustify engine_bibtex - convert helpers and statics

CraftSpider:
This starts by moving some of the helper functions into Rust, and rewriting them in terms of general good-practice for Rust code.

This is WIP because I'm hoping to move more things, but wanted to put it up for a general yea or nay reaction. Biggest roadblock right now is starting to push the absolutely giant number of global statics into Rust - still considering the best option for them, may try bundling some things up into structures and making it global but access only through helpers exposed from Rust.

(Also includes the change from #1031)


CraftSpider:
Note on aliasing: We can play a little fast-and-loose with it currently, due to the bibtex code only being able to run once at a time currently, so we can assume single-threaded execution. That means that in most situations, we can assume 'exclusive' access to a pointer - we should only break it if we call from Rust -> (C or Rust) -> Rust while using the same mutable reference both before and after the call. As stuff is converted I'll also probably make helpers safe as their call sites become all-Rust, with the goal of the final result being, if not perfectly good practice Rust, mostly safe, and with any unsafe properly documented and encapsulated.

pkgw:
Very cool, thanks for looking at this! I've always thought that the bibtex engine would be a good place to experiment with C-to-Rust porting since it's relatively simple and stable. My hope would be that a lot of the unsafe code and pointers can eventually go away, but that probably can't happen until things have been basically completely translated.

Can I answer any questions for you or offer any feedback on anything specific so far?

CraftSpider:
My only big question is whether you'd prefer that I make smaller PRs for individual components, or batch things up into larger changes? For smaller PRs, I could basically make the initial helpers one, move some statics in one, etc. This would be easier to review, but also may mean longer in that awkward middle ground where both ends are in a 'kind of iffy' state - see the buffer changes, which I'm not super happy with the verbosity on the C side, but will be really nice on the Rust end once it's more moved over.

CraftSpider:
New bibtex tests immediately caught a couple errors introduced at various points - serving their purpose admirably. I may want to add more later, to test more obscure code paths.

CraftSpider:
Well, this is gonna be something to get reviewed, but I'm really happy with how the code is starting to turn out - the top-down movement means that most of the extern functions are actually safe at this point, with the minority of unsafe ones being those that get passed pointers to values still in C.

The ownership isn't actually too bad to work with - it could do to be cleaned up after the transition, but fundamentally, it mostly follows a pattern of using RefCell and `with` functions at the C-to-Rust boundaries, and inside those functions we can just write normal well-borrow checked Rust, then bubble information out of them to the C code.

The ugliest bit so far is handling `longjmp` - it's UB to `longjmp` across Rust code, so I needed to write some temporary `unwrap` functions in C that take tagged unions generated by bindgen and return the value or jump. That will go away once everything is done - it can all just become a normal `Result`.

It feels like it's kind of getting faster to move stuff - as more of the base is complete, more and more stuff becomes a good candidate for porting.

pkgw:
That's all good to hear! I hope to sit down and read over the code in its current state at some point this weekend, although I may end up deciding to stay away from the laptop instead.

You probably saw that I released a bunch of crate updates involving your work on Tectonic's dependencies. Thanks again for your contributions on that front!

CraftSpider:
I'll be pushing a commit or two soon that do some cleanup, add some tests and documentation, etc - since most all of the statics are moved, I think this is a good point to start looking at making this mergeable. Doing it all in one go would be nice in some ways, but this is already probably a bit big, and with just a bit more work I think won't be too ugly on either side.

I've said this for many PRs before, but I think this one it may be most relevant, it might help to go commit by commit - not every one does it, but I largely tried to make each potentially stand on it's own, and mostly focus on specific areas.

CraftSpider:
Here's an interesting question - if there's an explicit warning that sorting may be unstable without a certain key, do you think it's okay to change the behavior? I've tried to be careful to exactly preserve bibtex's existing behavior, but this seems like a weird case. I 100% don't mind putting in the little bit of extra work to port the sorting exactly, but I think it would be nice to just use Rust's built in sorting at some point in the future (I've been considering an architecture after a full port that would allow making certain behaviors 'pluggable' also, so there could be a flag that enables/disables an 'exact compat' mode, then we could improve diagnostics or provide minor changes to functionality if they're judged better, but fall back to just doing what bibtex does)

pkgw:
Hmmm ... I think that for now, it would be best to preserve the existing sort semantics if it's not too inconvenient. The fewer externally-observable changes induced by this kind of PR (ideally none, but I don't know how realistic that is), the better, I think. That makes it easier to review the results of things like the Tectonic-on-Arxiv battery.

In principle, once an oxidized implementation gets "settled in", I wouldn't mind altering behaviors if it cleans up the implementation in some substantial way. But if it's basically a matter some extra lines of code needed to preserve the compatibility of outputs, I'd probably lean towards saying that we might as well leave the code in once it's been written.

src/unstable_opts.rs
```diff
@@ -107,7 +107,7 @@ pub struct UnstableOptions {
     pub continue_on_errors: bool,
     pub paper_size: Option<String>,
     pub shell_escape: bool,
-    pub min_crossrefs: Option<i32>,
+    pub min_crossrefs: Option<u32>,
```
CraftSpider, line 110: I can revert this change if you want - I figured since they were 'unstable' it might be okay to swap out the type, and negative minimum values don't make sense anyways - they'd all just be equivalent to `0`

pkgw: Yeah, this is fine to change.

crates/engine_bibtex/src/c_api/char_info.rs
```diff
@@ -0,0 +1,217 @@
...
+    // right brace
+    id_class[125] = IdClass::IllegalIdChar;
+
+    id_class
+};
+
+#[no_mangle]
+pub static CHAR_WIDTH: [i32; 256] = {
+    let mut char_width = [0; 256];
+
```
pkgw, line 119: It might be nice to have a comment here explaining where these magical values come from (even if it's just "these magical values come from the bibtex web2c implementation").

crates/engine_bibtex/src/c_api/exec.rs
```diff
@@ -0,0 +1,248 @@
+use crate::c_api::cite::with_cites;
+use crate::c_api::hash::with_hash;
+use crate::c_api::history::mark_error;
+use crate::c_api::log::{print_a_pool_str, print_confusion, write_logs};
+use crate::c_api::pool::{bib_set_pool_ptr, bib_set_str_ptr, bib_str_ptr, bib_str_start};
+use crate::c_api::xbuf::{xrealloc_zeroed, SafelyZero};
+use crate::c_api::{GlblCtx, HashPointer, StrNumber};
```
pkgw, line 7: For `use` sequences like this, I like to use nested curly braces, i.e. `use crate::c_api::{cite::with_cites, ...}`.

crates/engine_bibtex/src/c_api/log.rs
```diff
@@ -0,0 +1,715 @@
...
+    LOG_FILE.with(|out| {
+        let mut log = out.replace(None);
+        let res = f(log.as_mut().unwrap());
+        out.set(log);
+        res
+    })
+}
+
+pub(crate) fn write_logs<B: ?Sized + AsBytes>(str: &B) {
+    with_log(|log| log.write_all(str.as_bytes())).unwrap();
```
pkgw, line 74: Would these `unwrap()s` potentially panic on an I/O error?

crates/engine_bibtex/bibtex/bibtex.c
```diff
@@ -7,1929 +7,166 @@
...
-static bool scan_white_space(void)
-{
-    while ((buf_ptr2 < last) && (lex_class[buffer[buf_ptr2]] == 1 /*white_space */ ))
-        buf_ptr2 = buf_ptr2 + 1;
-    return buf_ptr2 < last;
-}
+/*:159*//*160: */
 
-static bool eat_bst_white_space(void)
+static bool eat_bst_white_space(GlblCtx* ctx)
```
pkgw, line 120: This is very much not a big deal, but I don't love the mega-abbreviated name `GlblCtx` ... maybe just `Bibtex`, or `Engine`, or something like that? Or just `GlobalCtx` would be a bit nicer to my tastes.

pkgw (commented):
OK, I've sat down and read over the current status of the PR here. This is looking fantastic! Everything seems delightfully clear and well-organized. I definitely have *not* read over to check whether every single line is translated correctly, but everything that I've checked has seemed correct.

I have a few quite superficial questions and comments below, but overall this PR looks like it's heading in a great direction. I will try to also take some time to sit down and ponder whether there are any big-picture issues to dig into here, but so far there isn't anything making me concerned.

Do you have a sense of what it would take to get this PR to a point where we could consider merging it? Looking at the amount of code left in `bibtex.c`, it seems like there's still quite a lot to be done before the engine is 100% oxidized, but there's also obviously a ton of work already in this PR, and I'd be excited to merge in a "step 1" refactoring even if it's short of the 100% point. As I've mentioned before, in my ideal world such a refactoring would leave the external behavior of the engine completely unaltered (as checked by, e.g., the tectonic-on-arxiv test), but I wouldn't be surprised if that goal will be unattainable.

crates/engine_bibtex/src/c_api/log.rs
CraftSpider: Yes, I think (memory, might be wrong) bibtex may have just ignored errors here. I can do any of the following:
- Ignore errors, exactly match bibtex
- Bubble up errors to match other bibtex error handling (may cause other issues if that causes more writes, probably not what we want here)
- Keep this behavior - panic instead of silently dropping missed writes

crates/engine_bibtex/src/c_api/char_info.rs
CraftSpider: Adding a comment for this, and also fixing it up to use the `b''` character syntax in most places - lets me remove some comments, and generally makes the intent a lot more obvious.

CraftSpider:
I appreciate the review. As for getting this to review - I think the current state wouldn't be a bad first step. It moves all the statics, aside from the jumpbuf items, and basically all the helper methods up to the line of the BST/AUX reading and execution, which I think makes for a fairly natural boundary for a first step. If this was merged as a first step, I'd likely start on the next step fairly quickly, but I think it would be nice to get this out there and let more changes come later.

In terms of behavior - it depends exactly what you mean by '100% identical'. I think it's very feasible to preserve exact byte-for-byte output - I've made an effort so far to ensure that the final state of logs and files is exactly the same as it would be with the original implementation, and I'd consider any differences to be bugs in my port. On the other hand, I can't guarantee the exact same system calls, so it theoretically may be possible to observe differences in log flushes or allocations. I wouldn't normally expect such things to be considered 'observable' for users to rely on, but I don't know the exact requirements of the tectonic-on-arXiv tests.

crates/engine_bibtex/src/c_api/log.rs
pkgw: I don't think that panicking is a desirable behavior here. It might be possible to bubble up the errors, but I have a strong suspicion that implementing that would be a big pain and could potentially introduce a recursion issue as you suggest. So I think that the least-bad approach here is to "handle" any errors within the function. I wouldn't love using `eprintln!()` to report any error, but it would be better than nothing, I think.

CraftSpider:
I'll take a look into it - I actually appreciate the tests, as this provides a bunch more real-world test cases I can compare against. I can immediately see a couple easy to fix issues where I did something like replace \`' with \`\`.

crates/engine_bibtex/src/c_api/log.rs
CraftSpider: I'm mostly worried with `eprintln` that it would just panic as well - if the log file fails it might work to report that on stdout, but if the stdout writing fails, it seems a bad bet that stdout writing will succeed. I know tectonic might 'fake' stdout through OutputHandle, but I think it's a fair assumption for a program to make that stdout is stdout.

CraftSpider:
Alright, after those 2 quick fixes we're down to 32 changes. At least half of them are definitely errors, but tracking them down is going to be a bit more involved, since they're largely output-only issues, or just flat out crashes.

CraftSpider:
Down to 24 errors, 10 of them related to bibtex output, all of those being some kind of crash with no files emitted, not even a log file. It's always the same packages, so it's deterministic, but I haven't yet figured out a good reproduction. I suspect they're all related in some way though, so hopefully once I figure it out, we'll be down to the 14 synctex failures.

crates/engine_bibtex/bibtex/bibtex.c
```diff
@@ -5480,1554 +3615,1496 @@ static void get_aux_command_and_process(void)
...
         {
-            num_cites = cite_ptr;
-            num_preamble_strings = preamble_ptr;
+            set_num_cites(cite_ptr());
+            ctx->num_preamble_strings = preamble_ptr();
             {
-                if ((num_cites - 1) * num_fields + crossref_num >= max_fields) {
+                CiteNumber cites = num_cites();
+                if (cites > 0) {
+                    cites -= 1;
```
CraftSpider, line 4738: This bug is like, a micro-example of two of the big pros of Rust here

A) this would have caused a clear panic instead of silent underflow on debug mode, making the problem immediately obvious
B) There's a dedicated method on integers, `saturating_sub`, which does this but shows the intent a lot better

pkgw: Word! Once you get used to the stricter behavior of a language like Rust, I feel like it becomes more and more amazing that we all put up with C's anything-goes style for so long.

CraftSpider:
I really hate Windows sometimes - you're not allowed to have a file or directory named AUX (among some other reserved names), and since we deal with `.aux` files, I keep trying to use that name

CraftSpider:
tectonic-on-arXiv is now down to only the pre-existing synctex failures - I wouldn't say that there aren't any bugs, but in a wide swath of situations, this branch now produces byte-for-byte identical output to master. All the fixes were relatively simple luckily, just minor typos or mistakes made while porting.

pkgw:
Awesome!! I will give this PR another look-over from the perspective that it may potentially be ready to merge. As ever, turnaround time will probably not be as quick as I'd like. Feel free to give me nudges as needed.

pkgw:
I'm working through my backlog today and planning to merge this PR soon! It will need to be rebased onto the current `master` to get a CI fix, which I was also planning to do once the other outstanding PRs are dealt with.

CraftSpider:
I just did the rebase, was waiting to see how CI did. Hoping for the arXiv tests to pass now, I think it should work now but if anything fails it would be there.

Mrmaxmeier:
The arXiv tests should be fine, but if you want to be sure you can push an empty commit or close and reopen the PR. The arXiv bot was offline for a bit so it missed your last push and I don't have a convenient way to manually trigger CI. :slightly_smiling_face:

pkgw:
I don't think this really affects anything here, but FWIW I've been planning to merge #1043 first since it should not affect the ArXiv testing pipeline at all, so hopefully it should provide a nice baseline check that the CI is working. So probably another hour or so until I retrigger this as the other CI builds process.

pkgw:
Never mind, the other one just finished. The merge caused some spurious CI errors here so I went ahead and pushed a new commit merging the new `master` in here.

pkgw:
tectonic-on-arXiv successful!

ncihnegn:
This is the cause of #1105.

PR #1036, opened 2023-05-14
Introduce `-Z deterministic-mode` to address a few reproducability issues

Mrmaxmeier:
I've dusted off `tectonic-on-arXiv`, and I looks like upgrading to the TeXLive 2022 bundle uncovered new reproducibility issues:
- Documents that request synctex files with `\synctex=1` contain absolute paths to the TeX files. ~~This seems intentional, so I've left absolute paths enabled for `tectonic --synctex` runs.~~
- The `hyperxmp` package embeds a bunch of metadata into the output PDF. In particular, it derives a random "xmpMM:InstanceId" from a bunch of information sources. One of those is the input file modification timestamp. I've wired this up to be the document build date instead. ~~I'm not sure if there's a "legitimate" use case for file modification timestamps, so let's just try always returning the build date instead. :slightly_smiling_face:~~

=> https://tt.ente.ninja/#/compare/05d445d074674bf1da7dd277bc8524fe85388ea9-untrusted-run-1/05d445d074674bf1da7dd277bc8524fe85388ea9-untrusted-run-2

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -122,7 +122,7 @@ tempfile = "^3.1"
 
 [package.metadata.vcpkg]
 git = "https://github.com/microsoft/vcpkg"
-rev = "ea222747b888b8d63df56240b262db38b095c68f"
+rev = "63366443439398a62afc9a63b34b9a3ba63b1cae"
 overlay-triplets-path = "dist/vcpkg-triplets"
 
 # If other targets start using custom triplets like x86_64-pc-windows-msvc,
--- a/crates/bridge_core/src/lib.rs
+++ b/crates/bridge_core/src/lib.rs
@@ -254,7 +254,7 @@ impl<'a> CoreBridgeLauncher<'a> {
         self
     }
 
-    /// Ditto for file modification timestamps. In reproducible mode, we return
+    /// Ditto for file modification timestamps. In deterministic mode, we return
     /// the configured build time (i.e. `SOURCE_DATE_EPOCH`) instead of the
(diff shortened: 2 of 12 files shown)


Mrmaxmeier:
Right, I agree that heuristics for "reproducible mode" are a bit misguided. There's merit in keeping the default reproducible, but that's not really realistic as `\today` should and does default to the current date.

I'll introduce a `--reproducible` (or maybe `--deterministic`?) flag instead. I thought about coupling the reproducible mode with the `SOURCE_DATE_EPOCH` env var, as it's generally used in environments where reproducible builds are desired, but I think having it be an explicit flag is a better idea.

Is this something that should live in the "unstable options" world? It's certainly niche, but the interface (reproducible mode on/off) seems reasonably stable :)

Note: tectonic-on-arXiv will break for a while until this PR is merged as I'm now passing `--reproducible` as a CLI flag.

---

Tectonic.lock is a cool idea, though I'm not sure where the line between "committed to .lock" and "part of the document context" is. (i.e. the document's TeX sources shouldn't be part of the lockfile, and the bundle should be, but what about external paths / fonts / results of write18 invocations?)

Mrmaxmeier:
Yep, deterministic mode generates broken synctex files. I've added a note in the unstable options help text that discusses this.

pkgw:
I finally have the chance to sit down and review this. This is looking excellent! I just pushed a commit to update vcpkg, which should hopefully get the build all green again.

PR #1037, opened 2023-05-15
Add more bibtex tests

CraftSpider:
These test some simple error cases, I plan to rebase #1032 on this so there are more tests available to avoid regressions

### DIFF
--- a/crates/engine_bibtex/bibtex/bibtex.c
+++ b/crates/engine_bibtex/bibtex/bibtex.c
@@ -533,11 +533,12 @@ input_ln(peekable_input_t *peekable)
     // For side effects - consume the eoln we saw
     int eoln = peekable_getc(peekable);
 
-    // Handle \r\n newlines on Windows by trying to consume a \n after a \r, unget if it's not that exact pair
-    int next = peekable_getc(peekable);
-    printf("Next Char: %d\n", next);
-    if (!(eoln == '\r' && next == '\n')) {
-        peekable_ungetc(peekable, next);
+    if (eoln == '\r') {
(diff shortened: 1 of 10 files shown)


CraftSpider:
Well, apparently, the existing code fails on Windows. It doesn't fail locally on Linux or Windows, which makes me think *somehow* the code is invoking UB, except I think it's UB in the *C* code, so no clue where it is yet. I'll spend some time trying to figure this out.

crates/engine_bibtex/bibtex/bibtex.c
```diff
@@ -529,8 +529,16 @@ input_ln(peekable_input_t *peekable)
...
     }
+    
+    // For side effects - consume the eoln we saw
+    int eoln = peekable_getc(peekable);
 
-    peekable_getc(peekable);
+    // Handle \r\n newlines on Windows by trying to consume a \n after a \r, unget if it's not that exact pair
+    int next = peekable_getc(peekable);
+    printf("Next Char: %d\n", next);
+    if (!(eoln == '\r' && next == '\n')) {
```
pkgw, line 539: Two items here:

1. Looks like there is a debug `printf` that should be removed.
2. I'm not sure if it would be any more efficient or anything, but I think I'd rather have the code here only attempt the second getc if the first character is `\r`. I think that makes the internal logic a bit clearer.

pkgw (commented):
Thanks for working on this! I have a couple of requested changes that ought to be straightforward to implement, if you don't mind.

pkgw:
Thanks! Looks like we had a spurious test failure here, I think that we should pass with a retry.

(The tectonic-on-arXiv test will just raise some failures incorrectly for a bit until we straighten out some reproducibility subtleties; see #1036)

PR #1038, opened 2023-05-15
Update non-watchexec deps

CraftSpider:
An intermediate PR while I figure out #1033 - update all but watchexec to latest versions and run `cargo update`. This still includes a significant update for `quick-xml` to remove one of the two forward-compat warnings.


pkgw:
Thank you!

PR #1039, opened 2023-05-19
Add a bunch more bibtex tests

CraftSpider:
This includes tests for crossrefs, multiple cites, and some bad aux lines. This should cover most of the important main code paths as well as some particularly useful warning/error paths.


CraftSpider:
These tests aren't *just* for my bibtex rewriting work - this batch catches another real error due to `cite_xptr` never getting set to `0` on init, which I *suspected* when I noticed it in rewrite, but running these tests confirms it, as it cause run-order dependent failures.

CraftSpider:
Error appears to be that on Linux, whatever opening a directory returns, it doesn't cause issues - but on Windows, you get `PermissionDenied`, which means it returns an error instead of trying with the extension. I'm just gonna move that test out of the directory for now, since it's a one-off anyways.

pkgw:
Great stuff! It looks like the main CI tests passed for the most recent iteration, but GitHub is only showing my the ArXiv check. I'm going to close this and then reopen to retrigger the CI just for completeness.

CraftSpider:
Yeah, from what I could tell, I pushed twice too close and the tests ran on the second commit but got attached to the first. Either way, looks happy now.

PR #1051, opened 2023-06-11
Rustify engine_bibtex stage 2 - begin porting main code

CraftSpider:
Already removed a couple hundred lines from the C code, just since I started this ~2 weeks ago. Opening this to start getting CI runs so I can ensure no behavior changes are being made.


CraftSpider:
The more of the scanning code I port, the more I want to rewrite this to use a more standard separation of AST and Interpret steps, so I can use a parser library like [chumsky](https://github.com/zesterer/chumsky) (disclaimer: I am a primary maintainer for chumsky) to turn these hundreds of lines of repetitive handwritten parsing into some shorter and easier to read, and 10x easier to maintain too.

CraftSpider:
@pkgw this is probably about ready to merge, modulo review and such - it gets everything up to the `x_*` functions, and for those, I'm working on a wide-reaching refactor to make `ExecVal` a Rust enum, and in general, I think this makes for a good enough stopping point.

pkgw:
The tour de force continues! While I definitely don't understand this piece of the codebase as well as you at this point, this looks like wondering, careful work, and it's very reassuring to have the test suite and @Mrmaxmeier's ArXiv battery passing. Thank you for your contributions!

Issue #766, opened 2021-04-18
Proposal : allow GitHub discussions


kpym:
IMO, many of the issues here are not real issues, but rather questions or discussions. So maybe allowing [GitHub discussions](https:/docs.github.comendiscussions) will keep only the "real" issues here. 

I know there is already a [tectonics forum](https:/tectonics.newton.cx), but switching to GitHub discussions will bring everything to the same place with all the advantages, especially the ability to move a topic between the issue tracker and the discussion forum.

pkgw:
Thanks for the proposal! I have to admit that the Tectonic forum has not taken off like I'd hoped and I don't do much to maintain or support it right now, so that's another thing to factor in.

I don't have any experience using GitHub Discussions although I'm familiar with the general concept. I'm certainly willing to give it a try but would be interested to hear folks' opinions about how well (or not-well) it works. Does anyone else have any comments on the topic?

kpym:
@pkgw Just to say that I have no experience with Github Discussions (it is relatively new) but it looks like to be very close to the issue tracker, so if somebody complains about it, it should complains about the GitHb issues too ;)

pkgw:
@gabestein You've made exactly this transition, no? Any thoughts about how it's gone so far?

gabestein:
Sorry I'm late! It's been great, honestly. It's done three things nicely for us:

1. Helped keep Issues, which we use for our sprint process, uncluttered.
2. Simultaneously decreased the barrier to entry for folks who found it intimidating to create an issue, particularly for things that were more ideas or questions than bugs/enhancements.
3. One less login/site to remember for folks to get help/share ideas/etc. We had a Discourse forum before, but it wasn't used all that much. The GitHub forum is (surprisingly and delightfully!) highly trafficked.

pkgw:
It took me a while but I've gone ahead and turned on discussions:

https://github.com/tectonic-typesetting/tectonic/discussions/878

Let's see how it goes!

goyalyashpal:
> Let's see how it goes!

@pkgw how are you feeling about it now after more than a year has been passed? 👀 💭

PR #1068, opened 2023-07-21
Update forum links

ColeFrench:
It seems from #766 and #878 that the old forums are being sunsetted, so this PR completes the cutoff from them.


pkgw:
Thanks, this kind of maintenance is much appreciated!

PR #1076, opened 2023-08-25
Make file expectations always check all expected files at once

CraftSpider:
This keeps bothering me while debugging bibtex failures, and just flipping the file order there would work, but it would be a band-aid, not a real solution. So here's a real solution - make all expectations test once, and panic once at the end.

(Also removed `genio` from bibtex tests, as it just confuses the output and all the useful logs go to both stdout and the file, IME)

### DIFF
--- a/crates/bridge_core/src/lib.rs
+++ b/crates/bridge_core/src/lib.rs
@@ -1227,7 +1227,7 @@ pub unsafe extern "C" fn ttbc_diag_append(diag: &mut Diagnostic, text: *const li
 #[no_mangle]
 pub extern "C" fn ttbc_diag_finish(es: &mut CoreBridgeState, diag: *mut Diagnostic) {
     // By creating the box, we will free the diagnostic when this function exits.
-    let rdiag = unsafe { Box::from_raw(diag as *mut Diagnostic) };
+    let rdiag = unsafe { Box::from_raw(diag) };
     es.status
         .report(rdiag.kind, format_args!("{}", rdiag.message), None);
 }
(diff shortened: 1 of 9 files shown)


CraftSpider:
Quick thought: Would it be possible to make CI prioritize, say, `fmt`, `clippy`, and one of the common/fastest platform tests? Since those seem to be the most likely to fail, might help with turnaround of realizing things are broken on a PR.

pkgw:
@CraftSpider The main tests should all run in parallel, so in principle there shouldn't be a wait ... but maybe if we hit the limit for simultaneous jobs, and the dispatcher files them in lexical order, there would be an issue? You could try editing `dist/azure-build-and-test.yml` to put things like Clippy at the beginning of the `jobs:` section, and see if that changes things? I think that if you do it in this PR, the tests here will be run with that setup.

CraftSpider:
Sounds like it's worth a try - I consistently see only some running, I kind of figured it was some max concurrency limit.

pkgw:
Did the reordering of the Azure Pipelines file seem to help?

CraftSpider:
Yep, it appears putty clippy and rustfmt at the top gave them more 'priority' in the queue - they tend to now be the first to finish, which is nice. 
(I could run clippy locally, but I enable a bunch of extra lints through a dev-env global config for personal projects, and haven't yet gotten around to making tectonic ignore them. Also I just forget to run it before pushing a lot :P)

CraftSpider:
Want to note, since it isn't in any of the comments on this PR: This removed the cross-mips build, due to the recent Rust decision to downgrade these platforms to T3, and consequent removal of host-tools builds for them causing CI to fail. If they are ever re-upgraded to T2 (could happen if someone steps up as a maintainer) the build can be easily re-added.

pkgw:
I did notice that, thanks. This means that we lose our big-endian test build, which is why I added MIPS in the first place. That's too bad but it doesn't seem to exactly be a major use case these days, and hopefully the existing code won't bit-rot on us.

PR #1077, opened 2023-08-29
Rustify engine_bibtex - part 3, bibstyle interpreter

CraftSpider:
Working my way through the main interpreter loop - this includes a significant refactor to start turning some data structures into Rust types, as more code moves to Rust I'm hoping to start using more Rusty good practice. I tried to also change `HashData`, but that's still a little much work on the C side to be worth it. OTOH, I'm getting ideas as I go for better APIs, and at some point may revisit older ported code and use some of the newer APIs I've introduced as I notice common patterns.


CraftSpider:
tectonic-on-arXiv is one of the most useful test suites I've used - it's relatively fast for allowing testing against such a massive bank of real-world use-cases, making me much more confident in correctness than the current quick tests. Those aren't *bad*, and they help with local checks before I push that my code basically works right, but the arxiv test finds all the little edge-case bugs I don't think of.

pkgw:
Indeed! Thanks as always to @Mrmaxmeier for setting it up. I shudder at the thought of trying to manually assemble a test suite that covers all of the corner cases exposed in the Arxiv corpus.

Mrmaxmeier:
I'm glad tectonic-on-arXiv is useful. :smile:
There surely is a lot of functionality not covered by the arXiv suite, but I hope it still provides a certain sense of _good enough_ for these kinds of rewrites ^^

CraftSpider:
This should be ready for review (module the final job passing), with a *very* satisfying final commit that removes a whole lot of exposed stuff, moving a significant chunk of logic into just-Rust territory. This should only need 1 or 2 more PRs likely, to move everything into Rust. After that will likely be some cleanup, trying to make some of the code more 'Rusty' and easier to maintain. Then it's on to another project - I know @pkgw you mentioned wanting xetex-layout ported, so I'll probably move to that after I'm satisfied enough with Bibtex.

pkgw:
Great! I'll give this a review.

If you might be interested in a very different sort of project, I've become convinced that it would be very valuable for Tectonic to offer integration with some kind of per-project package management system so that documents can be built using a stable tooling environment — the most obvious application being the endless issues of aligning versions between the `biblatex` TeX files and the `biber` program (which is written in Perl and so can't be bundled conveniently). Other useful applications like TeX-to-SVG could benefit from relying on other programs that are external to the core engine.

As it happens, some folks just released a tool based on Conda called Pixi that includes a backing library crate [rattler][1] that seems like it could be a good match. Conda supports Mac/Windows/Linux and has a robust ecosystem. I envision some kind of keyword in `Tectonic.toml` like `environment = "pixi"` when then basically triggers Tectonic to ensure that the relevant environment is "activated" before doing any compilation. Or you could autodetect the presence of a `pixi.toml` file alongside `Tectonic.toml`.

[1]: https://github.com/mamba-org/rattler

Anyway, if you're interested in focusing on Rustify-ing things, that's also awesome! I've just recently had this idea and I  think that it would make a lot of people happy, but lately at work I've had to lead up some new development efforts which is leaving me with very little energy for coding after-hours these days.

CraftSpider:
The idea seems really neat - I'll look into it. Honestly I started contributing to the oxidize effort in part for 'selfish' reasons - I used tectonic in a project, and the most annoying bit for me getting it set up was getting it to build as a dependency. On the other hand, I'm really liking working on the project, and that seems like a nice thing to have - a way for Tectonic to integrate with an 'environment', which it's already doing better than other Tex stuff I've used (ugh, every time I install any tex distribution on a new computer I want to pull my hair out), but I can imagine a world where it's even easier than today - you specify in `pixi.toml` what outside tools you want to integrate, and it 'just works'. I'm having more thoughts about such a thing, but that should probably go to a discussion thread. Anyways, I'll probably want to 'finish' (for a certain value of the word) bibtex, then I'll look into which project to do next.

crates/engine_bibtex/src/c_api/exec.rs
```diff
@@ -1,367 +1,285 @@
...
-        if res != CResult::Ok {
-            return res.into();
-        }
+        with_cites(|ci| rs_print_a_pool_str(ci.get_cite(ci.ptr()), pool))?;
     }
 
     write_logs("\nwhile executing-");
     // SAFETY: The `Bibtex` value pointed to by `ExecCtx` is guaranteed valid
-    bst_ln_num_print(unsafe { &*ctx.glbl_ctx })?;
+    bst_ln_num_print(unsafe { &*ctx.glbl_ctx }, pool)?;
```
pkgw, line 231: Could constructs like this be replaced with `ctx.glbl_ctx()` (or `.glbl_ctx_mut()`) now?

crates/engine_bibtex/src/c_api/exec.rs
```diff
@@ -779,35 +649,1835 @@ fn rs_check_command_execution(ctx: &mut ExecCtx) -> Result<(), BibtexError> {
...
+                                        buffers,
+                                        pool,
+                                        hash,
+                                        entries,
+                                        globals,
+                                        bibs,
+                                        cites,
+                                        other,
+                                    };
+                                    rs_execute_fn(&mut *ctx, &mut globals, ex_fn_loc)
```
pkgw, line 2430: Amazing.

pkgw (commented):
You can definitely tell that this is turning a corner! The biggest comment I have regards a small block of commented-out code.

crates/engine_bibtex/src/c_api/exec.rs
CraftSpider: My next MR makes this look a little nicer - though really, soon all the `with_*` methods will go away entirely.

crates/engine_bibtex/src/c_api/exec.rs
CraftSpider: It should be able to, at this point, I'll do a cleanup to replace the pointer derefs with them.

CraftSpider:
It is not, just a mistake. I'll revert it.

pkgw:
I think this needs an update to vcpkg, which I just put in in #1084. I believe that Azure tests with the PR branch merged to master, so a close/reopen to restart the CI might suffice here?

Mrmaxmeier:
Hmm. The `tectonic-on-arXiv` bot doesn't handle this correctly. It notices that each commit has already been tested and just cancels the job instead of reporting the known results. As a workaround, you can check the test run results by plugging the commit hashes into the web interface: https://tt.ente.ninja/#/compare/cc83b75b8ab3358f4a68773edbe1101a8dcfaaab/c2f16936967f2dbcb98d4587fa9fe4a0259f1785

pkgw:
Well, bot corner cases aside, it looks like we are now happy here! :rocket:

Mmagkooo (approved):

PR #1092, opened 2023-09-17
Update CI builds to Ubuntu 22.04 / OpenSSL 3

pkgw:
By updating the Linux builds to be based on Ubuntu 22.04, this should make the resulting binaries start linking against OpenSSL 3, instead of the end-of-life OpenSSL 1.1.x series. To be confirmed upon analysis of the built artifacts.


pkgw:
Looks like there is some new Windows CI issue to work through, but confirmed that the `x86_64-unknown-linux-gnu` artifact is now linking against `libssl.so.3`.

PR #1101, opened 2023-09-29
Check for lowercase "tectonic.toml"

rm-dr:
**Summary:**
With this change, Tectonic will behave a bit like make, accepting both `Tectonic.toml` and `tectonic.toml`

**Why we need this:**
TeX sources usually have lowercase file names, and `Tectonic.toml` looks a bit odd when every other file in the tree is lowercase.

### DIFF
--- a/crates/docmodel/src/workspace.rs
+++ b/crates/docmodel/src/workspace.rs
@@ -66,18 +66,33 @@ impl Workspace {
         let mut root_dir = initial_dir.clone();
         root_dir.push("tmp"); // simplifies loop logic
 
+        static TOML_PATHS: &[&str] = &["Tectonic.toml", "tectonic.toml"];
+
+        let mut doc_file: Option<fs::File> = None;
         while root_dir.pop() {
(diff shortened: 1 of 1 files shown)


pkgw:
Thanks for contributing yet another idea to the project!

Speaking for myself, I think that I prefer the current behavior of only recognizing the capitalized `Tectonic.toml` file. As far as I can see, this basically comes down to a stylistic choice, so I'm not sure that I can offer any rigorous arguments one way or the other. Besides being consistent with something like `Cargo.toml`, I think that the single capitalized letter does a good job of calling out the file as being special-ish, without going all the way to all-caps like a `README.md`. More than that, I'd say that in terms of engineering, I definitely think that in general it's good to define a single way to do things when it's not necessary to offer multiple ways. For instance, my opinion is definitely that it would be better if `make` only looked for `Makefile` and didn't also offer `makefile` (and `GNUmakefile` for GNU Make).

Anyone else have thoughts on this topic?

Issue #737, opened 2021-01-16
V2 cli: add external commands


jeffa5:
Similarly to how cargo searches `PATH` for binaries named `cargo-*` and treats them as subcommands with args passed through it would be cool to have this ability in tectonic, searching for `tectonic-*`, to allow other tools to be created quickly to experiment on new ideas.

pkgw:
Yup. I've got analogous [code in Cranko][xref] that could be used as a template.

[xref]: https://github.com/pkgw/cranko/blob/master/src/main.rs#L831

4e554c4c:
This issue should be closed, no?

PR #1103, opened 2023-10-04
V2 CLI: Add external commands

rm-dr:
Implemented the feature proposed in #737.
Is it really that easy?

**A question:**
Should `v2_main` exit earlier if an external command is given?
We probably don't need to set up colors and customizations if we're running an external command.

### DIFF
--- a/docs/src/ref/v2cli.md
+++ b/docs/src/ref/v2cli.md
@@ -40,6 +40,14 @@ nextonic build
 You can use various Unix `argv[0]` hacks to achieve this, or you can just rename
 or symlink the `tectonic` binary to `nextonic` manually.
 
+
+## External tools
+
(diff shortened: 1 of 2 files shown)


ColeFrench:
Cool! I think you should add tests with a dummy external subcommand, and document this feature.

pkgw:
Thanks! I'm going to close and reopen this to rerun the CI to check that this still builds cleanly against the latest `master`.

pkgw:
Agreed that it would be good to add test coverage for this, although I admit that it might take a fair amount of work to implement for relatively small benefit. It might be helpful to see if Cargo has a clever, lightweight way for testing this kind of functionality.

One further ask that's a little bit more important to me would be updating the documentation somewhere to mention this capability. I'm not sure if there's a single best place in the docs to describe it, but ideally it should be described somewhere.

pkgw:
The Azure Pipelines CI failures are spurious — a side effect of me merging another PR while this one was running. They should go away with the next fresh CI build.

PR #1106, opened 2023-10-17
Extra paths for shared resources

### LINKED ISSUE #985: Handle shared resources for multiple files
To give some context for this matter, allow me to describe my situation. I'm trying to write up a series of laboratory guides for students. Maybe someday these are compiled as part of a book, but for now, they're just a series of files using a shared format (really just a shared header and bib file). Part of this setup also sets the font to use the STIX2 text and math fonts (and I'd like to use Inconsolata or Fira Code as the monospace font). 

Some of this matter was discussed in issue #933, and I was able to work around a bit of it by using the `shell_escape_cwd = ".."` keyword (shared header and bib file now work). However, this litters the root directory with temporary TeX files specific to each lab report. Cleaning isn't that big an issue, but it is an _additional_ issue spawned by this workaround. What I can't get to work is using the `fontspec` package to specify the STIX2 fonts. This used to work, but now I get errors like 

``` 
warning: open of input /Users/<username>/Projects/357_experiment_guides/guide_src/shared/STIXTwoMath-Regular.otf:0-UCS32-Add failed
caused by: access to the path `/Users/<username>/Projects/357_experiment_guides/guide_src/shared/STIXTwoMath-Regular.otf:0-UCS32-Add` is forbidden
```

These fonts are installed by the system, but trying to use the system locations results in the same issue. I placed them in the shared resources folder in my project in an attempt to circumvent this problem, but was unsuccessful. 

Any suggestions on a better way to proceed are welcome. I love the compilation speed of Tectonic and would like to incorporate all of this into a reproducible resource for others, but I certainly can't release it in the current state.

rm-dr:
For context, see the discussion in #985.
Closes #985.

## TLDR:
This PR adds an `extra_paths` config key that allows a user to specify external resources.
An example file tree and `Tectonic.toml` are below.

```toml
[doc]
name = "Test"
bundle = "https://data1.fullyjustified.net/tlextras-2022.0r0.tar"

# New config key
extra_paths = [
    "../../resources"
]


[[output]]
name = "main"
type = "pdf"

```

```
workspace-repo-root/
├── resources/
│   └── class files, etc      <-- All documents pull files from here
│
├── Subdir1/
│   ├── this document/
│   │   ├── src/
│   │   └── Tectonic.toml     <-- this file is above
│   └── more documents...
│   
└── Subdir2/
    └── and so on...
```

### DIFF
--- a/crates/bridge_core/src/lib.rs
+++ b/crates/bridge_core/src/lib.rs
@@ -633,11 +633,7 @@ impl<'a> CoreBridgeState<'a> {
             }
         };
 
-        if let Some(t) = maybe_time {
-            t
-        } else {
-            1 // Intentionally make this distinguishable from the error value 0
-        }
+        maybe_time.unwrap_or(1)
     }
 
     fn input_seek(&mut self, handle: *mut InputHandle, pos: SeekFrom) -> Result<u64> {
--- a/crates/docmodel/src/document.rs
+++ b/crates/docmodel/src/document.rs
@@ -53,6 +53,10 @@ pub struct Document {
     /// Either a URL or a local path.
     pub bundle_loc: String,
 
+    /// Extra local search paths for this document.
+    /// May be absolute or relative to src_dir.
+    pub extra_paths: Vec<PathBuf>,
+
     /// The different outputs that are created from the document source. These
     /// may have different formats (e.g., PDF and HTML) or the same format but
     /// different settings (e.g., PDF with A4 paper and PDF with US Letter
@@ -98,6 +102,7 @@ impl Document {
             build_dir: build_dir.into(),
             name: doc.doc.name,
             bundle_loc: doc.doc.bundle,
+            extra_paths: doc.doc.extra_paths.unwrap_or_default(),
             metadata: doc.doc.metadata,
             outputs,
         })
@@ -116,10 +121,17 @@ impl Document {
             .map(syntax::TomlOutputProfile::from)
             .collect();
 
+        let extra_paths = if self.extra_paths.is_empty() {
+            None
+        } else {
+            Some(self.extra_paths.clone())
+        };
+
         let doc = syntax::TomlDocument {
             doc: syntax::TomlDocSection {
                 name: self.name.clone(),
                 bundle: self.bundle_loc.clone(),
+                extra_paths,
                 metadata: None,
             },
             outputs,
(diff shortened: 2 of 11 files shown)


ratmice:
FWIW, in the referenced thread, I added a patch which restricts the path to relative paths using `cap_std`.
This probably needs some error handling, currently it acts like e.g. passing an invalid path to `TEXINPUTS`, and just ignores them.
that branch is here:

https://github.com/ratmice/tectonic/tree/extrapaths

rm-dr:
> restricts the path to relative paths using `cap_std`.

Do we really want to do this? I don't see a reason to restrict input to relative paths, it's generally known that absolute paths will break on different systems.

ratmice:
> > restricts the path to relative paths using `cap_std`.
> 
> Do we really want to do this? I don't see a reason to restrict input to relative paths, it's generally known that absolute paths will break on different systems.

We generally know my opinion, but I wanted to add some additional context.

What I should probably say is not just relative paths but specifically subpaths of the `cwd` where `Tectonic.toml` is found.  So under cap_std `../bar` is still not a valid path even though it is relative.  This provides some benefit to reproducibility in that if you have a directory with a `Tectonic.toml` it *should* reproduce regardless of any external directories even if all paths are relative.

And we could always add some form of escape hatch like `unrestricted_paths` along the same lines of `shell-escape` to disable it.
But indeed it would be good to make a decision on this

rm-dr:
> What I should probably say is not just relative paths but specifically subpaths of the cwd where Tectonic.toml is found.

Ah, I see. I'd argue against this: there's no need to enforce reproducibility that strictly. Tectonic should have reproducible defaults & easy-to-track bundles, but I don't see a reason to enforce relative paths.

In fact, if you need this feature, you'll most certainly be using assets outside of the root `Tectonic.toml` dir! Assets inside the document dir could easily be moved into `src`, and shared assets shouldn't ever find themselves under a single document's directory (they're *shared*, after all).

ratmice:
> > What I should probably say is not just relative paths but specifically subpaths of the cwd where Tectonic.toml is found.
> 
> Ah, I see. I'd argue against this: there's no need to enforce reproducibility that strictly. Tectonic should have reproducible defaults & easy-to-track bundles, but I don't see a reason to enforce relative paths.
> 
> In fact, if you need this feature, you'll most certainly be using assets outside of the root `Tectonic.toml` dir! Assets inside the document dir could easily be moved into `src`, and shared assets shouldn't ever find themselves under a single document's directory (they're _shared_, after all).

Mostly I've been using this with `git clones` of packages, I don't want the package actually in the `src/` dir which should be reserved for my document, yes being *shared* is wonky, but if I use different versions of packages in different projects it is very likely to break when other projects use different versions of the same package.  I have used this scheme at times with both the kaobook and acro packages.

rm-dr:
I see... but I'd still argue that strict relative path checks aren't necessary here---most use cases would raise "not reproducible" errors, as I detailed above.

@pkgw, I think this is ready to merge! This is a small and non-breaking config API change, but consensus on the details would still be nice.

P.S: Github, this should close #985

PR #1119, opened 2023-11-21
Update Arch Linux package URL in index.md

felixonmars:
The old URL returns 404 now.


pkgw:
Thank you for the correction! I have used this PR as an opportunity to add a few housekeeping changes that I'd also like to incorporate.

PR #1123, opened 2023-11-26
Added Tectonic.toml input array

rm-dr:
This PR implements the feature detailed in #1100. Docs have been updated.

A final question I haven't had time to check:
Do all these fancy input file manipulations break latex error line numbers?

If they do, that's bad.

### DIFF
--- a/docs/src/ref/tectonic-toml.md
+++ b/docs/src/ref/tectonic-toml.md
@@ -1,13 +1,13 @@
 # The `Tectonic.toml` File
 
 **Starting with [the V2 interface][v2]**, the `Tectonic.toml` file defines a
-Tectonic document.
+Tectonic workspace.
 
 [v2]: ./v2cli.md
 
 ## Contents
 
-The `Tectonic.toml` file is expressed in, yes, [TOML] format. Allowed items in
+The `Tectonic.toml` file is written in the [TOML] format. Allowed items in
 the file are detailed below.
 
 [TOML]: https://toml.io/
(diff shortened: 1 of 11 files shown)


pkgw:
I think the general direction here is good, but I think it would be very undesirable to break compatibility with existing `Tectonic.toml` documents so dramatically. Existing TOML fields should be honored with their current semantics. It's OK to have new fields that are incompatible with ones that already exist, but you should be able to put old files into a new program and get the same results, error-free.

This is a bit more subjective, but I think that the "main" file for a project should stay with the name `index.tex`, rather than changing it to `main.tex` (even though I just described it as the "main file" ...). Besides the backwards-compatibility aspect, I like the consistency with `index.html` and `index.js` seen in other environments.

pkgw:
@rm-dr I noticed the same thing about the Harfbuzz stuff. It might help to rebase onto the latest `master`? I think that this branch might be based on an older commit before the update to Harfbuzz 8.3.0 ... which should then go away in the merge, but maybe git/GitHub is weird about that?

rm-dr:
There we go, the git history now looks pretty.
This is almost done. There's just one thing left to do:

I need to change the default input files from the array to a single `main.tex` (or `index.tex`), and update the default workspace creation method. This isn't as easy as it sounds, since the v2 tests are tightly coupled to the file naming scheme, and fail if the defaults are changed.

docs/src/ref/tectonic-toml.md
```diff
@@ -24,101 +24,52 @@ pubish = false
...
+#
+# Finally, you may include an "inline" document as follows:
+# inputs = [
+#   { inline = "\\documentclass[a4paper]{article}" },
+#   "main.tex"
+# ]
+# This will insert "\documentclass[a4paper]{article}" before main.tex
+# (with a newline), allowing you to set options without making a new file.
+inputs = "main.tex"
+```
```
pkgw, line 75: This new documentation structure is an improvement, I think. But the reference documentation should document the `preamble`, etc., fields, even if they're deprecated.

pkgw (commented):
I have a few suggestions relating to the documentation, but I like how this is looking! I think this change will make a lot of people happy.

pkgw (approved):
D'oh, looks like this needs to be updated to not conflict with the curly-quotes change, but once that's done, I think you should merge this.

Issue #1002, opened 2023-02-08
Provide options for V2 cli to override the default bundle_url


Sinofine:
When we use `tectonic -X new` or `tectonic -X init` to initiate projects, the output is as follows.
```
note: "version 2" Tectonic command-line interface activated
note: creating new document in directory `xxxx`
note: connecting to https://relay.fullyjustified.net/default_bundle.tar
note: resolved to https://data1.fullyjustified.net/tlextras-2020.0r0.tar
```

It seems that the url is hard coded in
https://github.com/tectonic-typesetting/tectonic/blob/8bf8da71424b2c166dbd4e442c908105ed36bce7/crates/bundles/src/lib.rs#L114-L122

and the config is default using that source
https://github.com/tectonic-typesetting/tectonic/blob/8bf8da71424b2c166dbd4e442c908105ed36bce7/src/config.rs#L194-L202

The source listed above is acceptable for daily usage, but for users in an internal network without access to internet, they may use mirrors or local copy for bundles. However, the command `tectonic -X new` or `tectonic -X init` themselves are default using the hard-coded bundle url to complete initiation, which is meant if the default url is not accessible then the initiation won't be accomplished (usually we waits for minutes and then it throws connection errors and fails).

In the V1 cli, it provides `--web-bundle` to override the default, but in the newer V2 cli, the option is absent. If we provide the option for `tectonic -X new` and `tectonic -X init`, it may be nicer for people unable to access the default bundle url.

pkgw:
Yes, this is a good point. This should be a fairly straightforward improvement for someone new to the project to implement, I'd hope, so I'll tag as such!

bryango:
I came up with a temporary fix for this at build time in #1131. Please see if it's okay!

Issue #893, opened 2022-04-25
Biblatex fails due to version missmatch


MTRNord:
tectonic fails to build with the 2020.0r0 bundle and biblatex:

```
Running external tool biber ...
error: the external tool exited with an error code; its stdout was:

===============================================================================
INFO - This is Biber 2.17
INFO - Logfile is 'default.blg'
INFO - Reading 'default.bcf'
ERROR - Error: Found biblatex control file version 3.7, expected version 3.8.
This means that your biber (2.17) and biblatex (3.14) versions are incompatible.
See compat matrix in biblatex or biber PDF documentation.
INFO - ERRORS: 1
===============================================================================
error: its stderr was:

===============================================================================
===============================================================================
error: the external tool exited with error code 2
```

any ideas how I can fix that?

pkgw:
My guess is that the best fix will be to upgrade to the soon-to-be-released next version of Tectonic that will bump us up to TeXLive 2021.3! I will try to get it done today.

MTRNord:
> My guess is that the best fix will be to upgrade to the soon-to-be-released next version of Tectonic that will bump us up to TeXLive 2021.3! I will try to get it done today.

While I agree that makes sense, I sadly ran into an issue trying it yesterday :/ https://github.com/tectonic-typesetting/tectonic/discussions/883#discussioncomment-2631950

MTRNord:
Hm I tried now with version 0.9.0 of tectonic and the newer bundle, and it seems that suffers from a different but essentially the same issue:

```
RROR - Error: Found biblatex control file version 3.7, expected version 3.8.
This means that your biber (2.17) and biblatex (3.16) versions are incompatible.
See compat matrix in biblatex or biber PDF documentation.
INFO - ERRORS: 1
===============================================================================
error: its stderr was:

===============================================================================
===============================================================================
error: the external tool exited with error code 2
[Finished running. Exit status: 0]
```

biblatex is now slightly newer, but the error still happens. Biber I guess is from my MikTex install, which should be up-to-date.

KevoSoftworks:
I'm running into the same issue as well. A workaround to this problem is to downgrade biber to version 2.16.

pkgw:
OK, well, the upgrade to TeXLive 2022.0 should happen much faster since the diffs are comparatively small and lots of good infrastructure is now in place, and maybe that will make biber 2.17 happy ...

doronbehar:
[...]

PR #1131, opened 2023-12-15
Allow overriding web bundle at _build time_

bryango:
This PR enables us to override the default web bundle **at build time,** through the environment variables (or in `build.rs`):
- `TECTONIC_BUNDLE_PREFIX`
- `TECTONIC_BUNDLE_LOCKED`

The PREFIX variable makes it easy for people to track a mirrored bundle, while the LOCKED variable ensures reproducible builds across all CLIs. This is related to #1002 and its resolution in #1132.

Currently, the default web bundle is hardcoded, as observed in #1002. A related issue is the version mismatch between biber and the tectonic bundle #893, which we ~~would like to~~ addressed in #1132 and nixpkgs https://github.com/NixOS/nixpkgs/pull/273740.

P.S. I am very new to rust, so if the code / interface is no good, please help me improve! And thank you for this wonderful package!


rm-dr:
This is somewhat related to #1121, which completely overhauls Tectonic's bundles (and thus conflicts with the changes here)

Notably, #1121 adds a new bundle format, and setting a variable for just the url prefix won't work. It may be a better idea (even without extra bundles) to have the variable contain the whole URL.

Also, I don't think I understand the difference between the LOCKED and PREFIX variables. Is there any reason to have two? I think one variable that overrides the default bundle url should be enough.


A few more notes about the issue in #1002:
 - Should this be a cli option or a env variable?
 - Why does `tectonic new` touch the network in the first place? I'd argue that the command should create a skeleton project completely offline (with the default url in `tectonic.toml`), which can be changed later by the user. This is clearer and doesn't need any extra config options.

bryango:
Hi @rm-dr! Thank you for the comment!

> This is somewhat related to #1121, which completely overhauls Tectonic's bundles (and thus conflicts with the changes here)

Wow amazing! I didn't know this before! This is a huge change though. Are you aware of https://github.com/tectonic-typesetting/tectonic/discussions/1122? I think being a member would probably make it easier to land these changes.

> Notably, #1121 adds a new bundle format, and setting a variable for just the url prefix won't work. It may be a better idea (even without extra bundles) to have the variable contain the whole URL.
> 
> Also, I don't think I understand the difference between the LOCKED and PREFIX variables. Is there any reason to have two? I think one variable that overrides the default bundle url should be enough.

Currently, the bundle url depends on the `format_version`:
https://github.com/tectonic-typesetting/tectonic/blob/c64e5243467fe94a1a6abe95cf8163ac2b3faacd/crates/bundles/src/lib.rs#L108-L123
So the PREFIX variant only modifies the prefix, in this case `https://relay.fullyjustified.net`, and keeps the version suffix `default_bundle_v{format_version}.tar`. If this is going away in the future, then only the LOCKED variant is needed. What do you think? Should I simply remove the PREFIX variant?
 
> Should this be a cli option or a env variable?

They are different! Let me clarify:
- the env variable provides a **build time** override for distro packagers, such that the tectonic bundle they distributed will match the biber version they distributed. This is related to #893.
- the `--web-bundle` option is for **runtime** override by the end user. This relates to #1002.

I think for the runtime override, we should only provide the CLI option, because env variables are more implicit and thus reduce reproducibility. For build time, however, I think env variables are the only way to pass around the customized build constants, so I have to use it.

> Why does `tectonic new` touch the network in the first place? I'd argue that the command should create a skeleton project completely offline (with the default url in `tectonic.toml`), which can be changed later by the user. This is clearer and doesn't need any extra config options.

AFAIK, `tectonic -X new` and `init`:
- takes a generic web bundle link `https://relay.fullyjustified.net/default_bundle_v33.tar`,
- redirects it to a _version-locked_ one, such like `https://data1.fullyjustified.net/tlextras-2022.0r0.tar`,
- and finally pins it down in the `Tectonic.toml` file.

I think this behavior is nice as it maximizes reproducibility. There are some suggestions in issues to put the locked url `...tlextras-2022.0r0.tar` in a dedicated `Tectonic.lock` file. When and if that happens, `tectonic -X new` and `init` can be completely offline, as it only needs to write the unlocked link `...default_bundle_v33.tar`.

doronbehar:
I really like the idea of having an environment variable for this, but I also don't understand why we need 2 environment variables, and not only 1.

Besides that, I also support having this environment variable read in runtime and not build time, as this would also help users to test different bundles in a different way, I think there is no advantage in terms of reproducibility between reading this environment variable in runtime vs build time; since Tectonic from different distributions may have different environments anywhere.

I also think it'd be fair for Tectonic to not guarantee reproducibility in case it reads a dedicated environment variable, and there's a conflict with the URL in the `Tectonic.toml` file. In such a case I think it'd be best to simply print a big warning, saying that we used the URL from the environment variable and not the the URL from `Tectonic.toml` file.

bryango:
Hi @doronbehar! Thank you very much for the comment!

1. I am absolutely okay with a single env variable such as `TECTONIC_WEB_BUNDLE_URL`. As explained above, the `PREFIX` variant is only introduced for backward compatibility and ease of mirroring. If we decide that it is confusing, I can easily remove it!

2. When it comes to runtime, I would still prefer explicit flags such as `--web-bundle`, which can be overridden by a second `--web-bundle`. (Is this the current behavior? I would have to try it out...)

> I think there is no advantage in terms of reproducibility between reading this environment variable in runtime vs build time; since Tectonic from different distributions may have different environments anywhere.

Yes, this is true in reality, but ideally, I would like Tectonic to _not_ read _any_ environment variables at all (at runtime), by design. It's like the new flake-y nix no longer reads the `NIXPKGS_*` env variables; sure, it makes things a bit more inconvenient, but I like the purity in the UI.

I know @pkgw is probably busy, but since this is now more than a bugfix and related to the UI design, I would very much like to hear your opinion, and I can fix the code accordingly!

doronbehar:
> 2\. When it comes to runtime, I would still prefer explicit flags such as `--web-bundle`, which can be overridden by a second `--web-bundle`. This is not yet the current behavior, and I can look into fixing it.

I did noticed that the 2nd commit of this PR as of currently is referring to this? But not completely? Maybe it'd be easier for upstream to digest this change only in a PR of its own.

rm-dr:
> They are different! Let me clarify:

Ah, I see. We really do need a compile-time option for this.
The two-link config is a bit messy, but I see why it's necessary...

and I don't think I know a better way to do this, if we want easy bundle updates through a "default" redirect.

bryango:
Following the suggestion of @doronbehar I have moved the CLI changes to a separate PR #1132. These two PRs are both related to the web bundle interface, but they are technically independent, so it would be easier to review this way.

pkgw:
I've merged #1132, so that this PR becomes a lot smaller.

It does seem reasonable to me to have the default bundle URL be build-time configurable, for specialized applications.

My only comment upon first glance is one of the evergreen ones: these changes should be documented somewhere!

bryango:
> My only comment upon first glance is one of the evergreen ones: these changes should come with corresponding documentation somewhere!

Yes indeed! But since #1132 has been merged, this one is no longer a blocking issue for me :laughing: I will try to document it some time!

bryango:
The top level bundle override `tectonic --web-bundle -X ...` broke after the bundle work:
- #1218

... so we need this again for downstream reproducible packaging / distribution of tectonic. I have added the relevant documentation in the tectonic book. 

It would be great if this patch is considered before the next release; that would make downstream packaging much easier :laughing: Thank you! @rm-dr @CraftSpider @pkgw

CraftSpider:
I'll have to take some time to familiarize myself with our bundle code, but I'll try to take a look at this.

docs/src/howto/build-tectonic/bundle-default-config.md
```diff
@@ -0,0 +1,56 @@
...
+### `TECTONIC_BUNDLE_PREFIX`
+
+This variable is **required** at compile time and specifies the base URL prefix for bundles.
+A default value is provided in the build script `crates/bundles/build.rs` in the source repository.
+
+The variable is used to construct bundle URLs in the following pattern:
+```bash
+$TECTONIC_BUNDLE_PREFIX/default_bundle$_vFORMAT_VERSION.tar
+```
+where `$_vFORMAT_VERSION` is a suffix from an internal variable, used to distinguish different versions
```
CraftSpider, line 22:
```suggestion
where `$FORMAT_VERSION` is a suffix from an internal variable, used to distinguish different versions
```

crates/bundles/build.rs
```diff
@@ -0,0 +1,37 @@
...
+fn bundle_presets() {
+    // load from env
+    let bundle_locked = env::var(LOCKED_VAR_NAME).unwrap_or("".into());
+    let bundle_prefix = match env::var(PREFIX_VAR_NAME) {
+        Ok(x) if !x.is_empty() => x,
+        _ => TECTONIC_BUNDLE_PREFIX_DEFAULT.into(),
+    };
+
+    // export to rustc
+    println!("cargo::rustc-env={LOCKED_VAR_NAME}={bundle_locked}");
```
CraftSpider, line 30: I'd remove this whole build script, and just do an `unwrap_or(TECTONIC_BUNDLE_PREFIX_DEFAULT)` in `lib.rs`. The environment variables will be passed through to `rustc` by default.

CraftSpider (changes requested):

crates/bundles/build.rs
bryango: Good idea! Let me try!

CraftSpider:
Thank you for keeping up on this after a whole year and a half of waiting - hopefully a 0.16 release can be cut soon.

bryango:
I should thank you for taking care of the project! Once I experienced Tectonic, I don't want to go back to texlive anymore, so I am just trying my best to help it live on 😆

PR #1132, opened 2023-12-19
`--web-bundle` for v2 CLI: fix #1002 and allow overrides

### LINKED ISSUE #1002: Provide options for V2 cli to override the default bundle_url

bryango:
This PR is split off from #1131. See #1131 for relevant discussions on the web bundle interface.

1. Fixes #1002: It enables us to override the default web bundle for the v2 CLI `tectonic -X new` and `tectonic -X init`, with the usual `--web-bundle` flag
2. Allows overrides: It allows user overrides of the `--web-bundle` by specifying it multiple times. The last specification of `--web-bundle` wins.
3. `tectonic -X build` is special as it never reads from the global config, but only from the local `Tectonic.toml` file, for maximal reproducibility. In this case, the global `--web-bundle` option, if specified, is ignored with a warning.

For a horrible example,
```bash
tectonic \
  --web-bundle "https://this-url-gets-overridden-1" \
  -X \
  --web-bundle "https://this-url-gets-overridden-2" \
  new \
  --web-bundle "https://this-url-gets-overridden-3" \
  --web-bundle "https://this-url-is-actually-used"
```

### DIFF
--- a/src/bin/tectonic/compile.rs
+++ b/src/bin/tectonic/compile.rs
@@ -36,11 +36,6 @@ pub struct CompileOptions {
     #[structopt(takes_value(true), parse(from_os_str), long, short, name = "file_path")]
     bundle: Option<PathBuf>,
 
-    /// Use this URL to find resource files instead of the default
-    #[structopt(takes_value(true), long, short, name = "url")]
-    // TODO add URL validation
-    web_bundle: Option<String>,
-
     /// Use only resource files cached locally
     #[structopt(short = "C", long)]
     only_cached: bool,
@@ -95,7 +90,12 @@ pub struct CompileOptions {
 }
 
 impl CompileOptions {
-    pub fn execute(self, config: PersistentConfig, status: &mut dyn StatusBackend) -> Result<i32> {
+    pub fn execute(
+        self,
+        config: PersistentConfig,
+        status: &mut dyn StatusBackend,
+        web_bundle: Option<String>,
+    ) -> Result<i32> {
         let unstable = UnstableOptions::from_unstable_args(self.unstable.into_iter());
 
         // Default to allowing insecure since it would be super duper annoying
@@ -193,7 +193,7 @@ impl CompileOptions {
         }
         if let Some(path) = self.bundle {
             sess_builder.bundle(config.make_local_file_provider(path, status)?);
-        } else if let Some(u) = self.web_bundle {
+        } else if let Some(u) = web_bundle {
             sess_builder.bundle(config.make_cached_url_provider(&u, only_cached, None, status)?);
         } else {
             sess_builder.bundle(config.default_bundle(only_cached, status)?);
--- a/src/bin/tectonic/main.rs
+++ b/src/bin/tectonic/main.rs
@@ -6,7 +6,6 @@ use std::{env, process, str::FromStr};
 use structopt::StructOpt;
 use tectonic_status_base::plain::PlainStatusBackend;
 
-use structopt::clap;
 use tectonic::{
     config::PersistentConfig,
     errors::SyncError,
@@ -40,7 +39,7 @@ mod v2cli {
 #[derive(Debug, StructOpt)]
 #[structopt(name = "Tectonic", about = "Process a (La)TeX document")]
 struct CliOptions {
-    /// Use experimental V2 interface (see `tectonic -X --help`); must be the first argument
+    /// Use experimental V2 interface (see `tectonic -X --help`)
     #[structopt(short = "X")]
     use_v2: bool,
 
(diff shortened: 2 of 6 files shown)


bryango:
Whoops, I should write tests, but I don't know how to... :crying_cat_face:

doronbehar:
> Whoops, I should write tests, but I don't know how to... 😿

You may want to do that after upstream approves the intentions of the PR.

I have a question: Is it possible to implement the following behavior:

```
tectonic --web-bundle <URL1> document.tex # Uses URL1
tectonic --web-bundle <URL1> --web-bundle <URL2> document.tex # Uses URL2
tectonic --web-bundle <URL1> -X build # Uses URL1
tectonic --web-bundle <URL1> -X build --web-bundle <URL2> # Uses URL2
tectonic --web-bundle <URL1> -X --web-bundle <URL2> build # Uses URL2
```

This way, the PR would help a lot both Nixpkgs' situation, and all other users. For Nix users, this would allow to wrap Tectonic with `--add-flags` (see [this](https://github.com/NixOS/nixpkgs/blob/4fa61810859ac4bbcdf0e8a1014040fac21a83c7/pkgs/build-support/setup-hooks/make-wrapper.sh#L24-L25)), and for other users it'd allow to simply add a shell alias such as:

```
alias tectonic='tectonic --web-bundle=<URL that matches my biber version>'
```

With the above described behavior to be able to override the web-bundle, a no-Nix user won't need to `unalias` the `tectonic` alias, and Nix users won't need to use the `tectonic-unwrapped` derivation. This is the ideal solution IMO to the issue, that is the least intrusive.

doronbehar:
@bryango this seems to get into great shape! Considering you have added extensive tests, do you think it should be ready for inclusion in Nixpkgs? Along with an additional `makeWrapper` argument `--add-flags`?

bryango:
Hi @doronbehar! Thank you for following along! I think your previous suggestions are great and have been trying to implement them.

> do you think it should be ready for inclusion in Nixpkgs? Along with an additional `makeWrapper` argument `--add-flags`?

This is the eventual goal! I think I finally got everything working, but I am not entirely satisfied about the current resolution. The code feels a bit too hacky... I will try to improve it a bit more, and after some more testing, I think it would be good to go into nixpkgs! But it might take some time :pleading_face:

bryango:
@pkgw @doronbehar @rm-dr Happy new year! I think this is the best I can do, and it's ready for review. The changes may look large and scattered, but in fact half of them are test changes. It is easy to review commit by commit: https://github.com/tectonic-typesetting/tectonic/pull/1132/commits

- Each commit is "bite-sized" and has a clear objective stated in its message.
- The package is buildable at every commit and passes `cargo test`.
- The first few commits are refactors of the current behavior, and the latter ones add new `--web-bundle` options.

pkgw:
Sorry for taking to long to review this :frowning_face:

Upon first read I wasn't a fan of the changes to allow `-X` later in the command line, but I am guessing that the motivation there is to make it possible to set up some kind of alias `mytectonic = tectonic --web-bundle=https://...../` and Things Will Work. So, that seems reasonable.

Likewise I don't *love* having the `-X build` command accept a non-functional `--web-bundle` argument, but the homogeneity also helps with the above.

So I'm going to go ahead and merge this. Sorry again for the lack of responsiveness.

bryango:
Thank you very much for merging this!

> Upon first read I wasn't a fan of the changes to allow `-X` later in the command line, but I am guessing that the motivation there is to make it possible to set up some kind of alias `mytectonic = tectonic --web-bundle=https://...../` and Things Will Work. So, that seems reasonable.

The current solution is indeed ugly, but once v2 CLI has fully taken over, this would hopefully look a bit better!

> So I'm going to go ahead and merge this. Sorry again for the lack of responsiveness.

Please there is no need to apologize! It is already hard enough to work in academics, let alone maintaining many open source codebases where users like me are constantly blabbing for features :laughing: Thank you very much for doing this!

PR #1138, opened 2024-02-05
Port xetex_layout to Rust

CraftSpider:
Been poking this on and off for a while, finally reached the point where it's... not done, but should pass most tests on Windows and hopefully Linux. Mac will require a bit more love before it works. Opening to see what CI has to say so far.

Merge Readiness:
- [x] Split tests into separate PR, merge first
- [x] Drop test-only changes
  - [x] CI changes
  - [x] ICU bridge changes
- [x] Split each bridge's Rust interface into its own PR
  - [x] Create fontconfig bridge
  - [x] Harfbuzz
  - [x] Graphite
  - [x] Freetype
  - [x] Mac Core
- [x] Ensure tectonic-on-arXiv works on the final result


pkgw:
Ooh, very exciting!

crates/xetex_layout/src/lib.rs
```diff
@@ -27,6 +27,175 @@ mod linkage {
     use tectonic_bridge_icu as clipyrenamehack5;
 }
 
+macro_rules! cstr {
```
Mrmaxmeier, line 30: Hi. [CStr literals](https://rust-lang.github.io/rfcs/3348-c-str-literal.html) have been stabilized recently so they'd fit in nicely here :)

CraftSpider: It looks like they were delayed by a release, so they're stabilizing in 1.77, the release after the one coming out in a couple days. I'll leave a TODO to fix it then, if this merges before that release.

CraftSpider:
I am so baffled by the linking errors on everything - they aren't happening locally, which makes them pretty hard to diagnose.

crates/xetex_layout/src/lib.rs
Mrmaxmeier: Oh whoops, right. "Tracking issue closed as stabilized" doesn't mean that the version is stable already ^^

Mrmaxmeier:
Yep, I saw `undefined reference to symbol 'BrotliDecoderDecompress'` in my CI logs and was also wondering what that's about. Probably some strange interaction with dependencies.. (?) I'll try poking at the CI container a bit.

crates/bridge_freetype2/src/lib.rs
```diff
@@ -1,5 +1,481 @@
...
+pub type FT_Module = *mut FT_ModuleRec;
+pub type FT_Glyph = *mut FT_GlyphRec;
+pub type FT_Fixed = libc::c_long;
+pub type FT_Error = libc::c_int;
+pub type FT_Pos = libc::c_long;
+pub type FT_String = libc::c_char;
+pub type FT_Generic_Finalizer = extern "C" fn(object: *mut ());
+pub type FT_Pointer = *mut ();
+
+#[link(name = "freetype", kind = "static")]
```
Mrmaxmeier, line 416: It seems like this directive links to a _second_ (static) copy of freetype that doesn't match up with our harfbuzz freetype build. I'm not sure of the consequences, but simply dropping the `#[link]` line both here and in crates/bridge_harfbuzz/src/lib.rs seems to work.

CraftSpider: Ah, good find, I'll try it locally. I'm using a Windows vcpkg dev environment, and it needs some stuff static so I suspect I added it while narrowing down another issue, and either it's only necessary in some cases, or it's not necessary at all and I just left it in by accident.

crates/xetex_layout/src/lib.rs
pkgw: Also, we don't have a formal Minimum Supported Rust Version (MSRV) policy, but I think it's generally good to give new stabilizations at least a few features to propagate through the ecosystem before starting to rely on them.

CraftSpider:
The icu/unicode stuff is probably because the library does symbol versioning, which you only disable normally on say system installs. I probably, somehow, am locally getting some system installed version pulled in which is providing the symbols. This is... kind of annoying, but I can fix it.

CraftSpider:
Found a solution but it will require updates to `vcpkg-rs` - that PR will go up tonight or tomorrow, probably.

CraftSpider:
I should look into whether there's an existing `CoreFoundation`/`CoreText` library for mac. Since they're, well, core, it should be possible to use a standard `sys` crate for them without jumping through the hoops other dependencies need due to supporting so many ways of handling them.

pkgw:
There's https://github.com/servo/core-foundation-rs ... the crates don't seem as popular as I'd have thought, but a Servo project is probably going to be trustworthy?

CraftSpider:
@pkgw What's your take on dropping support for MacOS < 10.7, OSX Lion? It's more than 10 years old, and some quick googling says the number of mac users on it should be basically negligible. It would allow some minor simplification.

CraftSpider:
Related note - the core-foundation-rs crate and similar from servo target 10.7 and above, so it would allow using them.

CraftSpider:
This is a big change - if desired, I can split it out into the primary change to Rust, then the individual library wrappers as follow-ups to the big change, as I made sure to get tests passing between each major change.

CraftSpider:
Obviously I'm biased, but I'm tempted to pull in my crate `enrede` to replace `icu` at some point. It means both one fewer allocation for recoding strings, but also would allow replacing most usages of the std `CString` with enrede's `CString<Utf8>`, an encoding-strict type that is null terminated, but safe to turn to a `&str` for easy use in Rust code.

CraftSpider:
Swapped to using to `enrede`, which avoids the need to change `vcpkg`. Thus this should be nearly ready to merge in theory, though I probably want to try to clean up the history or split it into multiple PRs first.

CraftSpider:
Wow, I actually didn't expect ToA to succeed first try - that means this just needs to be split into each bridge's changes and we're good.

CraftSpider:
Hmm, will probably revert that commit for now. I like the idea of using an external dep here, but it's missing one or two APIs that would be nice, and honestly I don't really like the code style it uses. Plus it forces a lot more string shuffling.

Enivex:
I see that harfbuzz is mentioned here. Would it be viable to use https://github.com/harfbuzz/harfrust (or the original rustybuzz)?

CraftSpider:
It may be viable, yes, though currently harfrust has a couple limitations - for example, it doesn't support Graphite2 currently. So for now, it would have to be an optional backend.

PR #1151, opened 2024-02-15
shell escape: fixes creation of temporary subdirectories

### LINKED ISSUE #1149: No such file or directory error when using shell escape and tex files in subdirectory
Hello,
I'm facing an issue where a temporary file cannot be written during the compilation process when shell-escape is enabled to use the minted package.
I suspect the issue comes from the fact, that some of my source files lie in a subdirectory of the root src directory and the file creation in the tmp directory fails as the subdirectory is not explicitly created before writing the file.

If this is the case, a simple call to the `std::fs::create_dir_all` function should be able to create the subdirectories before trying to write files into them.

<details>
  <summary>Full error log</summary>

```
warning: lineno.sty:296: Invalid UTF-8 byte or sequence at line 296 replaced by U+FFFD.
warning: open of input /dev/null.tex failed
caused by: access to the path `/dev/null.tex` is forbidden
warning: open of input /dev/null.tex failed
caused by: access to the path `/dev/null.tex` is forbidden
error: failed to create file `/tmp/.tmpdFJUpo/chapters/results.aux`
caused by: No such file or directory (os error 2)
error: failed to execute the shell-escape command "uname -s > "doc.w18"": execution of the request failed
error: ifplatform.sty:92: Package catchfile Error: File `doc.w18' not found.

See the catchfile package documentation for explanation.
Type  H <return>  for immediate help
error: halted on potentially-recoverable error as specified
```
</details>

derchr:
Fixes #1149.
The MR applies the fix as described in https://github.com/tectonic-typesetting/tectonic/issues/1149#issuecomment-1943822679.
I added also added some error reporting around the call of std::fs::create_dir_all.

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -704,6 +704,12 @@ impl DriverHooks for BridgeState {
                 }
 
                 let real_path = work.root().join(name);
+                if let Some(prefix) = real_path.parent() {
+                    std::fs::create_dir_all(prefix).map_err(|e| {
+                        tt_error!(status, "failed to create sub directory `{}`", prefix.display(); e.into());
+                        SystemRequestError::Failed
+                    })?;
+                }
                 let mut f = File::create(&real_path).map_err(|e| {
                     tt_error!(status, "failed to create file `{}`", real_path.display(); e.into());
                     SystemRequestError::Failed


pkgw:
Maybe the `cap_std` crate, mentioned in #1106, would also be helpful for limiting the depth of directory creation? It seems like it might be overkill, but if we're using it anyway, maybe that's OK.

rm-dr:
`cap_std` isn't in Tectonic yet, and is indeed a bit overkill. This could be handled with some fairly simple logic, once we decide on what Tectonic should do when there is no home directory.

Zelzahn:
Hey, what's the status of this PR? And what's still needed until it can be merged?

Issue #1150, opened 2024-02-15
internal consistency problem when checking if paper.bbl changed


KonradHoeffner:
tectonic reruns TeX compilation multiple times until it hits the limit of 6 passes, but I have no idea why.
This warning seems to be emitted by <https://github.com/tectonic-typesetting/tectonic/blob/fed29a8339e09b733a423ccc20bb440363dbc5bc/src/driver.rs#L1364>.

```
/tmp$ git clone git@github.com:annosaxfdm/anno-paper-swj paper
/tmp$ cd paper
paper$ git checkout 1af65946b752e453494f4f806fc7fb7b1f73deb4
paper$ tectonic paper.tex                                                                                                                        heads/master
Running TeX ...
Running BibTeX on paper.aux ...
Rerunning TeX because bibtex was run ...
Rerunning TeX because "paper.out" changed ...
warning: internal consistency problem when checking if paper.bbl changed
Rerunning TeX because "paper.bbl" changed ...
warning: internal consistency problem when checking if paper.bbl changed
Rerunning TeX because "paper.bbl" changed ...
warning: internal consistency problem when checking if paper.bbl changed
Rerunning TeX because "paper.bbl" changed ...
warning: internal consistency problem when checking if paper.bbl changed
Rerunning TeX because "paper.bbl" changed ...
warning: internal consistency problem when checking if paper.bbl changed
warning: TeX rerun seems needed, but stopping at 6 passes
Running xdvipdfmx ...
warning: ICC profile format spec. version 4.4.0 not supported in current PDF version setting.
warning: ICC profile not embedded.
```

derchr:
Probably it's the same issue as [1146](https://github.com/tectonic-typesetting/tectonic/issues/1146).

KonradHoeffner:
Oh sorry, I didn't found that one because it doesn't contain "warning: internal consistency problem", however could that mean that this is a different issue or are warnings just deactivated in the other issue output?

CraftSpider:
If anyone experiencing the issue is willing to try #1155, that should fix the issue by making bibtex correctly close log files on success.

KonradHoeffner:
@CraftSpider: I tried your bibtex-fix branch and while `cargo vcpkg` reported a failure (see below) I think that is unrelated to the branch, it still built successfully and the issue is indeed fixed, thank you!

## cargo vcpkg build error  (probably unrelated, still builds)

```
cargo vcpkg build                                                                                                                        bibtex-fix
[...]

Issue #1146, opened 2024-02-12
Tectonic 0.15 (regression vs 0.14.1) reruns TeX indefinitely believing the .bbl file to be changed (intermediates ignored?)


rmgk:
As the title says. With tectonic 0.15 it seems to go into an infinite (but capped at 6) loop of rerunning TeX, even though none of the files ever changes on disk (keep same sha1sum at least).
This does not happen with tectonic 0.14.1.
See logs below for details.

Additionally, the rerun reasons are weirdly nondeterministic, in the run below it first claimed that .out changed, and then loops on bbl. but sometimes it's also .aux, or only .bbl.

I did look for related issues, and there is a related discussion here (even though the topic is about biblatex, the discussion then goes in other directions)
https://github.com/tectonic-typesetting/tectonic/issues/195#issuecomment-546290732

Same thing on the gnu version instead of musl.
Same problem happens with and without --keep-intermediates (generally, the intermediates seem to be completely ignored) 

Thanks for tectonic, btw 😄

```
❯ tectonic15-musl --keep-intermediates paper.tex 2>| rg rerun
Running TeX ...
Running BibTeX on paper.aux ...
note: warnings were issued by BibTeX; use --print and/or --keep-logs for details.
Rerunning TeX because bibtex was run ...
Rerunning TeX because "paper.bbl" changed ...
Rerunning TeX because "paper.bbl" changed ...
Rerunning TeX because "paper.bbl" changed ...
Rerunning TeX because "paper.bbl" changed ...
Rerunning TeX because "paper.bbl" changed ...
Running xdvipdfmx ...
warning: TeX rerun seems needed, but stopping at 6 passes
Writing `paper.bbl` (14.04 KiB)
Writing `paper.out` (2.13 KiB)
Writing `paper.aux` (7.46 KiB)
Writing `paper.pdf` (152.59 KiB)

❯ tectonic14-musl --keep-intermediates paper.tex 2>| rg rerun
Running TeX ...
Running BibTeX on paper.aux ...
note: warnings were issued by BibTeX; use --print and/or --keep-logs for details.
Rerunning TeX because bibtex was run ...
Running xdvipdfmx ...
Writing `paper.out` (2.13 KiB)
Writing `paper.pdf` (152.59 KiB)
Writing `paper.bbl` (14.04 KiB)
Writing `paper.aux` (7.46 KiB)
```

CraftSpider:
[...]

PR #1154, opened 2024-02-20
Fix "internal consitency problem when checking bbl file"

mhuttner:
I had the same problem as #1150 and #1146.

I think this happens because after running BibTex the state of the bbl file is set to "ReadThenWritten" and 
is never reset causing infinite reruns. 

After resetting this in the loop with the helpful comment
" I am not super confident that the access_pattern data can just be left as-is when we do this, but, uh, so far it seems to work."  everthing seems to work as expected again.

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -1718,6 +1718,9 @@ impl ProcessingSession {
             // this, but, uh, so far it seems to work.
             for summ in self.bs.events.values_mut() {
                 summ.read_digest = None;
+                if summ.access_pattern == AccessPattern::ReadThenWritten {
+                    summ.access_pattern = AccessPattern::Read
+                }
             }
 
             warnings = self.tex_pass(Some(&rerun_explanation), status)?;


src/driver.rs
```diff
@@ -1718,6 +1718,9 @@ impl ProcessingSession {
             // this, but, uh, so far it seems to work.
```
CraftSpider, line 1718: nit: Could you update this comment?

CraftSpider (approved):
At a glance, this looks fine to me. I kind of want to check if we actually have any tests that require multiple runs with bibtex - I'm suspicious we may not, in which case this is effectively untested rerun code. If one of the others agrees that this is a good fix though, I'm happy to merge as-is and add a test later.

CraftSpider:
Turns out it was forgotten - when I ported the code, I ended up making a mistake such that the log files were never closed, and thus digest never updated, when bibtex succeeds. I've opened a PR fixing that, which *should* fix the original issue, though I want to also add a test to catch behavior like this.

mhuttner:
I can confirm #1155 solves the problem for me, and is the better fix. 
Thank you for the quick help! Closing this.

PR #1155, opened 2024-02-21
Fix not closing files when spotless history is returned

CraftSpider:
~~If you saw that master push/revert, no you didn't~~

This should supercede #1154 - on success, log files were not being closed, thus weren't getting write_digest set. See the original bibtex code at [@0.13.0/bibtex.c](https://github.com/tectonic-typesetting/tectonic/blob/tectonic%400.13.1/crates/engine_bibtex/bibtex/bibtex.c) for reference.

### DIFF
--- a/crates/engine_bibtex/src/lib.rs
+++ b/crates/engine_bibtex/src/lib.rs
@@ -333,9 +333,7 @@ pub(crate) fn bibtex_main(ctx: &mut Bibtex<'_, '_>, aux_file_name: &CStr) -> His
 
     let res = inner_bibtex_main(ctx, &mut globals, aux_file_name);
     match res {
-        Ok(History::Spotless) => (),
-        Ok(hist) => return hist,
-        Err(BibtexError::Recover) => {
+        Err(BibtexError::Recover) | Ok(History::Spotless) => {
             // SAFETY: bst_file guaranteed valid at this point
             unsafe { peekable_close(ctx, ctx.bst_file) };
             ctx.bst_file = None;
@@ -345,6 +343,7 @@ pub(crate) fn bibtex_main(ctx: &mut Bibtex<'_, '_>, aux_file_name: &CStr) -> His
             ttbc_output_close(ctx.engine, ctx.bbl_file);
         }
         Err(BibtexError::Fatal) => (),
+        Ok(hist) => return hist,
     }
 
     match get_history() {


pkgw:
This seems like it's more likely to be correcting the root problem than #1154. Do you think there's a straightforward way to add a test case for the issue here?

CraftSpider:
I'm not sure it will be straightforward, but I'm pretty sure I can make a test case for this. I'm kind of surprised none of the existing tests caught it, since it should happen basically always when bibtex is run.

CraftSpider:
Short answer after looking into why it fails: we only have one minimal test exercising the driver, and I'll need to do what some of the comments in the file say and turn it into a proper test suite, or write a targeted unit test in `formats.rs` since that replicates the `write_digest` behavior, but will be a bit more of a pain - I personally think writing more driver tests will get us better bang for our buck, in terms of effort to benefit.

CraftSpider:
Working on it a bit more - making a test is a pain because, basically, the test wants to be able to use LaTeX style, and the tests use a very stripped down LaTeX fmt to avoid doing exactly what these tests want - pulling in all the support stuff.

KonradHoeffner:
This fixes the issue for me!

pkgw:
@CraftSpider Oh, yeah, that sounds like a pain. It would be nice to get a test in, but I fully sympathize if you have trouble getting motivated to slog through that kind of stuff.

(In principle such tests could be attached to the bundle-generating repo, which has local access to the full TeXLive stack by definition; but then the testing workflow becomes a lot more complicated.)

CraftSpider:
I'm not quite sure I have the motivation right now to do all that would be needed to support that, I guess I could go with the other route of using a test driver - just run bibtex, assert all the files are closed at the end or something.

Fun fact, this reminded me of some of the post-Rust cleanups I started, and guess what? One of the commits in that branch already fixed this issue, the exact same way! Apparently I noticed it, fixed it, forgot about it.

pkgw:
Closing/opening this PR to get the CI to rerun against the updated `master`. I think we should merge once CI is happy since this is a high-priority issue.

CraftSpider:
Ah, it's upset I haven't increased coverage. I'll try to do that ASAP, it may or may not be today though as I've had quite a long week.

pkgw:
@CraftSpider I view the Codecov failures as merely advisory ... which maybe isn't the best standard to set, but, oh well. Especially in this case where a really useful test would, as we were just discussing, be fairly annoying to put together.

PR #1163, opened 2024-02-28
Bibtex cleanups / type improvements

CraftSpider:
Best reviewed commit-by-commit if you're looking to understand every change.

This work, so far:
- Renders Bibtex fully re-entrant outside of the use of CoreBridgeState
- Makes some changes to CoreBridgeState to allow sound direct usage from Rust
- Uses `Vec` instead of a custom buffer type everywhere
- Makes the hash data a lot more type-safe.
  - In the future, I plan to remove most of the panics and shift even more stuff into the type system for compile-time guarantee of correctness.
  - I also want to shift where `lookup_str` is defined - it's primarily about the hash, only tangentially touching the pool.
- There's still more to do, but this is a lot of the stuff I really wanted to happen


CraftSpider:
Created a new branch off this to play around with an idea I've had that will probably not be part of this PR (which I'm hoping to get merged after I fix up the test I broke) - take a look at https://github.com/CraftSpider/tectonic/commit/75f5c9656efd8f9df6be749202bd9fa644115fa5 on my bibtex-parsing branch for a new way to parse things. It manages to separate out aux parsing into two parts, just parsing the line, then emitting warnings and making state changes. The parsing part is entirely separate from state, and as such, can now be unit-tested quite easily.

<details>
  <summary>Disclaimer</summary>
  
The library I used, `chumsky`, is one that I'm a maintainer for. I chose it because both I'm familiar with it, and it has strong error handling/recovery support. The current release version, 0.9, is kind of slow, but the 1.0 alpha version is close to release and both makes it nearly as fast as hand-written parsers and supports zero-copy parsing, meaning aux should be zero-allocation again.
</details>

CraftSpider:
Alright, this passes tests, and after some git cleanup, I'm now pretty confident about it.

PR #1166, opened 2024-02-29
Feature: prefer `tectonic-biber` as the biber executable

bryango:
This is a proposed fix of #893.

- When an executable called `tectonic-biber` is found, either in the current working directory or in `$PATH`, prefer that as the biber executable instead of `biber`;
- Add test for tectonic-biber with cleanups. In particular, biber tests in `tests/executable` are refactored to reduce
duplication.
- Add a `flake.nix` file to integrate with the Nix package manager. This makes it easy to onboard new contributors, as flake.nix provides a fully configured development environment, package definition, and continuous integrations. It also serves as a zero-hassle fix for the biber issue, leveraging the power of an external package manager. However, if Nix feels too foreign to the project :laughing: I am okay to remove this commit.


pkgw:
To go on a moderate tangent ... I have been thinking that maybe a good solution for the biber problem, and other issues I see farther down the line, is to provide some lightweight integration with various package managers as you mention here. This would be most useful for systems where packages can be installed, and their associated environments can be activated, very "locally"; I don't know if Nix is like that.

The idea is that V2 `Tectonic.toml` file could support a directive like:

```
environment = <kind>
```

... where <kind> might be something like "nix", "venv", "nodejs", or "pixi". The last of these ([Pixi]) is a Conda-based system which explicitly targets local configuration and install, which is exactly what we would want for reproducibility, and it looks like it would be a great fit for this use case. Perhaps Nix would be as well. Tectonic would only have to integrate with any of these in a lighweight way: basically to be able to set up its executable search path for shell-escape operations. 

[Pixi]: https://prefix.dev/

CraftSpider:
Eventually, we'd like to replace this with #1304, but lets not let perfect be the enemy of good.

PR #1170, opened 2024-03-07
Update remaining out-of-date dependencies

CraftSpider:
Bumps several out-of-date dependencies to the newest versions. The biggest change is `watchexec 3.0`, with `hyper 1.0` being test-only.


CraftSpider:
kcov test failing is weird - it shouldn't even be running in the first place. Apple failure is *also* weird, as it's a failure in `typenum`.

PR #1171, opened 2024-03-12
Generate shell completions for `nextonic`

bryango:
This is made possible by recent cleanups #1162, #1167, #1169 of the cli implementations. Thank you all! @rm-dr @CraftSpider 

The completions are generated by `clap_complete`. The current v1 & v2 cli mixture makes it a bit difficult to offer clean completions for the `tectonic` command, so for now we only target the `nextonic` command exclusively for the v2 cli. To print the shell completions, simply run:
```bash
tectonic -X show shell-completions zsh # or bash, or your favorite <SHELL>
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -397,6 +397,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.5.1"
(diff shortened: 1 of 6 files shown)


bryango:
Documented shell completions in the book.

PR #1179, opened 2024-03-30
Add more tex-output tests

CraftSpider:
Tests that caught bugs in #1138. Also want to see if the Linux specific failures are happening on Master too, which will help me identify how to track them down.

Blocked on #1180


CraftSpider:
So, @pkgw - I've discovered something fun, certain Linux configurations result in different output from all other platforms, including several *other* Linux configurations. I'll try to debug these at some point, but for now, would you prefer to hold off on merging these tests entirely, or just ignore the flaky one on Linux for now?

CraftSpider:
Ugh. It looks like, most likely, somewhere between Harfbuzz 2.7 and 8 there was a change that affected how glyphs end up laid out in the buffer. Maybe a bugfix, maybe not, hard to say. Unfortunately, that means I don't think there's a nice way to 'fix' this - it's not our fault, but will likely become untested shortly when Ubuntu 24 becomes 'latest'. For now, I'll probably just try to ignore the tests on the bad platforms.

CraftSpider:
Test ignored on bad platforms for now - a bit overzealous of a cfg, but on the other hand, it's a test that just wasn't present before, so we're still gaining coverage by running it at least sometimes. And it should still identify bugs like it was originally designed to catch.

PR #1182, opened 2024-04-08
Create parent directories before writing files

### LINKED ISSUE #1027: `--keep-intermediates` breaks for tex files in subdirectories
I recently had an issue where `tectonic -X build --keep-intermediates` would crash due to a 'No such file or directory' error when writing the `.aux` files. After some debugging, I found that tectonic expects the file structure in `build/default` to be the same as `src` but does not create any missing directories, only missing files.

Yoz0:
Fixes #1027
tectonic did not create the subdirectories when using --keep-intermediates, now it creates them everytime.

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -1645,6 +1645,10 @@ impl ProcessingSession {
                 &format!(" ({})", byte_len.get_appropriate_unit(UnitType::Binary)),
             );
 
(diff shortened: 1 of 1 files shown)


CraftSpider:
Thanks for the contribution! Looks like a fine fix to me.

PR #1234, opened 2024-11-14
Use baseline cross instead of old, custom variant

CraftSpider:


CraftSpider:
It works! Once I drop the debug commits and squash it together, we'll be able to just use normal `cross`

PR #1250, opened 2025-02-11
Use `i64` to prevent overflows

### LINKED ISSUE #1249: Overflow error in bibtex engine
When using bibtex engine (tectonic is in master branch version), I met an overflow error:

```
Running BibTeX on tectonic_bug_reproduce.aux ...
thread 'main' panicked at crates/engine_bibtex/src/exec.rs:1947:25:
attempt to add with overflow
```

The error is caused by this line:
https://github.com/tectonic-typesetting/tectonic/blob/ddca0b805c7a8c3edd93effc2b188a05f409161d/crates/engine_bibtex/src/exec.rs#L1947

It looks like the addition to `num_chars` is spilling over its maximum value and a larger interger type is needed here.

naskya:
Fixes #1249

### DIFF
--- a/crates/bridge_flate/src/lib.rs
+++ b/crates/bridge_flate/src/lib.rs
@@ -113,7 +113,7 @@ struct Decompressor<'a> {
     done: bool,
 }
 
-impl<'a> Decompressor<'a> {
+impl Decompressor<'_> {
     fn decompress_chunk(&mut self, output: &mut [u8]) -> Result<usize, Error> {
         if self.done {
             return Ok(0);
--- a/crates/bundles/src/itar.rs
+++ b/crates/bundles/src/itar.rs
@@ -193,7 +193,7 @@ impl Bundle for ItarBundle {
     }
 }
 
-impl<'this> CachableBundle<'this, ItarFileIndex> for ItarBundle {
+impl CachableBundle<'_, ItarFileIndex> for ItarBundle {
     fn get_location(&mut self) -> String {
         self.url.clone()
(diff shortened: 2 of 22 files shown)


CraftSpider:
Hi, thanks for the PR! The failures look unrelated - if you wouldn't mind fixing the clippy one that would be appreciated, but if you'd rather I can merge this as-is and get around to it later.

naskya:
Accepted some clippy suggestions. I hope it helps!

However, the CI will still fail because I left this one untouched:

```console
warning: very complex type used. Consider factoring parts into `type` definitions
   --> crates/engine_spx2html/src/fonts.rs:308:10
    |
308 |     ) -> Result<impl Iterator<Item = (usize, Option<(char, String)>, FixedPoint)> + 'a> {
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity
    = note: `#[warn(clippy::type_complexity)]` on by default
```

it might be worth to add `#[allow(clippy::type_complexity)]` here (or pass `--allow clippy::type_complexity`), but I find that this is beyond what I’m supposed to change.

CraftSpider:
Thanks, looks good to me now. I already have a backlog of maintenance tasks for the CI tests, I'll add this to the list.

PR #1253, opened 2025-02-13
Fix multiple watchexec issues

### LINKED ISSUE #1213: Watch command does not exit when interrupt signal is sent
After executing the command `tectonic -X watch`, when I press Ctrl + C sending the interrupt signal, it seems like watchexec has received the signal successfully, but the program does not terminate. Upon testing, it seems that adding `action.quit_gracefully` or `action.quit` in
https://github.com/tectonic-typesetting/tectonic/blob/82484dbbfd698944c9d29cf6c0944c17e95d1e53/src/bin/tectonic/v2cli/commands/watch.rs#L138-L143
will do the job. But I'm not sure if it is considered a "hack", or perhaps it is an intended behaviour that the interrupt signal only stop the underlying jobs and not the actual tectonic program.

### LINKED ISSUE #1237: tectonic -X watch does not do anything
I am building tectonic from source@master (with the patch in #1213 applied). When I run `tectonic -X watch`, I get the following output 
```
note: "version 2" Tectonic command-line interface activated
```
... and that's it. It stays there forever. This happens on both Windows and Linux.

CraftSpider:
Make watchexec shutdown gracefully on ctrl-C, and make sure new jobs are started, not just existing ones.

Fixes #1213
Fixes #1237

### DIFF
--- a/src/bin/tectonic/v2cli/commands/watch.rs
+++ b/src/bin/tectonic/v2cli/commands/watch.rs
@@ -116,12 +116,17 @@ impl WatchCommand {
         let exec_handler = Watchexec::new_async(move |mut action| {
             let cmds = Arc::clone(&cmds);
             Box::new(async move {
+                // When we spawn a job it doesn't immediately become available. So we chain it
+                // with existing jobs.
+                let mut new_job = None;
+
                 if action.get_job(cmds[0].0).is_none() {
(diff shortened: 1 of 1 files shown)


CraftSpider:
I'm saying 3 lines regression isn't worth it - and it actually should be covered by the watch test it simply can't handle

PR #1257, opened 2025-02-15
Cache vcpkg build to improve CI speed

### LINKED ISSUE #1255: Cache CI vcpkg artifacts
Currently, our CI for vcpkg backends rebuilds vcpkg dependencies on every CI run. Azure Pipelines supports caching, as does vcpkg, so it would be nice if we could make use of that to shave 10-15 minutes off our slowest CI runners.

CraftSpider:
Cache is per-target, and will be invalidated by changes to Cargo.toml, so vcpkg metadata changes should invalidate it.

Fixes #1255


CraftSpider:
Confirmed, this *majorly* improves build times when set up correctly. Tomorrow I'll have to get Windows working.

Issue #935, opened 2022-09-23
Support attachfile Package


bradcarman:
If possible, it would be great if the "attachfile" package could be supported by Tectonic.

mnrvwl:
Hey @bradcarman, I've personally never worked with attached files on PDF before, could you provide a MWE and which PDF viewer you're using that supports them?

mnrvwl:
@bradcarman I got some time to check this with Adobe Acrobat DC

Tecctonic already includes the `attachfile` package in our last 2022 bundle.

When compiling with either `tectonic -p` or `XeTeX` we get:

```console
Package attachfile Warning: attachfile works _only_ with pdfLaTeX and LuaLaTeX
(attachfile)                and _only_ in PDF-generating mode. For this run,
(attachfile)                placeholders will be substituted for all
(attachfile)                attachfile commands.
```

It seems that XeTeX/Tectonic are missing the engine macros for `attachfile` to work, so it fails gracefully. I'm not sure if investigating which macros specifically is the package using but seems something suited for upstream so the package provides support for XeTeX/tectonic. Changing an engine may be harder than changing a package 💜

From `attachfile` documentation:

> _"attachfile requires either pdfLATEX version 0.14 or later or LuaLATEX. [...] While there are many other ways to produce pdf files from LATEX source, attachfile v1.9 supports only pdfLATEX and LuaLATEX."_

Seems this lack of support was a deliberate decision from the package author.

* * *

Looking for alternatives, there's the `embedfile` package, but that only supports pdfTEX/LuaTEX also.

Finally, there's the `attachfile2` package, which is less ancient than its v1 brother and has code that supports XeTeX.

MWE:

```latex
\documentclass{article}
  \usepackage{attachfile2}

\begin{document}
 \attachfile{Test.png}
\end{document}
```

Using it produces a file with an attachment, but when opening it with Acrobat DC the attachment name is an uncanny `˛ˇ` and it fails to open.
Since `attachfile2` used to work my guess is that its in the need of an update since last one was in 2019.

Do you know of other PDF readers that display attachments we could try?

9SMTM6:
Firefox works just fine for that purpose.

mnrvwl:
Thanks, yes I can reproduce the issue with Firefox. Using `attachfile` and `attachfile2` work with other LaTeX engines, but not with tectonic. I've checked and `shell-escape` is not needed for `attachfile2`.

[...]

PR #1261, opened 2025-02-22
Fix incomplete attached file problem

vlasakm:
Currently, we only have `ttstub_input_read` function, which corresponds to Rust's `read_exact`. This seems to work for most of xetex, but it's not semantically a replacement for buffered reading with `fread`, which is used in a dvipdfmx.

One of such places is reading a file to be included in a PDF file as attachment. This is done through dvipdfmx special like this:

    \special{pdf:fstream @SourceFile (example.txt)}

These reads are done in 1024 byte chunks, and if we read these chunks with read_exact, we are going to fail unless the file size is not exactly divisible by 1024.

Tectonic would report this as:

    warning: 1024-byte read failed
    caused by: failed to fill whole buffer

To fix this issue in a non-intrusive way, we introduce a new variant of reads - `ttstub_input_read_partial`, which follows Rust's `read` semantics, and not `read_exact`'s, and we use it for the dvipdfmx fstream reading.

Likely any place that calls `ttstub_input_read` in a loop is also a candidate for using `ttstub_input_read_partial`, and perhaps it would work better for others.

But there definitely are places in xetex that depend on the `read_exact` semantics. For example, when I naively just tried to make all reads into `ttstub_input_read_partial`, xetex failed pretty quickly on "undumping the format file", which has a few initial 4 byte reads, but then it tries to undump big string pool which was many kilobytes in size. Rust `read` however would only return 8196 minus couple of the the 4 bytes, which would fail a check on the TeX side which expects full read. AFAICT xetex uses `fread`, so this probably just comes down to buffering differences between `fread` and Rust's `Reader` / `BufReader`.

Relates to #935
Fixes #1260


vlasakm:
I provided more details in https://github.com/tectonic-typesetting/tectonic/discussions/1260#discussioncomment-12286253.

CraftSpider:
Thank you for the fix! At some point I'd like to get a fulldeps test suite set up for testing stuff with dependencies, but for now, do you think there's an easy way to write a test that covers this behavior without it?

vlasakm:
> At some point I'd like to get a fulldeps test suite set up for testing stuff with dependencies

Sorry, I am not familiar with Tectonic's tests at all. What does fulldeps mean?

> do you think there's an easy way to write a test that covers this behavior without it?

I assume you mean to test that `pdf:fstream` special gets emitted correctly. I posted LaTeX test case here: https://github.com/tectonic-typesetting/tectonic/discussions/1260#discussioncomment-12286253, I don't know exactly what our options are with regards to testing, but here is a minimal plain XeTeX test with no external dependencies:

```
\special{dvipdfmx:config C 0x40} % don't use object streams
\special{dvipdfmx:config z 0}    % don't use compression

\special{pdf:fstream @SourceFile (\jobname.tex)}
\special{pdf:ann width 10bp height 20bp
  <<
    /Type /Annot
    /Subtype /FileAttachment
    /FS <<
      /Type /Filespec
      /F (\jobname.tex)
      /EF << /F @SourceFile >>
    >>
    /Name /PushPin
    /C [0.8 0.2 0.2]
    /T (Source code of this document)
    /Subj (Source code of this document)
    /Contents (Source code of this document)
  >>
}

\bye
```

It's based on [dvipdfmx manual](https://texdoc.org/serve/dvipdfmx/0), and embeds the TeX source file as an attachment. Other options make the PDF "readable", so we could check that:

1. There are no warnings emitted.
2. That the embedded PDF stream for the source file has the expected length.

CraftSpider:
> Sorry, I am not familiar with Tectonic's tests at all. What does fulldeps mean?

Sorry, jargon borrowed from the Rust compiler. There, the 'fulldeps' are tests that need access to the 'full' compiler, including as an external dependency. Here, I'm using it to imply a project with access to the default bundle or `Tectonic.toml` usage, which would make writing some test cases much easier.

> \<snip> but here is a minimal plain XeTeX test with no external dependencies: \<snip>

Thanks, that's exactly what I meant! For now using external dependencies in tests is hard, so something that only relies on 
XeTeX is perfect. If you want, I can push such a test directly to this branch when I have a minute.

vlasakm:
> > Sorry, I am not familiar with Tectonic's tests at all. What does fulldeps mean?
> 
> Sorry, jargon borrowed from the Rust compiler. There, the 'fulldeps' are tests that need access to the 'full' compiler, including as an external dependency. Here, I'm using it to imply a project with access to the default bundle or `Tectonic.toml` usage, which would make writing some test cases much easier.

I see.

> If you want, I can push such a test directly to this branch when I have a minute.

Please do if you have a minute.

It would take me some time to figure out how to test this in Tectonic test suite, and to be honest, don't have much chances now. I promise this well serve as an example for me to write such test myself in the future.

CraftSpider:
Sorry for such a long delay - I got a bit stuck in the test suite myself, other stuff distracted me... Anyways, I pushed a test that I verified fails on master and passes now that tests this behavior.

CraftSpider:
Alright, neither of the failures are related to this PR, so I'll merge it and fix them in a bit

vlasakm:
@CraftSpider Thank you very much!

PR #1270, opened 2025-04-06
Fix v2 cli flag being incorrectly applied

mky2:
This fixes a problem with the watch command when using v2 cli with a symlink named `nextonic` .


CraftSpider:
Looks good! Thanks for the patience with the 2-month review lag.

PR #1286, opened 2025-06-25
Use Input/Output IDs instead of passing pointers through C

CraftSpider:
This both reduces the chance of weird soundness issues in relation to pointer provenance, and prepares the way for #1138


CraftSpider:
@Mrmaxmeier sorry to ping you again, tectonic-on-arXiv isn't appearing. If it would be helpful, is there anything I can do myself to triage/fix this kind of thing?

Mrmaxmeier:
> @Mrmaxmeier sorry to ping you again, tectonic-on-arXiv isn't appearing. If it would be helpful, is there anything I can do myself to triage/fix this kind of thing?

No problem :slightly_smiling_face:  Results should come in in a bit.
The machine that this was running on was retired a bit ago, so I've moved it somewhere else now. Looks like I'll need to bump up the test timeouts though since the new server is much slower. In the mean time [it'll show a few "crashed" builds](https://tt.ente.ninja/#/summary/825e11b71f7382da7c7e1a871e1590cf0cc9d7a2).

CraftSpider:
Working to trace the change - there's a chance this actually fixed a real issue, or that it somehow broke a condition that it shouldn't have.

CraftSpider:
Clippy failure is testing the wrong commit. I'm going to merge this, and back it out if master sees issues.

PR #1290, opened 2025-07-02
Add fontconfig bridge

CraftSpider:
The bridge compiles for all platforms when it's a dependency - fontconfig is available on mac, and this make it testable on mac if desired, I'll see how CI goes. This instead makes the bridge dependency conditional on targeting mac or not.

Part of #1138

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2954,6 +2954,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "tectonic_bridge_fontconfig"
+version = "0.0.0-dev.0"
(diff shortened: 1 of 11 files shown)


CraftSpider:
Two issues:
- Many cross tests don't have access to any font files
- Mac vcpkg isn't set to install fontconfig. Turns out it's automatically included in the pkgconfig setups.

For the second, I can exclude the crate on mac vcpkg targets easy enough. For the first, I'll see if it's easy to add some file that fontconfig can find, if not, then they may also need to be excluded.

PR #1291, opened 2025-07-06
Switch to using llvm-cov

CraftSpider:
`cargo kcov` has not recieved updates in years, and has increasingly many incompatibilities. Instead of adding more hacks, switch to a coverage tool that:
A) Is regularly updated
B) Is simpler to use
C) Provides more detailed coverage statistics
D) Halves the runtime of the coverage pipeline

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,79 @@
+comment:
+  layout: "condensed_header, condensed_files, diff, components, condensed_footer"
+component_management:
+  individual_components:
+    - component_id: tectonic
+      paths:
+        - "src/**"
+    - component_id: bridge_core
+      paths:
+        - "crates/bridge_core/**"
+    - component_id: bridge_flate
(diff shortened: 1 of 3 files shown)


CraftSpider:
Ouch. That number is more accurate, we simply weren't counting a lot of files before that we are now, and partials aren't considered part of the 'covered' percentage. So our coverage now looks a lot worse.

CraftSpider:
Considering excluding `pdf_io` and/or `xetex` C code - they're a lot of lines of code that is _mostly_ not ours, not really designed for testability, and just not particularly helpful to cover right now. The long-term goal would be to get them tested too, but most likely as part of the port-to-rust initiative

pkgw:
Thanks for tackling this @CraftSpider ! I think the switch is good even if the headline stat goes down; it's more about trying to push it in the right direction.

PR #1292, opened 2025-07-15
Improve coverage

CraftSpider:
Add codecov components, and run all tests


CraftSpider:
#1291 is going ahead, so no need for this

PR #1308, opened 2025-07-30
[bibtex] Refactor string pool

CraftSpider:
Encapsulates and converts to idiomatic Rust the StringPool.


CraftSpider:
This passes tectonic-on-arXiv, so no changes. There's a little fib in the logs now - they claim to have a max strings, but this change uncaps the `StringPool`. It now grows more dynamically, meaning max strings is no longer meaningful. But the goal is byte-for-byte compat, so for now, we keep that log the same.

PR #1316, opened 2025-08-08
fix(tests/bundle_env_overrides): make more robust

bryango:
Previously `test_bundle_prefix` would fail if TECTONIC_BUNDLE_LOCKED is set during the `cargo test` run. This would happen during downstream packaging (e.g. in Nixpkgs). 

Here we make the test more robust and meanwhile we can drop the `test_empty_locked_bundle_ignored` as it has now become redundant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -257,13 +257,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "793db76d6187cd04dff33004d8e6c9cc4e05cd330500379d2394209271b4aeee"
 
 [[package]]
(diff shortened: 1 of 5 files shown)


CraftSpider:
This isn't directly related to this fix, but if you look in `fuzz/Cargo.toml` you can see how to remove the test crate from the global workspace, and set it to not publish to crates.io.

bryango:
> This isn't directly related to this fix, but if you look in `fuzz/Cargo.toml` you can see how to remove the test crate from the global workspace, and set it to not publish to crates.io.

Done! I didn't know how to do this properly, thanks!

PR #1328, opened 2025-10-15
remove dollar signs

lababidi:
The dollar signs `$` in the docs' shell commands are copied when clicking the `copy to clipboard` button.
They then get pasted into the terminal when I'm trying to copy the command over. 

If dollar signs are needed to highlight that the command belongs in the terminal, I believe that there are ways to do that with style updates.

### DIFF
--- a/docs/src/v2cli/bundle.md
+++ b/docs/src/v2cli/bundle.md
@@ -28,7 +28,7 @@ tectonic -X bundle cat <filename>
 #### Example
 
 ```sh
-$ tectonic -X bundle cat latex.ltx
+tectonic -X bundle cat latex.ltx
 %%
 %% This is file `latex.ltx',
 %% generated with the docstrip utility.
@@ -55,7 +55,7 @@ tectonic -X bundle search [term]
 #### Example
 
 ```sh
-$ tectonic -X bundle search minted
+tectonic -X bundle search minted
 minted1.sty
 tcbminted.code.tex
 minted.4ht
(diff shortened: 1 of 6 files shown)


lababidi:
Hi, anyone want to take a look at this?

docs/src/v2cli/bundle.md
bryango: It is a bit weird to remove the dollar signs here (in this file) as they mark the input command prompt which differs from the following outputs. Better not do it in this case.

Also I believe in this case the code block should be ` ```console ` instead of `sh`.

docs/src/v2cli/bundle.md
```diff
@@ -55,7 +55,7 @@ tectonic -X bundle search [term]
 #### Example
 
 ```sh
-$ tectonic -X bundle search minted
```
CraftSpider, line 58: Same here

docs/src/v2cli/bundle.md
```diff
@@ -28,7 +28,7 @@ tectonic -X bundle cat <filename>
 #### Example
 
 ```sh
-$ tectonic -X bundle cat latex.ltx
```
CraftSpider, line 31: This removal is confusing, as now we're mixing shell command and output

CraftSpider (commented):
Largely positive, two small nits. If you don't want to spend more time on this I can merge and fix later. Thanks for the patience, I know wait times can be long.

PR #1338, opened 2026-01-04
fix(deps): replace atty with is-terminal

jxoesneon:
**Problem**: `atty` is unmaintained and flagged for security vulnerability (RUSTSEC-2021-0145: potential unaligned read on Windows).

**Solution**: Replaced `atty` dependency with `is-terminal` (or `std::io::IsTerminal` in future Rust versions).

**Changes**:
- `Cargo.toml`: Replaced dependency.
- `src/bin/tectonic/main.rs`: Updated `atty::is()` calls to `std::io::stdout().is_terminal()`.

**Verification**: `cargo check --bin tectonic` passes.


CraftSpider:
Thank you for the PR! Your branch appears to include a number of release commits from cranko, our release software. Could you please rebase only your changes onto master without the release commits? The release commits are intentionally not included in the master branch.

CraftSpider:
Taking another look at this - tectonic actually dropped atty in #1162. Thank you for the PR, but closing as it's already done.

Issue #1345, opened 2026-04-11
SIGBUS on any \setmainfont{} on macOS arm64 (0.16.0, 0.16.2); 0.15.0 works


jfcampbl:
## Summary

Tectonic 0.16.0 and 0.16.2 crash with SIGBUS (exit 138) on macOS arm64 on **any** `\setmainfont{}` call, with no log file written. Downgrading to 0.15.0 on the same machine, with the same input, produces the PDF successfully.

My suspicion is this is caused by the `xetex_layout` Rust rewrite landed in #1138 ("Port `xetex_layout` entirely to Rust"), but I have no way to confirm that without bisecting.

## Environment

- Host: macOS 26.4.1 (build 25E253), arm64 (Apple Silicon)
- Tectonic: 0.16.2 (installed via Homebrew). Same crash observed on 0.16.0 prior to the `brew upgrade`.
- Font: any system font (see below — not specific to one family).

## Minimal reproduction

```tex
\documentclass{article}
\usepackage{fontspec}
\setmainfont{Helvetica Neue}
\begin{document}
Hello world.
\end{document}
```

```
$ tectonic -X compile mwe.tex --outdir /tmp/mwe_out
note: "version 2" Tectonic command-line interface activated
note: Running TeX ...
$ echo $?
138
```

- Exit 138 = `128 + 10` = SIGBUS on macOS.
- stderr is empty.
- No `.log` file is written under `--outdir`, even with `--keep-logs --keep-intermediates`.
- Trivial `\documentclass{article}...\end{document}` without `fontspec` → works fine on 0.16.2.

## Not specific to Helvetica Neue

Reproduced with every font I tried on this machine:

```
\setmainfont{Helvetica Neue}   → SIGBUS
\setmainfont{Helvetica}        → SIGBUS
\setmainfont{Arial}            → SIGBUS
\setmainfont{Georgia}          → SIGBUS
```

All four fonts are present on the system (verified via `fc-list` and `system_profiler SPFontsDataType`).

## 0.15.0 is fine on the same machine with the same input

Downloaded the `tectonic-0.15.0-aarch64-apple-darwin.tar.gz` precompiled binary from the 0.15.0 release and ran the same MWE:

```
$ tectonic-0.15.0 -X compile mwe.tex --outdir /tmp/mwe_out
warning: accessing absolute path `/System/Library/Fonts/HelveticaNeue.ttc`; build may not be reproducible in other environments
[...]

PR #1346, opened 2026-04-12
Fix SIGBUS crash on macOS arm64 for any `\setmainfont{}` call (#1345)

nadimkobeissi:
Tectonic 0.16.x crashes with SIGBUS (exit 138) on macOS arm64 on any `\setmainfont{}` call (#1345). This PR fixes the root cause and several additional bugs introduced in the `xetex_layout` Rust port (#1138).

### Root cause

`CFArray`'s `Index` implementation (`crates/mac_core/src/array.rs`) was unsound. `CFArrayGetValueAtIndex` returns the stored `CFTypeRef` value directly (i.e., the pointer *is* the value), but the code treated it as a *pointer to* `T` and dereferenced it:

```rust
let ptr = CFArrayGetValueAtIndex(...).cast::<T>();
unsafe { &*ptr }  // reads CF object internals as a NonNull — garbage
```

When `find_font_with_name` called `matches[0].clone()`, this garbage pointer was passed to `CFRetain`, crashing immediately with `EXC_BAD_ACCESS` in `CFRetain`.

### Fix

Replace the broken `Index<usize>` impl with a `get(index) -> T` method that correctly calls `CFRetain` via `new_borrowed` on the returned `CFTypeRef`.

### Additional fixes

While investigating this issue, I stumbled upon other bugs which I initially thought were causing the problem here with `\setmainfont{}` but weren't. I fixed them anyway:

- `c_api.rs`: `Fixed` type on macOS defined as `u32` instead of `i32`, mismatching the C `SInt32` typedef. Negative `scaled_size` values (used by TeX for font probing) became ~65K pt sizes. Changed to `i32`.
- `font.rs:277`: AFM file reading used the already-closed main font `handle` instead of `afm_handle`. Changed to `afm_handle`.
- `font.rs` (macOS init): Several `.unwrap()` calls could panic across `extern "C"` boundaries (undefined behavior). Replaced with `ok_or(())?` / `.map_err()` error propagation.
- `mac_core/font.rs`: `CTFont::new_descriptor` unconditionally unwrapped a potentially-null CoreText return (the original C++ code checked for null). Changed return type to `Option<CTFont>`.

### DIFF
--- a/crates/mac_core/src/font.rs
+++ b/crates/mac_core/src/font.rs
@@ -85,8 +85,7 @@ cfty! {
 
 impl CTFont {
     /// Create a new font from a [`CTFontDescriptor`] and a size to render at.
-    /// Returns `None` if CoreText cannot create the font (rare, but possible).
-    pub fn new_descriptor(descriptor: &CTFontDescriptor, size: f64) -> Option<CTFont> {
+    pub fn new_descriptor(descriptor: &CTFontDescriptor, size: f64) -> CTFont {
         // SAFETY: Provided descriptor is guaranteed valid. Any size will work. Null matrix is always
         //         allowed.
         let ptr = unsafe {
@@ -96,8 +95,9 @@ impl CTFont {
                 ptr::null_mut(),
             )
         };
+        let ptr = NonNull::new(ptr.cast_mut()).unwrap();
         // SAFETY: If non-null, pointer from CTFontCreateWithFontDescriptor is a new, owned CTFont.
-        NonNull::new(ptr.cast_mut()).map(|ptr| unsafe { CTFont::new_owned(ptr) })
+        unsafe { CTFont::new_owned(ptr) }
     }
 
     /// Get an attribute of this font, if present
--- a/crates/mac_core/src/font.rs
+++ b/crates/mac_core/src/font.rs
@@ -85,7 +85,8 @@ cfty! {
 
 impl CTFont {
     /// Create a new font from a [`CTFontDescriptor`] and a size to render at.
-    pub fn new_descriptor(descriptor: &CTFontDescriptor, size: f64) -> CTFont {
+    /// Returns `None` if CoreText cannot create the font (rare, but possible).
+    pub fn new_descriptor(descriptor: &CTFontDescriptor, size: f64) -> Option<CTFont> {
         // SAFETY: Provided descriptor is guaranteed valid. Any size will work. Null matrix is always
         //         allowed.
         let ptr = unsafe {
@@ -95,9 +96,8 @@ impl CTFont {
                 ptr::null_mut(),
             )
         };
-        let ptr = NonNull::new(ptr.cast_mut()).unwrap();
         // SAFETY: If non-null, pointer from CTFontCreateWithFontDescriptor is a new, owned CTFont.
-        unsafe { CTFont::new_owned(ptr) }
+        NonNull::new(ptr.cast_mut()).map(|ptr| unsafe { CTFont::new_owned(ptr) })
     }
 
     /// Get an attribute of this font, if present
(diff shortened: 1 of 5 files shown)


CraftSpider:
Thank you! Could you run rustfmt on the code? I've kicked off CI, and once tests pass and I finish review I'd like to merge this soon, so we can get a fix patch out for macOS users.

This is a great example of a place that could use more tests, but I won't block a fix for users on that.

crates/xetex_layout/src/c_api.rs
```diff
@@ -30,7 +30,7 @@ pub struct GlyphBBox {
 pub type Fixed = i32;
 /// cbindgen:ignore
 #[cfg(target_os = "macos")]
-pub type Fixed = u32;
+pub type Fixed = i32;
```
CraftSpider: This definition is taken from here: https://developer.apple.com/documentation/kernel/fixed, where it's defined as a u32. If this is causing actual bugs, it should be fixed at the boundary between macOS and other code some other way.

crates/mac_core/src/font.rs
```diff
@@ -95,9 +96,8 @@ impl CTFont {
                 ptr::null_mut(),
             )
         };
-        let ptr = NonNull::new(ptr.cast_mut()).unwrap();
```
CraftSpider, line 98: This was an intentional unwrap, as per the documentation [here](https://developer.apple.com/documentation/coretext/ctfontcreatewithfontdescriptor(_:_:_:)): "A best match font is always returned". If this returns null something deeper is wrong.

nadimkobeissi:
> Could you run rustfmt on the code?

@CraftSpider Done!

CraftSpider (changes requested):

CraftSpider:
I have tests for some of this code already on the way. The unwrap changes, as well as the Fixed change, I'm unsure on as several of those are intentional choices, so if those could be split out to another PR I'd like to spend more time on them, while the crash fixes and afm_handle fix I'm happy with and ready to merge as-is.

sophiaeagent-beep:
Independent analysis confirms this PR is correct.

The root cause is in `CFArray<T>::index()` — `CFArrayGetValueAtIndex` returns the stored CFTypeRef directly (it IS the object pointer), but the code casts it to `*const T` and dereferences with `&*ptr`, treating the object's internal memory as a T struct. When `.clone()` calls `CFRetain` on the resulting garbage pointer → SIGBUS.

ARM64 catches this immediately due to strict alignment enforcement. x86_64 is more lenient, which is why it passed CI when PR #1138 landed.

The fix to replace `Index<usize>` with `get() -> T` using `T::new_borrowed()` is semantically correct — CFArray element access inherently involves refcount manipulation that can't be expressed through Index's `&T` return type.

The additional fixes (Fixed type i32 vs u32, AFM handle, unwrap-across-FFI, null CTFont) are all real bugs. Particularly the `Fixed = u32` one — negative scaled_size values used by TeX for font probing wrap to ~65535pt as unsigned.

Tested the analysis against the code paths. This should be merged.

nadimkobeissi:
@CraftSpider I've reverted the more spurious changes, this PR is now focused as you requested.

CraftSpider:
Thank you! As noted, if you'd like to propose the other changes in a different PR, I'd be interested in taking a deeper look. The UB-over-boundaries thing is right, using C-unwind may help there as I think it's worth assuming, even if we remove as much as possible, that we may panic. Similar with `Fixed` - I think just changing the type may be wrong if the docs are correct and it's a u32 on Mac's end, but that definitely could also cause issues on the Tectonic side so there should be a more principled solution there (possible a newtype and explicit conversion from it to the local type, with failures handled)

CraftSpider (approved):

bryango:
This seems important! Will it be released soon-ish? If not, it's also okay, we can cherry-pick it downstream (when packaging).

CraftSpider:
I wanted to get in #1347, since that shows some issues with `CFString` that I'd like to fix to avoid any other macOS issues. I can just make a release now to get this out then merge further improvements later though.

PR #1358, opened 2026-05-01
Add tectonic-on-arxiv as Github Action

CraftSpider:
This is a WIP idea to port tectonic-on-arxiv to use the actions workflow, allowing us to rely on github's artifacts and compute.

@Mrmaxmeier for thoughts here - as this is your brainchild primarily.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,3 +12,7 @@ jobs:
     needs: prep
     uses: ./.github/workflows/build_and_test.yml
     secrets: inherit
+  tectonic-on-arxiv:
+    needs: prep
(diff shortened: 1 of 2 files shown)


CraftSpider:
Some elaboration on specific ideas: 
- We can compare runs using artifacts
- Markdown run results display the diff, I haven't tested yet but I hope they support HTML collapsibles
- If a run is missing, maybe possible to trigger the workflow for it, not sure
- With manual workflow triggers, people can request a comparison of any two commits hopefully easily
- We do lose out a bit on long term dataset storage, without some hacks. Maybe add the dataset to release tags, or just store it elsewhere

PR #1361, opened 2026-05-04
Drop error-chain dep, switch binary to use `anyhow`

CraftSpider:
I'd like to get this in before 0.17 - if we're changing from `app_dirs2` to `directory`, I'd like to make this breaking change in the same release.

Drops error-chain, as it's significantly out of date and deprecated. Instead we now use `anyhow` in the binary crate through the re-export in `tectonic_errors`. This both removes several transitive dependencies and allows cleaning up some warnings related to using such an old dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 4
 
 [[package]]
-name = "addr2line"
-version = "0.25.1"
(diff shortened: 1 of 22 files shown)


CraftSpider:
@Mrmaxmeier funny ToA error about failing to open a downloaded crate

PR #1363, opened 2026-05-09
Change bundle caching to only re-check remote once per week

CraftSpider:
Uses chrono to write another file in addition to the hash which carries a bundle's last-checked time. This prevents constant re-checks on every engine run. This is important because nearly half of tectonic's runtime is spent reading the entire bundle off the internet on every invocation.

@rm-dr, since you did the original bundle rework.


CraftSpider:
So, without this, _every_ build on Tectonic-on-arXiv is failing. With it, only most are. So that's good :P

PR #1368, opened 2026-05-14
Use ctrl-c to kill watchexec for profiling

CraftSpider:


CraftSpider:
Still busted because it breaks on Windows, sad

PR #1369, opened 2026-05-16
Remove usage of long-deprecated `CARGO_BIN_PATH`

CraftSpider:
Changes to use the better-supported `CARGO_BIN_EXE_<name>`


CraftSpider:
CI failures are pre-existing, merging as-is and will fix them later

PR #1372, opened 2026-06-09
bibtex: don't panic on an empty @string macro

osdnk:
## Problem

Tectonic's built-in BibTeX panics whenever a `@string` macro that expands to the
empty string is referenced at the **start** of a stored field:

```
thread 'main' panicked at crates/engine_bibtex/src/scan.rs:663:41:
index out of bounds: the len is 0 but the index is 0
```

In practice this means tectonic cannot compile **any** paper that cites
[cryptobib](https://github.com/cryptobib/export), the crypto community's shared
bibliography: its `abbrev3.bib` defines well over a thousand venue
`name`/`address` macros as `""` (e.g. `@string{acispname = ""}`), and as soon as
an entry references one in a stored field such as `address`, the BibTeX pass
aborts — even when a correct `.bbl` already exists, since the pass re-runs
regardless.

## Root cause

In `scan_a_field_token_and_eat_white`, a macro token is expanded into `str` and,
at the start of a field value, its first byte is inspected to collapse leading
whitespace:

```rust
let mut str = globals.pool.get_str(strnum);

if globals.buffers.offset(BufTy::Ex, 1) == 0
    && LexClass::of(str[0]) == LexClass::Whitespace   // panics when str == ""
{
```

For an empty macro, `str` is a zero-length slice, so `str[0]` is out of bounds.
The original WEB/C BibTeX reads `str_pool[tmp_ptr]` from a shared array here and
thus reads one byte past harmlessly; the Rust port returns an exact-length
slice, turning the same access into a panic. The copy loop a few lines below
already guards with `!str.is_empty()` — the leading-whitespace check just didn't.

## Fix

One line — guard the check the same way:

```rust
if globals.buffers.offset(BufTy::Ex, 1) == 0
    && !str.is_empty()
    && LexClass::of(str[0]) == LexClass::Whitespace
{
```

An empty macro then contributes nothing to the field, matching reference
`bibtex`/`bibtex8`.

## Test

Adds `tests/bibtex/cites/empty_macro` (an empty `@string` referenced via `#` at
the start of a field). Its expected `.bbl` is **byte-identical** to the output
of `bibtex8` on the same input. The full bibtex suite (24 tests) passes.

## Reproduction

Minimal, self-contained (two files plus a driver; no cryptobib needed):
https://github.com/osdnk/tectonic-bibtex-empty-macro-repro

```sh
git clone https://github.com/osdnk/tectonic-bibtex-empty-macro-repro
cd tectonic-bibtex-empty-macro-repro && ./reproduce.sh
```

Marked draft pending maintainer feedback on the approach.

### DIFF
--- a/crates/engine_bibtex/src/scan.rs
+++ b/crates/engine_bibtex/src/scan.rs
@@ -660,6 +660,7 @@ fn scan_a_field_token_and_eat_white(
                     let mut str = globals.pool.get_str(strnum);
 
                     if globals.buffers.offset(BufTy::Ex, 1) == 0
+                        && !str.is_empty()
                         && LexClass::of(str[0]) == LexClass::Whitespace
                     {
                         if globals.buffers.offset(BufTy::Ex, 1) >= globals.buffers.len() {
--- a/tests/bibtex.rs
+++ b/tests/bibtex.rs
@@ -96,8 +96,6 @@ fn test_single_entry() {
     TestCase::new(&["cites", "single_entry"]).go()
 }
 
-/// Empty `@string` macro ("") at the start of a stored field: used to panic in
-/// `scan_a_field_token_and_eat_white` (cryptobib defines many such macros).
 #[test]
 fn test_empty_macro() {
     TestCase::new(&["cites", "empty_macro"]).go()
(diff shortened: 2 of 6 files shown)


osdnk:
@CraftSpider -- is it possible for you to take a look at that? Thanks in advance!

CraftSpider:
Looks great, thanks! Once CI finishes it should merge (if anything fails for reasons _not_ related to your change I'll merge it after I confirm)

PR #1377, opened 2026-06-30
pdf_io: support fonts with only a Unicode (0,4) cmap

hhh2210:
## Problem

Some TrueType fonts used on macOS, including STHeiti and STXihei, expose a single Unicode `cmap`: platform 0, encoding 4, format 12. Tectonic renders these fonts, but affected text can be missing `/ToUnicode` metadata and cannot be searched or copied correctly.

## Root cause

Tectonic's vendored dvipdfm-x probes fixed `(platform, encoding)` lists in three paths:

- `/ToUnicode` generation;
- `unicode` font-map loading;
- CIDFontType2 embedding.

Those paths recognized other Unicode subtables but skipped `(0,4)`. Fonts with a platform-3 subtable worked, while `(0,4)`-only fonts did not.

## Fix

Port all three fallbacks from [the upstream dvipdfm-x change](https://github.com/texjporg/tex-jp-build/commit/df04c0c6317e5c566c52577fe2a80499e487479d):

- add `(0,4)` to `cmap_plat_encs[]` for `/ToUnicode` generation;
- add the `(0,4)` fallback in `otf_load_Unicode_CMap()`;
- add `(0,4)` to `known_encodings[]` and update its bound for CIDFontType2 embedding.

The existing format checks and mapping logic are unchanged.

## Review guide

The two commits separate code from fixtures:

1. `pdf_io: support Unicode cmap (0,4) fallbacks` contains the three small dvipdfm-x changes.
2. `tests: cover cmap (0,4)-only TrueType fonts` adds two complementary regressions and their generated fixture.

`cmap04-only.ttf` is built from simple test-only outlines by `tests/assets/generate_cmap04_test_font.py`. It is 1,284 bytes, has three glyphs, and contains exactly one `(0,4)` format-12 cmap. Its name and license records identify it as the MIT-licensed `TectonicCmap04Test`; it contains no third-party font data.

- `cmap04_unicode` covers both `unicode` font-map loading and generated `/ToUnicode`.
- `cidtype2_cmap04` covers the separate CIDFontType2 probe table.

## Validation

```console
cargo test --test tex-outputs cmap04 -- --nocapture
# 2 passed
cargo test --test tex-outputs
# 32 passed
cargo clippy --all --all-targets --all-features -- --deny warnings
```

The generated `cmap04_unicode.pdf` embeds `TectonicCmap04Test` with `uni=yes`, and `pdftotext` extracts `A`. Removing either the `otf_load_Unicode_CMap()` fallback or the `cmap_plat_encs[]` entry makes that regression fail. Removing the `known_encodings[]` entry makes `cidtype2_cmap04` fail.

On the original Chinese thesis, extracted CJK characters increased from 23,957 to 26,918, matching XeLaTeX. The STHeiti and STXihei objects gained `/ToUnicode`, and a page-by-page pixel comparison found no layout changes.

## Scope

This change only extends cmap discovery. It does not alter font discovery, shaping, layout, or variable-font handling.


hhh2210:
@CraftSpider This is ready for code review. I rebased it onto current master, ported all three upstream `(0,4)` cmap fallbacks, and added two regression tests using a generated test-only font. The fork workflow is waiting for maintainer approval before GitHub will start it. PTAL when you have time.

PR #1380, opened 2026-07-10
prevent slice::from_raw_parts from being passed a null pointer

z-erica:
when nfont is zero in a FcFontSet, the fonts field can be a null pointer. this can make the test suite panic:

> thread 'trip_test' (7786) panicked at crates/bridge_fontconfig/src/font_set.rs:23:18:
> unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`
>
> This indicates a bug in the program. This Undefined Behavior check is optional, and cannot be relied on for safety.
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> thread caused non-unwinding panic. aborting.

this fix maintains the safety invariant by always returning an empty slice with a lifetime of 'static when the array is empty


CraftSpider:
Thanks for the catch! Failures are unrelated to this PR

